PR #2, opened 2019-08-07
Bug 1738495: Invalid token is reported as error in cluster status

smarterclayton:
If the token for the cluster contains a newline, the error is
reported in a generic fashion to the cluster operator status which
contains the auth token value. This is an information leak and
should not happen.

Prevent future disclosure bugs by not printing the exact error when
we build an invalid request. Check the token value in both the client
(before setting the header) and when we read from config (before
loading the token) so that we report this error in an earlier place
with a specific error.

To reproduce, put a newline inside the cloud.openshift.com token in
the cluster pull-secret (in openshift-config namespace).

/assign @mfojtik

### DIFF
--- a/pkg/config/configobserver/configobserver.go
+++ b/pkg/config/configobserver/configobserver.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
@@ -77,9 +78,15 @@ func (c *Controller) retrieveToken() error {
 			if err := json.Unmarshal(data, &pullSecret); err != nil {
 				klog.Errorf("Unable to unmarshal cluster pull-secret: %v", err)
 			}
-			if auth, ok := pullSecret.Auths["cloud.openshift.com"]; ok && len(auth.Auth) > 0 {
-				klog.V(4).Info("Found cloud.openshift.com token")
-				nextConfig.Token = auth.Auth
+			if auth, ok := pullSecret.Auths["cloud.openshift.com"]; ok {
+				token := strings.TrimSpace(auth.Auth)
+				if strings.Contains(token, "\n") || strings.Contains(token, "\r") {
+					return fmt.Errorf("cluster authorization token is not valid: contains newlines")
+				}
+				if len(token) > 0 {
+					klog.V(4).Info("Found cloud.openshift.com token")
+					nextConfig.Token = token
+				}
 			}
 		}
 		nextConfig.Report = true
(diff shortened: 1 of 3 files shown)


pkg/config/configobserver/configobserver.go
```diff
@@ -77,9 +78,15 @@ func (c *Controller) retrieveToken() error {
...
 			}
-			if auth, ok := pullSecret.Auths["cloud.openshift.com"]; ok && len(auth.Auth) > 0 {
-				klog.V(4).Info("Found cloud.openshift.com token")
-				nextConfig.Token = auth.Auth
+			if auth, ok := pullSecret.Auths["cloud.openshift.com"]; ok {
+				token := strings.TrimSpace(auth.Auth)
+				if strings.Contains(token, "\n") || strings.Contains(token, "\r") {
+					return fmt.Errorf("cluster authorization token is not valid: contains newlines")
+				}
+				if len(token) > 0 {
```
mfojtik: how about empty token?

pkg/config/configobserver/configobserver.go
```diff
@@ -77,9 +78,15 @@ func (c *Controller) retrieveToken() error {
...
 				klog.Errorf("Unable to unmarshal cluster pull-secret: %v", err)
 			}
-			if auth, ok := pullSecret.Auths["cloud.openshift.com"]; ok && len(auth.Auth) > 0 {
-				klog.V(4).Info("Found cloud.openshift.com token")
-				nextConfig.Token = auth.Auth
+			if auth, ok := pullSecret.Auths["cloud.openshift.com"]; ok {
+				token := strings.TrimSpace(auth.Auth)
+				if strings.Contains(token, "\n") || strings.Contains(token, "\r") {
+					return fmt.Errorf("cluster authorization token is not valid: contains newlines")
+				}
```
mfojtik: seeing original error can be useful for debug

mfojtik: github on mobile sux, this is for line 115 in insightsclient

pkg/config/configobserver/configobserver.go
smarterclayton: I don't know how many users out there have edited their pull secrets to remove the token but didn't remove the key.  I don't feel strongly, can make this be an error, but since only telemetry and insights use it I was unsure how many we would have in prod that behaved this way.  Also, in docker serialization empty and absent are the same (in most of the places that consume this) so it's likely there are people who would remove the value but not the key.

pkg/config/configobserver/configobserver.go
smarterclayton: It's a security leak, since the value of the token is printed to cluster operator and if we gather that then we're leaking their pull secret back to us.

PR #4, opened 2019-08-09
WIP: Prototype what would it take to wire config recorder

mfojtik:
This will wire experimental controller that list watch all CRD's with "config.openshift.io" suffix and create dynamic informer for each CRD within this group.
The informer then list and watch for changes in each instance and record all changes into local GIT repository.
The repository is not yet wired or send to insights, this will require more discussion.

/cc @smarterclayton


mfojtik:
@smarterclayton if this works, we can enable it for CI runs maybe and collect the git repo folder (`git archive`) with other artifacts ?

mfojtik:
@deads2k since you asked to have this enabled in CI runs... I think long-term this is the right place for this to live.

iNecas:
My feeling is we want to have a way to truncate the old commits at some point. What would be the story? Just restarting the pod and counting on the fact that the storage is ephemeral? Do we need something more sophisticated?

mfojtik:
@iNecas before we build some clever pruning mechanism, I would want to know when the size/number of commits become problem. Is 1M commits problem? Is 10M commits problem?

Additionally, if we say, that we are going to upload the `git archived` repository every 15 minutes to insights, then if we nuke this repository on every upgrade, when insights operator redeploy, then I would say we have plenty backups already on insights side.

The problem this is hopefully going to solve is: "What happened to my cluster between July/1 and July/6". In this case, we can look into insights, find the Git repo we uploaded on July/6 and then `git log`. 

Another option would be to build-in visualization directly to operator. Something that output long diff file that we upload?

mfojtik:
To try this out, you can use cluster-bot and point it to this PR. Then `oc rsh` into insights operator pod and check `/var/lib/insights-history-repo` folder. ([sample GIT repo](http://file.brq.redhat.com/~mfojtik/insights-history-repository.tgz))

mfojtik:
Another idea is to only upload diffs between last sync. So basically, you generate a patch every 15 minutes that will contain all changes that happened in last 15 minutes. On insights side, you can apply the patch and have complete history.

mfojtik:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_insights-operator/4/pull-ci-openshift-insights-operator-master-e2e-aws/10/artifacts/e2e-aws/pods/openshift-insights_insights-operator-6676b9cfdd-hfzqw_operator.log

yay!

mfojtik:
@smarterclayton as the concern was this is big, I added a knob and disabled the history recording by default. I still think this can be super useful for our CI system (where we can enable it).

mfojtik:
I added sample git reporter that export the changes in git-log-like way.

alexandrevicenzi:
@mfojtik what is the status of this WIP?

PR #8, opened 2019-08-22
pkg/insights/insightsclient/insightsclient: Change 'file' -> 'upload'

wking:
To get better server logging while we wait for RedHatInsights/insights-ingress-go#81 to go into production.


wking:
/hold

RedHatInsights/insights-ingress-go#81 may be in production now, so we shouldn't need to merge this.  I'll leave it open so the CI jobs will run just in case, and then close it.

wking:
All the tests have wrapped up.

/close

PR #10, opened 2019-08-23
Provide initial proxy support for insights operator

mfojtik:
This wire the proxy certs into operator and enable proxy vars injections. The insights client then use transport that has the correct trust set.

@danehans i think we missed this one.

/cc @smarterclayton @iNecas

### DIFF
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -6,13 +6,12 @@ import (
 	"fmt"
 	"os"
 
-	"k8s.io/client-go/rest"
-
 	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/version"
+	"k8s.io/client-go/rest"
 
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -2,6 +2,9 @@ package insightsclient
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -10,15 +13,16 @@ import (
 	"net"
 	"net/http"
 	"net/textproto"
+	"os"
 	"strconv"
 	"time"
 
+	"k8s.io/client-go/pkg/version"
 	"k8s.io/client-go/transport"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/klog"
 
-	"k8s.io/client-go/pkg/version"
+	"github.com/prometheus/client_golang/prometheus"
 
 	configv1 "github.com/openshift/api/config/v1"
 
@@ -52,7 +56,7 @@ var ErrWaitingForVersion = fmt.Errorf("waiting for the cluster version to be loa
 
 func New(client *http.Client, maxBytes int64, metricsName string, authorizer Authorizer, clusterInfo ClusterVersionInfo) *Client {
 	if client == nil {
-		client = &http.Client{Transport: DefaultTransport()}
+		client = &http.Client{}
 	}
 	if maxBytes == 0 {
 		maxBytes = 10 * 1024 * 1024
@@ -64,6 +68,49 @@ func New(client *http.Client, maxBytes int64, metricsName string, authorizer Aut
 		authorizer:  authorizer,
 		clusterInfo: clusterInfo,
 	}
+}
+
+func getTrustedCABundle() (*x509.CertPool, error) {
+	caBytes, err := ioutil.ReadFile("/var/run/configmaps/trusted-ca-bundle/ca-bundle.crt")
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	if len(caBytes) == 0 {
+		return nil, nil
+	}
+	certs := x509.NewCertPool()
+	if ok := certs.AppendCertsFromPEM(caBytes); !ok {
+		return nil, errors.New("error loading cert pool from ca data")
+	}
+	return certs, nil
+}
+
+func clientTransport() http.RoundTripper {
+	// default transport, proxy from env
+	clientTransport := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		Dial: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}).Dial,
+		TLSHandshakeTimeout: 10 * time.Second,
+		DisableKeepAlives:   true,
+	}
+
+	// get the cluster proxy trusted CA bundle in case the proxy need it
+	rootCAs, err := getTrustedCABundle()
+	if err != nil {
+		klog.Errorf("Failed to get proxy trusted CA: %v", err)
+	}
+	if rootCAs != nil {
+		clientTransport.TLSClientConfig = &tls.Config{}
+		clientTransport.TLSClientConfig.RootCAs = rootCAs
+	}
+
+	return transport.DebugWrappers(clientTransport)
 }
 
 func (c *Client) Send(ctx context.Context, endpoint string, source Source) error {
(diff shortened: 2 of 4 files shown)


pkg/insights/insightsclient/insightsclient.go
```diff
@@ -66,6 +72,32 @@ func New(client *http.Client, maxBytes int64, metricsName string, authorizer Aut
 	}
 }
 
+func (c *Client) setProxyClientTransport() error {
```
iNecas: Where is this function being used?

pkg/controller/operator.go
```diff
@@ -2,21 +2,24 @@ package controller
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/version"
+	"k8s.io/client-go/rest"
 
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 
+	v1 "github.com/openshift/api/config/v1"
```
iNecas: Mind using more descriptive name, ideally using `config` in it. `v1` is too vague IMO

mfojtik: fixed

pkg/insights/insightsclient/insightsclient.go
mfojtik: i refactored all this code

mfojtik:
@smarterclayton @iNecas i think this is ready for review... 

it is missing test, but I promise I add one later :]

pkg/controller/operator.go
```diff
@@ -112,6 +116,36 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
...
+			return nil, nil
+		}
+		if err != nil {
+			return nil, err
+		}
+		return &proxyConfig.Status, nil
+	}
+
+	// read proxy CA bundle
+	insightsClient.GetProxyRootCAsFunc = func() (*x509.CertPool, error) {
```
stlaz: Should probably be `GetTrustedCABundleFunc`, the only connection `trusted-ca-bundle` has to proxy is that it may contain CA cert for a TLS MITM proxy, but otherwise it's really quite a general trust bundle

mfojtik:
@smarterclayton now I'm not sure if CVO injects the proxy env vars into operator. If yes, we don't need this. However we still need to wire the certs.

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -66,6 +73,49 @@ func New(client *http.Client, maxBytes int64, metricsName string, authorizer Aut
...
+	// get the cluster proxy configuration and override the client transport
+	proxyStatus, err := c.GetProxyStatusFunc()
+	if err != nil {
+		klog.Errorf("Failed to get proxy configuration: %v", err)
+		return transport.DebugWrappers(clientTransport)
+	}
+	if proxyStatus == nil {
+		return transport.DebugWrappers(clientTransport)
+	}
+	clientTransport.Proxy = func(req *http.Request) (*url.URL, error) {
```
stlaz: You wouldn't need to be getting the proxy object and only have your operator injected with the vars by the CVO if you use this annotation: https://github.com/openshift/cluster-authentication-operator/blob/master/manifests/07_deployment.yaml#L9

That would restart your operator on config change. I'll let you pick whichever you like better.

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -66,6 +70,32 @@ func New(client *http.Client, maxBytes int64, metricsName string, authorizer Aut
 	}
 }
 
+func (c *Client) clientTransport() http.RoundTripper {
+	// default transport, proxy from env
+	clientTransport := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
```
stlaz: This may not work correctly for CIDRs in NO_PROXY when built with Go < 1.11, not sure if that could be a problem here, but it should be possible to avoid that with `k8s.io/apimachinery/pkg/util/net.NewProxierWithNoProxyCIDR()`

Can be addressed in a follow up if the above's a problem.

mfojtik:
/hold

For @iNecas or @smarterclayton approval.

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -66,6 +70,32 @@ func New(client *http.Client, maxBytes int64, metricsName string, authorizer Aut
...
+		}).Dial,
+		TLSHandshakeTimeout: 10 * time.Second,
+		DisableKeepAlives:   true,
+	}
+
+	// get the cluster proxy trusted CA bundle in case the proxy need it
+	rootCAs, err := c.GetTrustedCABundleFunc()
+	if err != nil {
+		klog.Errorf("Failed to get proxy trusted CA: %v", err)
+		return transport.DebugWrappers(clientTransport)
```
iNecas: This line can be left out, which I would prefer, as it reduces the number of places too look at when searching for what the function returns (+ reduces duplicate code)

pkg/controller/operator.go
```diff
@@ -112,6 +113,23 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	authorizer := clusterauthorizer.New(configObserver)
 	insightsClient := insightsclient.New(nil, 0, "default", authorizer, configPeriodic)
 
+	// read proxy CA bundle
+	insightsClient.GetTrustedCABundleFunc = func() (*x509.CertPool, error) {
```
iNecas: Given now the function doesn't depend on anything from this context, would be simpler just put it directly to `insightsclient.go` or you did that more for testing purposes?

pkg/controller/operator.go
```diff
@@ -112,6 +113,23 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
...
+			if os.IsNotExist(err) {
+				return nil, nil
+			}
+			return nil, err
+		}
+		if len(caBytes) == 0 {
+			return nil, nil
+		}
+		certs := x509.NewCertPool()
+		certs.AppendCertsFromPEM(caBytes)
```
iNecas: Nitpick: AppendsCertsFromPEM actually returns tuple with `ok` variable, so we should probably check it similarly as other operators do https://github.com/openshift/cluster-authentication-operator/blob/df3160d1ab9f24448c853bfe567c44a2763017cb/pkg/operator2/transport.go#L42-L45

pkg/controller/operator.go
iNecas: Even for testing purposes, it would be good to set the default function in the `insightsclient.New` function, otherwise, there is a possibility for `GetTrustedCABundleFunc` to return `nil`, risking `panic: runtime error: invalid memory address or nil pointer dereference` eventually. I know the chances are low, but if we can choose between two approaches, let's go the safer one.

smarterclayton:
You have to request injection of the proxy variables

iNecas:
Btw. is there a procedure described somewhere on how to setup the proxy on openshift, to test this e2e?

mfojtik:
@iNecas comments addressed, thanks!
@smarterclayton I already updated the deployment in manifests to request injection.

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -66,6 +70,49 @@ func New(client *http.Client, maxBytes int64, metricsName string, authorizer Aut
...
+		return nil, nil
+	}
+	certs := x509.NewCertPool()
+	if ok := certs.AppendCertsFromPEM(caBytes); !ok {
+		return nil, errors.New("error loading cert pool from ca data")
+	}
+	return certs, nil
+}
+
+func (c *Client) clientTransport() http.RoundTripper {
```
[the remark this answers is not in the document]
mfojtik: done

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -94,20 +141,29 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
...
-			pw.CloseWithError(err)
+			if err := pw.CloseWithError(err); err != nil {
+				klog.Warningf("Failed to close pipe: %v", err)
+			}
 			return
 		}
 		r := &LimitedReader{R: source.Contents, N: c.maxBytes}
 		if _, err := io.Copy(fw, r); err != nil {
-			pw.CloseWithError(err)
+			if err := pw.CloseWithError(err); err != nil {
```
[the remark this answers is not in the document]
mfojtik: done

PR #14, opened 2019-08-29
gather: Capture failing nodes during gather pass

smarterclayton:
A failing node has significant information for understanding why the node is failed. Anonymize most information on the node and return it when it is marked as NotReady.

Moves recorder flush to the gatherer under the same execution context.

### DIFF
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -59,7 +59,7 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	ctx := context.Background()
 
 	// these are operator clients
-	client, err := kubernetes.NewForConfig(controller.ProtoKubeConfig)
+	kubeClient, err := kubernetes.NewForConfig(controller.ProtoKubeConfig)
 	if err != nil {
 		return err
 	}
@@ -69,11 +69,19 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	}
 
 	// these are gathering clients
-	gatherCRKubeConfig := rest.CopyConfig(controller.KubeConfig)
+	gatherProtoKubeConfig := rest.CopyConfig(controller.ProtoKubeConfig)
 	if len(s.Impersonate) > 0 {
-		gatherCRKubeConfig.Impersonate.UserName = s.Impersonate
+		gatherProtoKubeConfig.Impersonate.UserName = s.Impersonate
 	}
-	gatherConfigClient, err := configv1client.NewForConfig(gatherCRKubeConfig)
+	gatherKubeConfig := rest.CopyConfig(controller.KubeConfig)
+	if len(s.Impersonate) > 0 {
+		gatherKubeConfig.Impersonate.UserName = s.Impersonate
+	}
+	gatherKubeClient, err := kubernetes.NewForConfig(gatherProtoKubeConfig)
+	if err != nil {
+		return err
+	}
+	gatherConfigClient, err := configv1client.NewForConfig(gatherKubeConfig)
 	if err != nil {
 		return err
 	}
@@ -86,7 +94,7 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	}
 
 	// configobserver synthesizes all config into the status reporter controller
-	configObserver := configobserver.New(s.Controller, client)
+	configObserver := configobserver.New(s.Controller, kubeClient)
 	go configObserver.Start(ctx)
 
 	// the status controller initializes the cluster operator object and retrieves
@@ -100,7 +108,7 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 
 	// the gatherers periodically check the state of the cluster and report any
 	// config to the recorder
-	configPeriodic := clusterconfig.New(gatherConfigClient)
+	configPeriodic := clusterconfig.New(gatherConfigClient, gatherKubeClient.CoreV1())
 	periodic := periodic.New(s.Interval, recorder, map[string]gather.Interface{
 		"config": configPeriodic,
 	})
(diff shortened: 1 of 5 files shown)


pkg/controller/operator.go
```diff
@@ -69,10 +69,18 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	}
 
 	// these are gathering clients
+	gatherProtoKubeConfig := rest.CopyConfig(controller.ProtoKubeConfig)
+	if len(s.Impersonate) > 0 {
+		gatherProtoKubeConfig.Impersonate.UserName = s.Impersonate
+	}
 	gatherCRKubeConfig := rest.CopyConfig(controller.KubeConfig)
```
iNecas: What does `CR` stand for here?

pkg/controller/operator.go
```diff
@@ -69,10 +69,18 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
...
 	if len(s.Impersonate) > 0 {
 		gatherCRKubeConfig.Impersonate.UserName = s.Impersonate
 	}
+	gatherClient, err := kubernetes.NewForConfig(gatherProtoKubeConfig)
```
iNecas: `gatherClient` is too generic, what about `gatherK8sClient`?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -219,20 +213,36 @@ func (a ClusterOperatorAnonymizer) Marshal(_ context.Context) ([]byte, error) {
...
-// 	}
-// 	sort.Strings(rvs)
-// 	return strings.Join(rvs, ",")
-// }
+func anonymizeNode(node *corev1.Node) *corev1.Node {
+	for k := range node.Annotations {
+		node.Annotations[k] = ""
+	}
+	for k, v := range node.Labels {
+		node.Labels[k] = strings.Repeat("x", len(v))
```
iNecas: Mind extracting the replacement into separate function?

pkg/controller/operator.go
smarterclayton: Custom resource, you can’t use proto with crds yet from the same client

smarterclayton:
Comments addressed

pkg/controller/operator.go
iNecas: Ty

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -18,38 +21,66 @@ import (
...
+				return records, nil
+			},
+			func() ([]record.Record, []error) {
+				nodes, err := i.coreClient.Nodes().List(metav1.ListOptions{})
+				if err != nil {
+					return nil, []error{err}
+				}
+				records := make([]record.Record, 0, len(nodes.Items))
+				for i := range nodes.Items {
+					if isFailingNode(&nodes.Items[i]) {
```
iNecas: I've tried to simulate this with a cluster where I made one of the workers marked as 'NotReady' but nothing got collected about it. Is this aiming for worker nodes as well, or only masters?

smarterclayton: how did you. make the worker mark as not ready?

smarterclayton: When nodes are partitioned from the cluster they are status Unknown.  Fixed, thanks

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -18,38 +21,67 @@ import (
...
-			for i := range config.Items {
-				records = append(records, record.Record{Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[i].Name), Item: ClusterOperatorAnonymizer{&config.Items[i]}})
-			}
-			return records, nil
-		},
+	return record.Collect(ctx, recorder,
+		record.Aggregate(
+			func() ([]record.Record, []error) {
+				config, err := i.client.ClusterOperators().List(metav1.ListOptions{})
+				if errors.IsNotFound(err) {
```
eparis: How would this ever happen? I guess you are ignoring it just so we might get other data?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -18,38 +21,67 @@ import (
...
-				return nil, []error{err}
-			}
-			records := make([]record.Record, 0, len(config.Items))
-			for i := range config.Items {
-				records = append(records, record.Record{Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[i].Name), Item: ClusterOperatorAnonymizer{&config.Items[i]}})
-			}
-			return records, nil
-		},
+	return record.Collect(ctx, recorder,
+		record.Aggregate(
```
eparis: Why do we record.Aggregate() the first 2 things, and just comma list the others things?

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: Right, someone screws up incredibly badly and deletes CRDs

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: Can do it in a follow up - agree it's not as clean as it should be

smarterclayton: (probably would be clearer if the entire thing was `fns ... func() ([]Record, []error)`

eparis:
/lgtm
I looked at the anonymization. Looks good to me.

pkg/gather/clusterconfig/clusterconfig.go
iNecas: I can confirm this fixes the issue for me

PR #15, opened 2019-08-29
gather: Capture failing pods from broken operators during gather pass

smarterclayton:
Pods that are failing within namespaces referenced by cluster operators that are degraded or unavailable should be returned in order to get termination messages or other status conditions that will assist in debugging seriously broken clusters.

First commit is from #14

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -46,133 +46,131 @@ var reInvalidUIDCharacter = regexp.MustCompile(`[^a-z0-9\-]`)
 
 func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error {
 	return record.Collect(ctx, recorder,
-		record.Aggregate(
-			func() ([]record.Record, []error) {
-				config, err := i.client.ClusterOperators().List(metav1.ListOptions{})
-				if errors.IsNotFound(err) {
-					return nil, nil
-				}
-				if err != nil {
-					return nil, []error{err}
-				}
-				records := make([]record.Record, 0, len(config.Items))
-				for i := range config.Items {
-					records = append(records, record.Record{Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[i].Name), Item: ClusterOperatorAnonymizer{&config.Items[i]}})
-				}
+		func() ([]record.Record, []error) {
+			config, err := i.client.ClusterOperators().List(metav1.ListOptions{})
+			if errors.IsNotFound(err) {
+				return nil, nil
+			}
+			if err != nil {
+				return nil, []error{err}
+			}
+			records := make([]record.Record, 0, len(config.Items))
+			for i := range config.Items {
+				records = append(records, record.Record{Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[i].Name), Item: ClusterOperatorAnonymizer{&config.Items[i]}})
+			}
 
-				for _, item := range config.Items {
-					if isHealthyOperator(&item) {
-						continue
+			for _, item := range config.Items {
+				if isHealthyOperator(&item) {
+					continue
+				}
+				for _, namespace := range namespacesForOperator(&item) {
+					pods, err := i.coreClient.Pods(namespace).List(metav1.ListOptions{})
+					if err != nil {
+						klog.V(2).Infof("Unable to find pods in namespace %s for failing operator %s", namespace, item.Name)
 					}
-					for _, namespace := range namespacesForOperator(&item) {
-						pods, err := i.coreClient.Pods(namespace).List(metav1.ListOptions{})
-						if err != nil {
-							klog.V(2).Infof("Unable to find pods in namespace %s for failing operator %s", namespace, item.Name)
+					for i := range pods.Items {
+						if isHealthyPod(&pods.Items[i]) {
+							continue
 						}
-						for i := range pods.Items {
-							if isHealthyPod(&pods.Items[i]) {
-								continue
-							}
-							records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pods.Items[i].Namespace, pods.Items[i].Name), Item: PodAnonymizer{&pods.Items[i]}})
-						}
+						records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pods.Items[i].Namespace, pods.Items[i].Name), Item: PodAnonymizer{&pods.Items[i]}})
 					}
 				}
+			}
 
-				return records, nil
-			},
-			func() ([]record.Record, []error) {
-				nodes, err := i.coreClient.Nodes().List(metav1.ListOptions{})
-				if err != nil {
-					return nil, []error{err}
+			return records, nil
+		},
+		func() ([]record.Record, []error) {
+			nodes, err := i.coreClient.Nodes().List(metav1.ListOptions{})
+			if err != nil {
+				return nil, []error{err}
+			}
+			records := make([]record.Record, 0, len(nodes.Items))
+			for i := range nodes.Items {
+				if isHealthyNode(&nodes.Items[i]) {
+					continue
 				}
-				records := make([]record.Record, 0, len(nodes.Items))
-				for i := range nodes.Items {
-					if isHealthyNode(&nodes.Items[i]) {
-						continue
-					}
-					records = append(records, record.Record{Name: fmt.Sprintf("config/node/%s", nodes.Items[i].Name), Item: NodeAnonymizer{&nodes.Items[i]}})
-				}
+				records = append(records, record.Record{Name: fmt.Sprintf("config/node/%s", nodes.Items[i].Name), Item: NodeAnonymizer{&nodes.Items[i]}})
+			}
 
-				return records, nil
-			},
-		),
-		func() (record.Record, error) {
+			return records, nil
+		},
+		func() ([]record.Record, []error) {
 			config, err := i.client.ClusterVersions().Get("version", metav1.GetOptions{})
 			if errors.IsNotFound(err) {
-				return record.Record{}, record.ErrSkipRecord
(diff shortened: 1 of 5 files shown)


smarterclayton:
Captured these during upgrade:

```
x config/pod/openshift-machine-config-operator/machine-config-daemon-blnlt
x config/pod/openshift-machine-config-operator/machine-config-daemon-xnhvz
x config/pod/openshift-machine-config-operator/machine-config-daemon-zsq8n
x config/pod/openshift-machine-config-operator/machine-config-server-2khx5
x config/pod/openshift-machine-config-operator/machine-config-server-tb5x9
```

which correspond to the containers exiting with code 255.  Opened openshift/machine-config-operator#1093 to ensure they have termination messages.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -18,38 +22,82 @@ import (
...
+									records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pods.Items[i].Namespace, pods.Items[i].Name), Item: PodAnonymizer{&pods.Items[i]}})
+								}
+							}
+						}
+					}
+				}
+
+				return records, nil
+			},
+			func() ([]record.Record, []error) {
```
mfojtik: thought you handle nodes in separate PR

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -18,38 +22,82 @@ import (
...
+				if err != nil {
+					return nil, []error{err}
+				}
+				records := make([]record.Record, 0, len(config.Items))
+				for i := range config.Items {
+					records = append(records, record.Record{Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[i].Name), Item: ClusterOperatorAnonymizer{&config.Items[i]}})
+				}
+
+				for _, item := range config.Items {
+					if isFailingOperator(&item) {
```
mfojtik: nit:

```
if !isFailingOperator(&item) { continue }
```

mean less nesting.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -18,38 +22,82 @@ import (
...
+						records = append(records, record.Record{Name: fmt.Sprintf("config/node/%s", nodes.Items[i].Name), Item: NodeAnonymizer{&nodes.Items[i]}})
+					}
+				}
+
+				return records, nil
+			},
+		),
 		func() (record.Record, error) {
 			config, err := i.client.ClusterVersions().Get("version", metav1.GetOptions{})
+			if errors.IsNotFound(err) {
```
mfojtik: hrm, if this returns 404, it means something really bad happened to that cluster, right? don't we want to know?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -219,20 +230,89 @@ func (a ClusterOperatorAnonymizer) Marshal(_ context.Context) ([]byte, error) {
...
+	return runtime.Encode(kubeSerializer, anonymizePod(a.Pod))
+}
+
+func anonymizePod(pod *corev1.Pod) *corev1.Pod {
+	// pods gathered from openshift namespaces and cluster operators are expected to be under our control and contain
+	// no sensitive information
+	return pod
+}
+
+func isFailingPod(pod *corev1.Pod) bool {
```
mfojtik: do we consider Pending pods as failing if the operator report degraded? In some cases (like DS pods), the operator  report unavailable or degraded because it does not see pods rolled out. The reason why the pod is not running might be useful to know. (in most cases it is failed node, but it might be bad image or something)

pkg/record/interface.go
```diff
@@ -30,3 +37,79 @@ type JSONMarshaller struct {
 func (m JSONMarshaller) Marshal(_ context.Context) ([]byte, error) {
 	return json.Marshal(m.Object)
 }
+
+// ErrSkipRecord instructs Collect to ignore the provided recorder.
+var ErrSkipRecord = fmt.Errorf("skip recording")
```
mfojtik: errors.New

pkg/record/interface.go
```diff
@@ -30,3 +37,79 @@ type JSONMarshaller struct {
...
+func Collect(ctx context.Context, recorder Interface, bulkFn func() ([]Record, []error), fns ...func() (Record, error)) error {
+	var errors []string
+	if bulkFn != nil {
+		records, errs := bulkFn()
+		for _, err := range errs {
+			errors = append(errors, err.Error())
+		}
+		for _, record := range records {
+			if err := recorder.Record(record); err != nil {
+				errors = append(errors, fmt.Sprintf("unable to record %s: %v", record.Name, err.Error()))
```
mfojtik: no need for err.Error() ?

pkg/gather/clusterconfig/clusterconfig.go
iNecas: Good point, getting the errors to the reports might be a good thing to do in general. @jhjaggars suggested to keep metadata about the gathering for the parts in separate jsons, and I think that's where this errors should be recorded. Probably out of scope for this PR though.

iNecas: There is a quite a lot of duplicate code here around the `IsNotFound`, which makes me think if that shouldn't rather be in somewhere else?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -219,20 +230,89 @@ func (a ClusterOperatorAnonymizer) Marshal(_ context.Context) ([]byte, error) {
...
+
+type PodAnonymizer struct{ *corev1.Pod }
+
+func (a PodAnonymizer) Marshal(_ context.Context) ([]byte, error) {
+	return runtime.Encode(kubeSerializer, anonymizePod(a.Pod))
+}
+
+func anonymizePod(pod *corev1.Pod) *corev1.Pod {
+	// pods gathered from openshift namespaces and cluster operators are expected to be under our control and contain
+	// no sensitive information
```
iNecas: So far, we have considered the cluster name/url a sensitive information, which is contained in `nodeName` field

iNecas: Another place where the urls are visible are in node selectors and potentially in args for the pod, such as:

```
"args":["start","--apiserver-url=https://api-int.my-cluster.example.com:6443"]
```

For this particular case of the hostnames, given we know the value for the hostame, doing some stupid sed on the raw serialized string might be the safest way to avoid this

pkg/record/interface.go
```diff
@@ -30,3 +37,79 @@ type JSONMarshaller struct {
...
+	}
+	if len(errors) > 0 {
+		sort.Strings(errors)
+		errors = uniqueStrings(errors)
+		return fmt.Errorf("failed to gather cluster infrastructure: %s", strings.Join(errors, ", "))
+	}
+	return nil
+}
+
+func uniqueStrings(arr []string) []string {
```
iNecas: I know this code has already been in, so not requesting to do anything with it here, but wondering, was using https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go considered?

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: It's different per gatherer, and ultimately this error isn't important (the file won't be in the zip, which is ok).

There are a few gatherers that don't call 404, so it shouldn't be generic, and a more complex gatherer won't help understanding what it does (this code is intended to be simple and easy to reason about).

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: This PR has that PR as the first commit.

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: ok.

pkg/record/interface.go
smarterclayton: NOOOO.  I hate errors.New and overlaps with the api errors package.

pkg/record/interface.go
smarterclayton: Yeah

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: Node name and cluster URL won't be sensitive going forward.  Other data in the node object will be.  We have to have node name from alerts and from telemetry.

pkg/record/interface.go
smarterclayton: It's heavy weight and expensive

pkg/record/interface.go
smarterclayton: I'm going to fix this by removing the need for it.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -214,6 +233,27 @@ func (a ClusterOperatorAnonymizer) Marshal(_ context.Context) ([]byte, error) {
 	return runtime.Encode(serializer, a.ClusterOperator)
 }
 
+func isHealthyOperator(operator *configv1.ClusterOperator) bool {
+	for _, condition := range operator.Status.Conditions {
+		switch {
+		case condition.Type == configv1.OperatorDegraded && condition.Status == configv1.ConditionTrue,
+			condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionFalse:
```
eparis: Upgradeable == false seems like a reason to collect

eparis: or do we think that the additionally collected data would not be useful in any way to resolve such a situation?

smarterclayton: I *think* we don't need to look for failing pods in this case.  Fortunately would be easy to fix if we do.

PR #16, opened 2019-09-09
Bug 1745973: Don't reported Degraded on upload error, report UploadDegraded

iNecas:
Given the upload failures are outside of control of the OCP administrator
and they are not fatal for the the functionality of the cluster,
don't set the cluster as degraded for the upload failures.

Separate stats "Uploading" is introduce to indicate the health of the
upload process.

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -24,6 +24,10 @@ import (
 	"github.com/openshift/insights-operator/pkg/controllerstatus"
 )
 
+// How many upload failures in a row we tolerate before starting reporting
+// as UploadDegraded
+const uploadFailuresCountThreshold = 5
+
 type Reported struct {
 	LastReportTime metav1.Time `json:"lastReportTime"`
 }
@@ -114,6 +118,7 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 	var last time.Time
 	var reason string
 	var errors []string
+	var uploadErrorReason, uploadErrorMessage, disabledReason, disabledMessage string
 	allReady := true
 	for i, source := range c.Sources() {
 		summary, ready := source.CurrentStatus()
@@ -129,8 +134,30 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 			klog.Errorf("Programmer error: status source %d %T reported an empty message: %#v", i, source, summary)
 			continue
 		}
-		reason = summary.Reason
-		errors = append(errors, summary.Message)
+
+		degradingFailure := true
+
+		if summary.Operation == controllerstatus.Uploading {
+			if summary.Count < uploadFailuresCountThreshold {
+				klog.V(4).Infof("Number of last upload failures %d lower than threshold %d. Not marking as degraded.", summary.Count, uploadFailuresCountThreshold)
+				degradingFailure = false
+			} else {
+				klog.V(4).Infof("Number of last upload failures %d exceeded than threshold %d. Marking as degraded.", summary.Count, uploadFailuresCountThreshold)
+			}
+			uploadErrorReason = summary.Reason
+			uploadErrorMessage = summary.Message
+			// NotAuthorized is a special case where we want to disable the operator
+			if isNotAuthorizedReason(summary.Reason) {
+				degradingFailure = false
+				disabledReason = summary.Reason
+				disabledMessage = summary.Message
+			}
+		}
+		if degradingFailure {
+			reason = summary.Reason
+			errors = append(errors, summary.Message)
+		}
+
 		if last.Before(summary.LastTransitionTime) {
 			last = summary.LastTransitionTime
 		}
@@ -148,14 +175,9 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 		sort.Strings(errors)
 		errorMessage = fmt.Sprintf("There are multiple errors blocking progress:\n* %s", strings.Join(errors, "\n* "))
 	}
-	var disabledMessage string
 	if !c.configurator.Config().Report {
+		disabledReason = "Disabled"
 		disabledMessage = "Health reporting is disabled"
-	}
-	// NotAuthorized is a special case where we want to disable the operator
-	if reason == "NotAuthorized" {
-		disabledMessage = errorMessage
-		errorMessage = ""
 	}
 
 	existing = existing.DeepCopy()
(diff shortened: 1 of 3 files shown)


mfojtik:
@iNecas the upload failures are fatal for the operator as that is his primary job, to upload data from the clusters... so when that upload is broken, the I would say the operator is indeed degraded...

Now the question is how much retries the operator will do before it puts itself to "degraded" state.

iNecas:
/assign @smarterclayton 
/cc @mfojtik 

Obviously, I have no idea what I'm doing, so suggestion on how to more this PR forward welcome. Just keep in mind that we should solve this before OCP 4.2.

My objectives are:

1. don't let the upload errors influence the degraded state
2. keep the information about the uploader state somewhere for troubleshooting purposes

I would like to add more stuff to telemetry as well, but that's outside of the scope of this PR

pkg/controller/status/status.go
```diff
@@ -347,6 +375,9 @@ func (c *Controller) updateStatus(initial bool) error {
 // OperatorDisabled reports when the primary function of the operator has been disabled.
 const OperatorDisabled configv1.ClusterStatusConditionType = "Disabled"
 
+// Uploading reports true when the operator is successfully uploading
+const Uploading configv1.ClusterStatusConditionType = "Uploading"
```
smarterclayton: We generally try not to create new ClusterStatusConditions.  Putting it on a new condition means it won't show up in any user interface except the generic ones, which may be ok, but I'm not certain yet of the name.  For now, let's clean up the PR as if the name is fine and then we can discuss naming.

pkg/controllerstatus/controllerstatus.go
```diff
@@ -26,6 +26,14 @@ type Simple struct {
 	count   int
 }
 
+type UploaderStatus struct {
```
smarterclayton: If you're using this to do the type check, set an "Operation" field on summary and give that a specific constant that you check for instead.  Wrapping sucks.

pkg/controller/status/status.go
```diff
@@ -219,6 +230,23 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 				Status: configv1.ConditionFalse,
 			})
 		}
+
+		if len(uploadErrorReason) > 0 {
+		    klog.Errorf("=== Uplaoderror")
```
smarterclayton: Debug output?

pkg/controller/status/status.go
```diff
@@ -219,6 +230,23 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
...
+			setOperatorStatusCondition(&existing.Status.Conditions, configv1.ClusterOperatorStatusCondition{
```
smarterclayton: Someone has to reset this condition when uploading is not false.  You'd have to do that in both halves of this else block

pkg/controller/status/status.go
```diff
@@ -129,8 +131,22 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
...
 		}
-		reason = summary.Reason
-		errors = append(errors, summary.Message)
+
+		switch source.(type) {
+		case *controllerstatus.UploaderStatus:
+			uploadErrorReason = summary.Reason
+			uploadErrorMessage = summary.Message
+			// NotAuthorized is a special case where we want to disable the operator
+			if summary.Reason == "NotAuthorized" {
```
smarterclayton: having multiple sets of "uploadError* / disabled*" is going to be confusing, if you add an operation field to the struct use `uploadReason` `uploadMessage` instead.

pkg/controller/operator.go
```diff
@@ -121,7 +121,7 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	// upload results to the provided client - if no client is configured reporting
 	// is permanently disabled, but if a client does exist the server may still disable reporting
 	uploader := insightsuploader.New(recorder, insightsClient, configObserver, statusReporter)
-	statusReporter.AddSources(uploader)
+	statusReporter.AddSources(&uploader.UploaderStatus)
```
smarterclayton: When wrapping is removed remove this

pkg/controller/status/status.go
iNecas: It doesn't fit very well, as there are other reasons for disabling (https://github.com/openshift/insights-operator/pull/16/files#diff-bf2e5e59fb3ef5c42d555b1b4da9c210R168), that are not related to `uploadError*`.

And I don't want to reuse `reason` either, as you might have other reason for `degraded` state, and other for `disabled` or `uploading`

pkg/controller/status/status.go
iNecas: added `removeOperatorStatusCondition` in the else, well as in `isInitializing` case, if that's what this ask is about.

iNecas:
@smarterclayton more feedback welcome

pkg/controller/status/status.go
```diff
@@ -129,8 +132,31 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
...
+
+		degradingFailure := true
+
+		if summary.Operation == controllerstatus.Uploading {
+			if summary.Count < uploadFailuresCountThreshold {
+				klog.V(4).Infof("Number of last upload failures %d lower than threshold %d. Not marking as degraded.", summary.Count, uploadFailuresCountThreshold)
+				degradingFailure = false
+			} else {
+				klog.V(4).Infof("Number of last upload failures %d exceeded than threshold %d. Marking as degraded.", summary.Count, uploadFailuresCountThreshold)
+				degradingFailure = true
```
mfojtik: nit: not needed

pkg/controller/status/status.go
mfojtik: make this a helper `IsNotAuthorizedReason(summary.Reason)`

mfojtik:
2 nits, LGTM otherwise

pkg/controller/status/status.go
iNecas: updated

pkg/controller/status/status.go
iNecas: I wanted to be a bit explicit here, but can do both ways.

iNecas:
I've addressed the last nitpicks and squashed the commits, that should make it ready for merge

smarterclayton:
What is going on - this only reports once?

smarterclayton:
Spawned https://bugzilla.redhat.com/show_bug.cgi?id=1753755 to cover the bug in logging in CI env.

PR #18, opened 2019-09-10
Bug 1750665 - increase default interval

iNecas:
For now, every 2h should be enough. We will look into makint the
interval smaller once the infrastructure settles down a bit.


iNecas:
Alternative PR that includes this change, but also allows updating the config via configmap is here https://github.com/openshift/insights-operator/pull/19

PR #19, opened 2019-09-10
Bug 1750665 - increase default interval and use configmap

iNecas:
For now, every 2h should be enough. We will look into making the
interval smaller once the infrastructure settles down a bit.

In order to be able to override the operator configuration, the pod
configuration has been converted to a configmap.

### DIFF
--- a/config/pod.yaml
+++ b/config/pod.yaml
@@ -1,8 +1,0 @@
-kind: GenericOperatorConfig
-apiVersion: operator.openshift.io/v1alpha1
-leaderElection:
-  disable: true
-interval: "10m"
-storagePath: /var/lib/insights-operator
-endpoint: https://cloud.redhat.com/api/ingress/v1/upload
-impersonate: system:serviceaccount:openshift-insights:gather
--- a/manifests/04_configmap.yaml
+++ b/manifests/04_configmap.yaml
@@ -1,0 +1,25 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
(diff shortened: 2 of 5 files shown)


config/pod.yaml
```diff
@@ -1,8 +0,0 @@
-kind: GenericOperatorConfig
-apiVersion: operator.openshift.io/v1alpha1
-leaderElection:
-  disable: true
-interval: "10m"
```
smarterclayton: This PR can just be setting this to 2h

manifests/04_configmap.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-insights
+  name: insights-operator-config
+data:
+  operator-config.yaml: |
```
smarterclayton: This breaks CI.  Why did you remove the default config?

smarterclayton: You shouldn't override this, instead the config map is a customization.  Our default ootb is set via config/pod.yaml so that way you don't have to do this.

iNecas: I've mainly inspired by https://github.com/openshift/cluster-authentication-operator.

I agree with config maps to serve mainly for overrides. Is there some good operator that would be following this pattern of having static file as defautls + config file from configmaps for overrides?

smarterclayton:
/hold

the other approach was fine

PR #20, opened 2019-09-11
Refresh the token periodically in case of a change

iNecas:


pkg/config/configobserver/configobserver.go
```diff
@@ -50,6 +50,9 @@ func New(defaultConfig config.Controller, kubeClient kubernetes.Interface) *Cont
 
 func (c *Controller) Start(ctx context.Context) {
 	wait.Until(func() {
+		if err := c.retrieveToken(); err != nil {
+			klog.Warningf("Unable to retrieve initial token config: %v", err)
```
mfojtik: s/initial//

mfojtik: @smarterclayton should this propagate to operator status (unavailable/degraded=true ?)

iNecas: Done

iNecas: We'll get notified about that at the next attempt to actually use the pull secret.

iNecas:
Bump @mfojtik?

PR #21, opened 2019-09-20
Bug 1753755 - fix overriding the refresh interval in CI

iNecas:
The interval for uploading was refreshed, but the period module
was not watching the config change, so no new reports were sent
in CI.

Also, this fixes propagating error messages when the interval is set too
low (using `duration, err := ...` was causing the top level `err` value
to be shadowed.)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -5,6 +5,11 @@ import (
 	"time"
 )
 
+type Configurator interface {
+	Config() *Controller
+	ConfigChanged() (<-chan struct{}, func())
+}
+
 // Controller defines the standard config for this operator.
 type Serialized struct {
 	Report      bool   `json:"report"`
--- a/pkg/config/configobserver/configobserver.go
+++ b/pkg/config/configobserver/configobserver.go
@@ -109,42 +109,36 @@ func (c *Controller) retrieveConfig() error {
 	if err != nil {
 		if errors.IsNotFound(err) {
 			klog.V(4).Infof("Support secret does not exist")
-			err = nil
+			return nil
 		} else if errors.IsForbidden(err) {
 			klog.V(2).Infof("Operator does not have permission to check support secret: %v", err)
-			err = nil
+			return nil
 		} else {
-			err = fmt.Errorf("could not check support secret: %v", err)
+			return fmt.Errorf("could not check support secret: %v", err)
 		}
 	}
-	if secret != nil {
-		if username, ok := secret.Data["username"]; ok {
-			nextConfig.Username = string(username)
-		}
-		if password, ok := secret.Data["password"]; ok {
-			nextConfig.Password = string(password)
-		}
-		if endpoint, ok := secret.Data["endpoint"]; ok {
-			nextConfig.Endpoint = string(endpoint)
-		}
-		nextConfig.Report = len(nextConfig.Endpoint) > 0
 
-		if intervalString, ok := secret.Data["interval"]; ok {
-			duration, errp := time.ParseDuration(string(intervalString))
-
-			if errp == nil && duration < time.Minute {
-				errp = fmt.Errorf("too short")
-			}
-			if errp == nil {
-				nextConfig.Interval = duration
-			} else {
-				err = fmt.Errorf("insights secret interval must be a duration (1h, 10m) greater than or equal to one minute: %v", errp)
-				nextConfig.Report = false
-			}
-		}
+	if username, ok := secret.Data["username"]; ok {
+		nextConfig.Username = string(username)
 	}
-	if err != nil {
-		return err
+	if password, ok := secret.Data["password"]; ok {
+		nextConfig.Password = string(password)
+	}
+	if endpoint, ok := secret.Data["endpoint"]; ok {
+		nextConfig.Endpoint = string(endpoint)
+	}
+	nextConfig.Report = len(nextConfig.Endpoint) > 0
+
+	if intervalString, ok := secret.Data["interval"]; ok {
+		duration, err := time.ParseDuration(string(intervalString))
+
+		if err == nil && duration < time.Minute {
+			err = fmt.Errorf("too short")
+		}
+		if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/config/configobserver/configobserver.go
```diff
@@ -130,14 +130,15 @@ func (c *Controller) retrieveConfig() error {
 		nextConfig.Report = len(nextConfig.Endpoint) > 0
 
 		if intervalString, ok := secret.Data["interval"]; ok {
-			duration, err := time.ParseDuration(string(intervalString))
-			if err == nil && duration < time.Minute {
-				err = fmt.Errorf("too short")
+			duration, errD := time.ParseDuration(string(intervalString))
```
iNecas: The `err` variable was not propagated, so we lost it outside of this block

smarterclayton: style: we use `errp` in Go for these sorts of things (single lowercase letter after `err`)

smarterclayton: You would lose it outside the block anyway - errD is local.

iNecas: Yes, so was `err` in this case and that's why we were loosing the message: the point of this change, to make it local explicitly, and set it to the global `err` here https://github.com/openshift/insights-operator/pull/21/files#diff-e217a6a94e2f38d09ce7df512a06b266R141.

Btw. we would be able to avoid the `errp` workaround if we used the "return early" style, which makes me think why it's not that way in the first place and the only reason

iNecas: Thinking about it a bit more, it looks using "return early" approach would avoid us having to care about the scope of `err` and simplifies the method a bit.

pkg/config/configobserver/configobserver.go
```diff
@@ -109,41 +109,36 @@ func (c *Controller) retrieveConfig() error {
...
-				err = fmt.Errorf("insights secret interval must be a duration (1h, 10m) greater than or equal to one minute: %v", err)
-				nextConfig.Report = false
-			}
-		}
+	if username, ok := secret.Data["username"]; ok {
+		nextConfig.Username = string(username)
 	}
-	if err != nil {
-		return err
+	if password, ok := secret.Data["password"]; ok {
```
mfojtik: do we allow empty passwords/username/endpoint? is there any validation done?

pkg/controller/periodic/periodic.go
```diff
@@ -140,3 +147,37 @@ func (c *Controller) processNextWorkItem() bool {
 
 	return true
 }
+
+func (c *Controller) updateConfig() bool {
+	cfg := c.configurator.Config()
+	if c.interval != cfg.Interval {
```
mfojtik: nit: can you rename `c.interval` to `c.currentInterval` ?

pkg/controller/periodic/periodic.go
```diff
@@ -140,3 +147,37 @@ func (c *Controller) processNextWorkItem() bool {
...
+		c.interval = cfg.Interval
+		klog.V(4).Infof("Gathering interval set to %s", cfg.Interval)
+		return true
+	} else {
```
mfojtik: else not needed.

pkg/controller/periodic/periodic.go
```diff
@@ -140,3 +147,37 @@ func (c *Controller) processNextWorkItem() bool {
...
+	}
+}
+
+// Runs loop for checking on configuration change to check for periodic-related changes.
+// Returns a channel that it writes to, when the config change influenced the controller.
+func (c *Controller) watchConfig(stopCh <-chan struct{}) chan struct{} {
+	configCh := make(chan struct{}, 1)
+	globalConfigCh, cancelFn := c.configurator.ConfigChanged()
+
+	watchChanges := func() {
```
mfojtik: nit: `go func() { }`

(you don't need this func stored, right?)

pkg/config/config.go
```diff
@@ -5,6 +5,11 @@ import (
 	"time"
 )
 
+type Configurator interface {
+	Config() *Controller
+	ConfigChanged() (<-chan struct{}, func())
```
mfojtik: nit: docs... please document the second return value

pkg/controller/periodic/periodic.go
iNecas: I know this, but want to name the func to describe what's it's purpose. Would like to keep it that way if you don't mind

pkg/config/configobserver/configobserver.go
```diff
@@ -109,41 +109,36 @@ func (c *Controller) retrieveConfig() error {
...
+	if endpoint, ok := secret.Data["endpoint"]; ok {
+		nextConfig.Endpoint = string(endpoint)
+	}
+	nextConfig.Report = len(nextConfig.Endpoint) > 0
+
+	if intervalString, ok := secret.Data["interval"]; ok {
+		duration, err := time.ParseDuration(string(intervalString))
+
+		if err == nil && duration < time.Minute {
+			err = fmt.Errorf("too short")
```
alexandrevicenzi: duration too short explains a bit better

pkg/config/configobserver/configobserver.go
```diff
@@ -109,41 +109,36 @@ func (c *Controller) retrieveConfig() error {
...
+		}
+		if err != nil {
+			return fmt.Errorf("insights secret interval must be a duration (1h, 10m) greater than or equal to one minute: %v", err)
```
alexandrevicenzi: the `(1h, 10m)` does not really fit well in the message

alexandrevicenzi:
@iNecas it seems that [this PR](https://github.com/openshift/insights-operator/pull/40/commits) made a fix to reset the interval and could probably have fixed this issue.

Would you mind to take a look in it?

alexandrevicenzi:
This was fixed in #40, we retested in 4.3 and it worked as expected.

PR #23, opened 2019-10-02
Updated README, step-by-step installation procedure

tisnik:


README.md
```diff
@@ -10,7 +10,17 @@ This cluster operator gathers anonymized system configuration and reports it to
 
 ## Building
 
-To build the operator, install Go 1.11 or above and run:
+To build the operator, install Go 1.11 or above. Don't forget to set `$GOPATH` environment variable properly.
+
+Then retrieve sources of insights-operator:
+
+    go get -u github.com/openshift/insights-operator
```
mfojtik: This is not very common way to get the sources ;-) We usually do `git clone`.

natiiix: Wouldn't this make importing more convenient?

tisnik:
@mfojtik I'd change the 'go get' to the following command
'git clone https://github.com/openshift/insights-operator $GOPATH/src/github.com/openshift/insights-operator'

Would it be better?

tisnik:
Finally we updated that repo to use Go modules, so this PR is no longer relevant :)

PR #27, opened 2019-10-18
Collect the proxy config

eparis:
We anonomize basically everything, but at least we'd know if it was set.


eparis:
@smarterclayton @jwforres @deads2k

mfojtik:
@iNecas do we anonymize pod env vars when we dumping pods?

iNecas:
> @iNecas do we anonymize pod env vars when we dumping pods?

Let me grep it for you :)

https://github.com/openshift/insights-operator/blob/cc22b344078d4de4cfd17d11d68193c48dc03c1d/pkg/gather/clusterconfig/clusterconfig.go#L347

smarterclayton:
We do not, they're our pods.  I don't think we have a concrete requirement yet to anonymize pods.

PR #28, opened 2019-10-21
Switch to Kubernetes 1.16 and Go Modules

natiiix:
- Switch from Kubernetes 1.14 to 1.16
- Switch from [Glide](https://glide.sh/) to Go modules (from `glide.yaml` and `glide.lock` to `go.mod` and `go.sum` respectively)

New versions of dependencies that already switched to Go modules made it too difficult to split these changes into two separate PRs, hence a single PR for both.

The Go module requires Golang version 1.12, which also matches the base image used in the `Dockerfile` and majority of other OpenShift Go modules as far as I have seen so far.


iNecas:
Tested locally as well as via cluster-bot, and everything just worked e2e. I also personally prefer go modules over other solutions, so :+1: for this move (that we need anyway before we could move forward with reusing the code for `oc inspect`

/lgtm

PR #34, opened 2019-10-25
Use github.com/eparis/urlhash to anonymize URLs

eparis:
It should give us stable values of ip addresses and URLs but not actually leak
customer data. This isn't simple to work backwards to the original IP because
the values are salted with a random string.

Input:  https://my.openshift-console-ingress.api.console.customer.com/path/to/something/openshift
Output: https://05.openshift-console-5bd922e.api.console.ee8ca6dd.com/b5bf/39/ca74813cb/openshift

Input:  https://127.0.0.1
Output: https://04c.7e9.7e9.b4b

Input:  https://127.0.0.2
Output: https://04c.7e9.7e9.b35

This also attempts to keep things like 'http://' or 'https://'.
It also will retain certain words in URLs, like openshift, or console.
It also represent ip addresses a bit wierdly, since you get 3 hex characters, but you can make
useful comparisions between hashed values.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/certifi/gocertifi v0.0.0-20180905225744-ee1a9a0726d2 // indirect
 	github.com/coreos/bbolt v1.3.3 // indirect
 	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f // indirect
+	github.com/eparis/urlhash v0.0.0-20191022152723-5838cc59d83c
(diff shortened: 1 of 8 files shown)


eparis:
A cluster with this running can be found at 5318290/cd 626bd7fe-6591-4ace-b863-1dca7442b040/ for someone who wants to look at the results..

eparis:
/retest
failure that aws said the ignition bucket didn't exist

alexandrevicenzi:
@eparis this PR has conflicts and is stale for a long time, Do you care to fix it? Is there a BZ issue related to this?

tremes:
@eparis any update on this pls?

PR #35, opened 2019-10-30
gather: consider restarting pods as not healthy

mfojtik:
In case a container is restarting or has restart count higher than zero, consider it not healthy and copy it to the report in case the related operator is degraded.


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -361,6 +361,9 @@ func isHealthyPod(pod *corev1.Pod) bool {
 		if status.State.Terminated != nil && status.State.Terminated.ExitCode != 0 {
 			return false
 		}
+		if status.RestartCount > 0 {
```
sferich888: Can we make this restart could configurable, or time-bound? 
1-2 restarts may not be bad (10-100 are). But it also may be dependant on time, was this 10 min ago or 4 days ago.

PR #36, opened 2019-10-30
gather: include events in report

mfojtik:
Collect all events for `openshift-` namespaces that are not older than 1 hour (for starter). The events are compacted to minimize the size in payload. If we still see big file, we can tweak the 1h interval to 30 min.

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -4,13 +4,16 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/sets"
 	kubescheme "k8s.io/client-go/kubernetes/scheme"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -26,6 +29,10 @@ import (
 var (
 	serializer     = scheme.Codecs.LegacyCodec(configv1.SchemeGroupVersion)
 	kubeSerializer = kubescheme.Codecs.LegacyCodec(corev1.SchemeGroupVersion)
+
+	// maxEventTimeInterval represents the "only keep events that are maximum 1h old"
+	// TODO: make this dynamic like the reporting window based on configured interval
+	maxEventTimeInterval = 1 * time.Hour
 )
 
 type Gatherer struct {
@@ -59,12 +66,12 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error 
 			for i := range config.Items {
 				records = append(records, record.Record{Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[i].Name), Item: ClusterOperatorAnonymizer{&config.Items[i]}})
 			}
+			namespaceEventsCollected := sets.NewString()
 
 			for _, item := range config.Items {
 				if isHealthyOperator(&item) {
 					continue
 				}
-				failingPods := []corev1.Pod{}
 				for _, namespace := range namespacesForOperator(&item) {
 					pods, err := i.coreClient.Pods(namespace).List(metav1.ListOptions{})
 					if err != nil {
@@ -75,20 +82,20 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error 
 						if isHealthyPod(&pods.Items[i]) {
 							continue
 						}
-						failingPods = append(failingPods, pods.Items[i])
 						records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pods.Items[i].Namespace, pods.Items[i].Name), Item: PodAnonymizer{&pods.Items[i]}})
 					}
-				}
-				networkEvents, err := handlePendingPodsNetworkEvents(failingPods, i.coreClient)
-				if err != nil {
-					klog.V(2).Infof("Unable to gather network events: %v", err)
-					continue
-				}
-				for i := range networkEvents {
-					records = append(records, record.Record{Name: fmt.Sprintf("config/events/%s/network", networkEvents[i].Namespace), Item: EventAnonymizer{&networkEvents[i]}})
+					if namespaceEventsCollected.Has(namespace) {
+						continue
+					}
+					namespaceRecords, errs := i.gatherNamespaceEvents(namespace)
+					if len(errs) > 0 {
+						klog.V(2).Infof("Unable to collect events for namespace %q: %#v", namespace, errs)
+						continue
+					}
+					records = append(records, namespaceRecords...)
+					namespaceEventsCollected.Insert(namespace)
 				}
 			}
-
 			return records, nil
 		},
 		func() ([]record.Record, []error) {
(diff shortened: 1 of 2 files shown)


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -63,18 +64,29 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
...
 					}
 					for i := range pods.Items {
 						if isHealthyPod(&pods.Items[i]) {
 							continue
 						}
+						failingPods = append(failingPods, pods.Items[i])
 						records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pods.Items[i].Namespace, pods.Items[i].Name), Item: PodAnonymizer{&pods.Items[i]}})
 					}
 				}
+				networkEvents, err := handlePendingPodsNetworkEvents(failingPods, i.coreClient)
```
smarterclayton: just gather all events in the namespace since the last ~ interval

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -185,6 +197,36 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
...
+	}
+	podNamespaces := sets.NewString()
+	for _, pod := range pods {
+		if podNamespaces.Has(pod.Namespace) {
+			continue
+		}
+		podNamespaces.Insert(pod.Namespace)
+	}
+	for _, namespace := range podNamespaces.List() {
+		namespaceEvents, err := eventsGetter.Events(namespace).List(metav1.ListOptions{})
```
smarterclayton: We probably need to bound the runtime of this call in the future - i.e. if it hangs it hangs the gather loop.  That does not have to be done in this PR.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -63,18 +64,29 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
...
 						records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pods.Items[i].Namespace, pods.Items[i].Name), Item: PodAnonymizer{&pods.Items[i]}})
 					}
 				}
+				networkEvents, err := handlePendingPodsNetworkEvents(failingPods, i.coreClient)
+				if err != nil {
+					klog.V(2).Infof("Unable to gather network events: %v", err)
+					continue
+				}
+				for i := range networkEvents {
+					records = append(records, record.Record{Name: fmt.Sprintf("config/events/%s/network", networkEvents[i].Namespace), Item: EventAnonymizer{&networkEvents[i]}})
```
smarterclayton: Prefer this just be events, not network related.  We'll want other types of events no matter what.

mfojtik:
@smarterclayton @iNecas the collected events is ~70kb decompressed, it is much smaller compressed and it is 70kb I would pay in gold to have to debug degraded clusters as the events provide timeline when problems happen (we can detect broken CNI for example).

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -67,6 +73,7 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
 					pods, err := i.coreClient.Pods(namespace).List(metav1.ListOptions{})
 					if err != nil {
 						klog.V(2).Infof("Unable to find pods in namespace %s for failing operator %s", namespace, item.Name)
+						continue
```
mfojtik: the `pods` can be nil on error and this will then panic...

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -25,6 +28,9 @@ import (
 var (
 	serializer     = scheme.Codecs.LegacyCodec(configv1.SchemeGroupVersion)
 	kubeSerializer = kubescheme.Codecs.LegacyCodec(corev1.SchemeGroupVersion)
+
+	// maxEventTimeInterval represents the "only keep events that are maximum 1h old"
+	maxEventTimeInterval = 1 * time.Hour
```
smarterclayton: Technically this needs to be dynamic based on the interval - that would be the right thing to do.  Can you just add a TODO here for `TODO: make this dynamic like the reporting window based on configured interval`?  The interval is available via a getter for the reporter function anyway.

mfojtik: todo added

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -76,9 +84,41 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
 					}
 				}
 			}
-
 			return records, nil
 		},
+		func() ([]record.Record, []error) {
+			events, err := i.coreClient.Events(corev1.NamespaceAll).List(metav1.ListOptions{})
```
smarterclayton: NOOOOOOOO

smarterclayton: You can't do this.

mfojtik: ok, i will do this per-namespace :-)

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: good catch

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -185,6 +204,40 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
...
+	}
+	events, err := i.coreClient.Events(namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return nil, []error{err}
+	}
+	// filter the event list to only recent events
+	now := time.Now()
+	var filteredEventIndex []int
+	for i := range events.Items {
+		if events.Items[i].LastTimestamp.Time.Before(now.Add(-1 * maxEventTimeInterval)) {
```
smarterclayton: do this calculation earlier (not in the loop)

smarterclayton: also `- maxEventTimeInterval` is clearer

mfojtik: i don't know what you mean @smarterclayton with the calculation, in that loop I only pick events that are matching that condition (maxEventTimeInterval) then I build slice with the size  matching the number of filtered events and fill that slice with that events... I think it is already efficient?

smarterclayton: Don't do `now.Add(-1 *x)` in the loop.  Calculate that up front `oldestEvent := ...` and then make this `...Time.Before(oldestEvent)`

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -185,6 +204,40 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
...
+	for i, index := range filteredEventIndex {
+		compactedEvents.Items[i] = CompactedEvent{
+			Namespace:     events.Items[index].Namespace,
+			LastTimestamp: events.Items[index].LastTimestamp.Time,
+			Reason:        events.Items[index].Reason,
+			Message:       events.Items[index].Message,
+		}
+	}
+	sort.Slice(compactedEvents.Items, func(i, j int) bool {
+		return compactedEvents.Items[i].LastTimestamp.Before(compactedEvents.Items[i].LastTimestamp)
```
smarterclayton: Typo - you're not actually sorting (you use `i` in both places)

mfojtik: good catch, fixed...

iNecas:
@mfojtik mind updating the openshift docs as well with details on this change? https://github.com/openshift/openshift-docs/blob/master/modules/insights-operator-what-information-is-collected.adoc

mfojtik:
@iNecas https://github.com/openshift/openshift-docs/pull/17816

PR #40, opened 2019-11-03
gather: Collect some prometheus metrics for the cluster

smarterclayton:
Grab running alerts from /federate using the gather client. More
metrics will be added later.

### DIFF
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -132,7 +132,7 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	// the gatherers periodically check the state of the cluster and report any
 	// config to the recorder
 	configPeriodic := clusterconfig.New(gatherConfigClient, gatherKubeClient.CoreV1(), metricsClient)
-	periodic := periodic.New(s.Interval, recorder, map[string]gather.Interface{
+	periodic := periodic.New(configObserver, recorder, map[string]gather.Interface{
 		"config": configPeriodic,
 	})
 	statusReporter.AddSources(periodic.Sources()...)
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -9,7 +9,9 @@ import (
 	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/pkg/version"
 	"k8s.io/client-go/rest"
 
@@ -77,6 +79,27 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	if len(s.Impersonate) > 0 {
 		gatherKubeConfig.Impersonate.UserName = s.Impersonate
 	}
+
+	// the metrics client will connect to prometheus and scrape a small set of metrics
+	// TODO: the oauth-proxy and delegating authorizer do not support Impersonate-User,
+	//   so we do not impersonate gather
+	metricsGatherKubeConfig := rest.CopyConfig(controller.KubeConfig)
+	metricsGatherKubeConfig.CAFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
+	metricsGatherKubeConfig.NegotiatedSerializer = scheme.Codecs
+	metricsGatherKubeConfig.GroupVersion = &schema.GroupVersion{}
+	metricsGatherKubeConfig.APIPath = "/"
+	metricsGatherKubeConfig.Host = "https://prometheus-k8s.openshift-monitoring.svc:9091"
+
+	// If we fail, it's likely due to the service CA not existing yet. Warn and continue,
+	// and when the service-ca is loaded we will be restarted.
+	var metricsClient rest.Interface
+	metricsRESTClient, err := rest.RESTClientFor(metricsGatherKubeConfig)
+	if err != nil {
+		klog.Warningf("Unable to load metrics client, no metrics will be collected: %v", err)
+	} else {
+		metricsClient = metricsRESTClient
+	}
+
 	gatherKubeClient, err := kubernetes.NewForConfig(gatherProtoKubeConfig)
 	if err != nil {
 		return err
@@ -108,7 +131,7 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 
 	// the gatherers periodically check the state of the cluster and report any
 	// config to the recorder
-	configPeriodic := clusterconfig.New(gatherConfigClient, gatherKubeClient.CoreV1())
+	configPeriodic := clusterconfig.New(gatherConfigClient, gatherKubeClient.CoreV1(), metricsClient)
 	periodic := periodic.New(s.Interval, recorder, map[string]gather.Interface{
 		"config": configPeriodic,
 	})
--- a/pkg/controller/periodic/periodic.go
+++ b/pkg/controller/periodic/periodic.go
@@ -12,27 +12,32 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/workqueue"
 
+	"github.com/openshift/insights-operator/pkg/config"
 	"github.com/openshift/insights-operator/pkg/controllerstatus"
 	"github.com/openshift/insights-operator/pkg/gather"
 	"github.com/openshift/insights-operator/pkg/record"
 )
 
-type Controller struct {
-	interval time.Duration
+type Configurator interface {
+	Config() *config.Controller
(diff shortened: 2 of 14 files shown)


iNecas:
What's the idea behind this compared to using thanos data?

smarterclayton:
Users will be able to disable sending data.  To get subs enabled, you'll take the latest insights-operator grab on disk and upload it to OCM, which will contain metrics and alerts.

Also, telemetry can fail, and not every metric we want to gather will be sent via telemetry.  Certain high cardinality metrics will never be gathered.

must-gather is going to grab a larger slice, that'll come after I get this sorted out.

smarterclayton:
This breaking (probably) because oauth-proxy does not support impersonation because it’s using a too old version of the authenticating filters.  Will be fixing that.

smarterclayton:
```
-rw-r----- 0/0            7613 2019-11-05 00:25 config/metrics
-rw-r----- 0/0             594 2019-11-05 00:25 config/network
-rw-r----- 0/0             480 2019-11-05 00:25 config/oauth
-rw-r----- 0/0            5986 2019-11-05 00:25 config/pod/openshift-insights/insights-operator-86477cc47f-pzgfb
-rw-r----- 0/0             315 2019-11-05 00:25 config/proxy
-rw-r----- 0/0            1641 2019-11-05 00:25 config/version
-rw-r----- 0/0            1530 2019-11-05 00:25 events/openshift-insights

cat ... / 0aa35200-7fe7-4e16-b4f8-74938324cbd7 / 

# TYPE ALERTS untyped
ALERTS{alertname="ClusterOperatorDegraded",alertstate="pending",condition="Degraded",endpoint="metrics",instance="10.0.159.189:9099",job="cluster-version-operator",name="insights",namespace="openshift-cluster-version",pod="cluster-version-operator-9767b7fdf-8qtvf",reason="PeriodicGatherFailed",service="cluster-version-operator",severity="critical",prometheus="openshift-monitoring/k8s",prometheus_replica="prometheus-k8s-0"} 1 1572913529213
ALERTS{alertname="ClusterOperatorDown",alertstate="pending",endpoint="metrics",instance="10.0.159.189:9099",job="cluster-version-operator",name="insights",namespace="openshift-cluster-version",pod="cluster-version-operator-9767b7fdf-8qtvf",service="cluster-version-operator",severity="critical",version="0.0.1-2019-11-04-235332",prometheus="openshift-monitoring/k8s",prometheus_replica="prometheus-k8s-0"} 1 1572913529213
ALERTS{alertname="KubeClientErrors",alertstate="pending",instance="10.0.159.189:6443",job="apiserver",severity="warning",prometheus="openshift-monitoring/k8s",prometheus_replica="prometheus-k8s-0"} 1 1572913532039
```

smarterclayton:
@deads2k can you make cases for additional metrics right now?  This can't do queries, but it can gather raw metrics (queries would be a separate call that is harder to read, I think we can do that later).

smarterclayton:
The crash here is that service-ca.crt isn't available for the first few minutes (before service-ca-operator comes up).  We're going to have to loop.

smarterclayton:
Ok, this should be working now.  Had to fix gathering to configure its interval dynamically.

PTAL, still want to review final metrics list and format with Michal and David.

pkg/controller/operator.go
```diff
@@ -77,6 +79,27 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
 	if len(s.Impersonate) > 0 {
 		gatherKubeConfig.Impersonate.UserName = s.Impersonate
 	}
+
+	// the metrics client will connect to prometheus and scrape a small set of metrics
+	// TODO: the oauth-proxy and delegating authorizer do not support Impersonate-User,
+	//   so we do not impersonate gather
+	metricsGatherKubeConfig := rest.CopyConfig(controller.KubeConfig)
+	metricsGatherKubeConfig.CAFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
```
mfojtik: making this const in `pkg/cmd/start/start.go` but hardcoding here seems strange

pkg/controller/operator.go
```diff
@@ -77,6 +79,27 @@ func (s *Support) Run(controller *controllercmd.ControllerContext) error {
...
+	metricsGatherKubeConfig.GroupVersion = &schema.GroupVersion{}
+	metricsGatherKubeConfig.APIPath = "/"
+	metricsGatherKubeConfig.Host = "https://prometheus-k8s.openshift-monitoring.svc:9091"
+
+	// If we fail, it's likely due to the service CA not existing yet. Warn and continue,
+	// and when the service-ca is loaded we will be restarted.
+	var metricsClient rest.Interface
+	metricsRESTClient, err := rest.RESTClientFor(metricsGatherKubeConfig)
+	if err != nil {
+		klog.Warningf("Unable to load metrics client, no metrics will be collected: %v", err)
```
mfojtik: should this be degraded condition?

pkg/controller/periodic/periodic.go
```diff
@@ -101,15 +105,31 @@ func (c *Controller) Gather() {
 }
 
 func (c *Controller) periodicTrigger(stopCh <-chan struct{}) {
```
mfojtik: note to myself... we should really stop using stopCh and use context.Context :-)

/cc @iNecas 
(sounds like a good intern job :-)

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -36,24 +37,41 @@ var (
...
 var reInvalidUIDCharacter = regexp.MustCompile(`[^a-z0-9\-]`)
 
 func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error {
 	return record.Collect(ctx, recorder,
+		func() ([]record.Record, []error) {
+			if i.metricsClient == nil {
+				return nil, nil
+			}
+			data, err := i.metricsClient.Get().AbsPath("federate").
+				Param("match[]", "ALERTS").
```
mfojtik: @smarterclayton i wish this is more visible (thinking like defaults.go?)

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -36,24 +37,41 @@ var (
...
+				Param("match[]", "etcd_object_counts").
```
mfojtik: you also promised the cluster_installer? ;-)

pkg/controller/operator.go
smarterclayton: Consts leaking across packages is bad.  In that case it was just referencing the same variable twice.  Here, eh.

pkg/controller/operator.go
smarterclayton: I'd probably say no, because if service-ca.crt is broken everything else should break, and we don't need prom metrics.  I may have metrics gracefully handle down prometheus (was looking at the errors we get during startup) simply to avoid confusing the issue.  Insights operator isn't responsible for reporting health of other components today.

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: Once we have a full list I agree it should be in its own package.

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: Yup, need a list.

pkg/controller/operator.go
mfojtik: @smarterclayton it is not other component health, I think if insights operator can't get data it should write to report, it should be communicated and visible to user/admin (not just logged...) maybe warning event? anyway, can be a follow up

pkg/gather/clusterconfig/clusterconfig.go
smarterclayton: Will do the three for now, then take feedback tomorrow and Friday as you suggested.

pkg/controller/operator.go
smarterclayton: I made even metrics retrieval errors survivable, and instead have them record something in the insights data (comments are allowed in the file format).

mfojtik:
/lgtm

lets get this soak tested sooner and we can add to the list as we go

pkg/gather/clusterconfig/clusterconfig.go
deads2k: - [ ] aggregator_unavailable_apiservice_count
- [ ] aggregator_unavailable_apiserver_gauge
- [ ] work queue metrics for kube-controller-manager, kube-scheduler, and kube-apiserver processes (can we do pattern matches?)

@mfojtik we should generally collect CSRs

pkg/gather/clusterconfig/clusterconfig.go
tremes: Just out of curiosity. Why do we use the federate API instead of e.g `api/v1/query`?

PR #42, opened 2019-11-08
api to metrics service changed from prometheus to legacyregistry

SamuelStuchly:
Api to metrics service has been changed suddenly without our knowledge, causing metrics to not be registered anymore, therefore this change is needed to resolve the issue.

### DIFF
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -19,10 +19,10 @@ import (
 	knet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/client-go/pkg/version"
 	"k8s.io/client-go/transport"
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
 
 	"k8s.io/klog"
-
-	"github.com/prometheus/client_golang/prometheus"
 
 	configv1 "github.com/openshift/api/config/v1"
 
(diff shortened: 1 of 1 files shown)


tisnik:
Nice catch @SamuelStuchly.
Note for the rest of the team: in the future we'd need to test the /metrics endpoint to ensure that any new change won't break this functionality.

iNecas:
I've tested this, but I'm still not sure why it got broken (seems like related to rebase to k8s 1.16). 

@mfojtik any idea? Michal also: can you confirm other operators using the same approach are still reporting the metrics ok (such as cluster-auth-operator https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/version/version.go - I don't see the metric openshift_cluster_authentication_operator_build_info in nightly clusters).

I've seen `image_registry_operator_storage_reconfigured_total` from https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/metrics/metrics.go, but this operator is running their own metrics server.

What is the recommended way doing this anyway?

tisnik:
@iNecas yes Ivan, it was caused by rebasing k8s. Previously the `prometheus` package was used to expose metrics, now the `k8s.io/component-base/metrics/legacyregistry` is used instead. Please note that we don'n expose `/metrics` manually - it is done inside k8s (can be grepped as "/metrics" string is used on just one place). So when `prometheus` was encapsulated by `legacyregistry`, the call of `prometheus.Register` or `prometheus.MustRegister` is no-op, as `/metrics` endpoint is handled by another package.

mfojtik:
1) you need e2e test so this won't break again
2) i think using legacyregistry is discouraged, see https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/20190404-kubernetes-control-plane-metrics-stability.md
3) you might need your custom registry for metrics long term, so you can guarantee "stable" metrics and possibly deprecate metrics (that is what upstream wants)
4) short term, you can use legacyregistry, but somebody should work out how to wire custom registry and replace legacy.

tisnik:
@mfojtik yes Michal, we talked about #1, it really needs to be tested

iNecas:
/lgtm, as I've seen this working.

For the custom metrics, I would like to see some other openshift operator going over this process, to be able to inspire from it: the OpenShift core team will probably be more effective finding the right way of doing this that we do.

PR #43, opened 2019-11-11
WIP Use own registry for insights-operator metrics

iNecas:
Let's see if that fixes the issue with missing metrics.


iNecas:
Doesn't work unless one runs it's own prometheus server. Closing WIP.

PR #47, opened 2019-11-14
add more metrics to gather

deads2k:
Follow on to https://github.com/openshift/insights-operator/pull/40 that adds metrics interesting to group-b.

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -63,8 +63,24 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error 
 			}
 			data, err := i.metricsClient.Get().AbsPath("federate").
 				Param("match[]", "ALERTS").
+				// aggregator metrics indicate how apiservers are performing as viewed from the kube-apiserver
+				Param("match[]", "aggregator_unavailable_apiservice_count").
+				Param("match[]", "aggregator_unavailable_apiserver_gauge").
+
+				// let's us know how many of each kind of object we have.  This give information about usage
 				Param("match[]", "etcd_object_counts").
+
 				Param("match[]", "cluster_installer").
+
+				// these work queue metrics can be used to determine how individual control loops are performing
+				Param("match[]", "workqueue_adds_total").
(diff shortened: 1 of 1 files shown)


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -63,8 +63,24 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
 			}
 			data, err := i.metricsClient.Get().AbsPath("federate").
 				Param("match[]", "ALERTS").
+				// aggregator metrics indicate how apiservers are performing as viewed from the kube-apiserver
+				Param("match[]", "aggregator_unavailable_apiservice_count").
+				Param("match[]", "aggregator_unavailable_apiserver_gauge").
```
smarterclayton: do these have alerts?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -63,8 +63,24 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
...
+
+				// let's us know how many of each kind of object we have.  This give information about usage
 				Param("match[]", "etcd_object_counts").
+
 				Param("match[]", "cluster_installer").
+
+				// these work queue metrics can be used to determine how individual control loops are performing
+				Param("match[]", "workqueue_adds_total").
```
smarterclayton: what's the cardinality of these?  if it's big I want to start with the ones that indicate problems, vs debugging them all.

pkg/gather/clusterconfig/clusterconfig.go
deads2k: > do these have alerts?

Not sure actually.  Not in the kas-o where I would expect them to be defined.

pkg/gather/clusterconfig/clusterconfig.go
deads2k: > what's the cardinality of these? if it's big I want to start with the ones that indicate problems, vs debugging them all.

potentially one per control loop.  Do we make it easy to say: all these metrics from the kas, kcm, ks, oas, and ocm?  Those have the most value, but it also picks up most of the cardinality.

PR #48, opened 2019-11-15
insightsclient: Report the amount of data written

smarterclayton:
Allow us to judge from logs how much we are sending.


smarterclayton:
Tagging so I can judge David’s pr

PR #54, opened 2019-11-27
Added dummy e2e target for release 4.2

quarckster:


iNecas:
@mfojtik mind doing similar magic here, as you did in https://github.com/openshift/insights-operator/pull/55 (I don't have enough permissions to mangle the labels here, which is good thing in general :)

mfojtik:
@iNecas this need valid bug

PR #56, opened 2019-11-27
Bug 1767719: Fix the event reporting in insights operator

mfojtik:
The previous backport only backported the network events capture, not all events. This will revert the previous backport and pick the correct commits from release-4.3.


mfojtik:
self-lgtm as this is only correcting already approved backport.

**patch manager**: can we get this in early in the cycle? it is improving our observability into failed telemetry clusters and we kind of needed this yesterday.

PR #50, opened 2019-11-18
initial commit for tests

lina-nikiforova:
Output locally
```
$ go test ./test/integration
ok      github.com/openshift/insights-operator/test/integration 10.722s
```

PR #58, opened 2019-11-29
Bug 1788118: initial commit for tests for release 4.3

lina-nikiforova:
same as https://github.com/openshift/insights-operator/pull/50


iNecas:
Since it's targeting 4.3, the BZ needs to be cloned as targeted to different release. @lina-nikiforova I expect you have rights to do it in BZ, let me know if you don't

PR #61, opened 2019-12-16
added TestOptOutOptIn and moved some code to functions

lina-nikiforova:
Added new test to verify that insights operator opt-in/opt-out works

PR #62, opened 2019-12-17
Add Pavel Tisnovsky into list of repo owners

tisnik:


tisnik:
@lina-nikiforova I assume that after https://github.com/openshift/insights-operator/pull/61 will be merged, the tests will be ok (passing), right? :)

lina-nikiforova:
@tisnik I believe so, we will see in #61 as the tests were restarted :)

lina-nikiforova:
@tisnik yes, it passed there :)

PR #63, opened 2020-01-06
Bug 1788112: override node selector (4.3)

iNecas:
To ensure the deployment works properly in case of default node selector
is set.

(cherry picked from commit 53cfc74e5b5edf7fb3fc192080ee4d7299d5ad96)


alexandrevicenzi:
@iNecas could you fix the BZ issue to pass tide?

PR #65, opened 2020-01-08
Bug 1797563: added TestUnreachableHost

lina-nikiforova:
added test to verify insights operator reports degraded after 5 unsuccessful upload attempts

### DIFF
--- a/test/integration/bugs_test.go
+++ b/test/integration/bugs_test.go
@@ -12,7 +12,7 @@ import (
 // https://bugzilla.redhat.com/show_bug.cgi?id=1750665
 func TestDefaultUploadFrequency(t *testing.T) {
 	// delete any existing overriding secret
-	err := kubeClient.CoreV1().Secrets("openshift-config").Delete("support", &metav1.DeleteOptions{})
+	err := clientset.CoreV1().Secrets("openshift-config").Delete("support", &metav1.DeleteOptions{})
 
 	// if the secret is not found, continue, not a problem
 	if err != nil && err.Error() != `secrets "support" not found` {
@@ -20,14 +20,14 @@ func TestDefaultUploadFrequency(t *testing.T) {
 	}
 
 	// restart insights-operator (delete pods)
-	RestartInsightsOperator(t)
+	restartInsightsOperator(t)
 
 	// check logs for "Gathering cluster info every 2h0m0s"
-	CheckPodsLogs(t, kubeClient, "Gathering cluster info every 2h0m0s")
+	checkPodsLogs(t, clientset, "Gathering cluster info every 2h0m0s")
 }
 
+// TestUnreachableHost checks if insights operator reports "degraded" after 5 unsuccessful upload attempts
 // https://bugzilla.redhat.com/show_bug.cgi?id=1745973
-// Check if insights operator reports "degraded" after 5 unsuccessful upload attempts
 func TestUnreachableHost(t *testing.T) {
 	// Replace the endpoint to some not valid url.
 	// oc -n openshift-config create secret generic support --from-literal=endpoint=http://localhost --dry-run -o yaml | oc apply -f - -n openshift-config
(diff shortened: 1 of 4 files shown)


test/integration/main_test.go
```diff
@@ -39,6 +40,33 @@ func KubeClient() (result *kubernetes.Clientset) {
...
+	obj := map[string]interface{}{}
+	err = json.Unmarshal(data, &obj)
+	if err != nil {
+		panic(err.Error())
+	}
+	return obj
+}
+
+func IsOperatorDegraded(t *testing.T, config map[string]interface{}) bool {
+	status := config["status"].(map[string]interface{})
```
iNecas: Could we use the plumbing around cluster operator API to deserialize the data, so that we don't need to deal with the `map[string]interface{}` which is pretty painful to work with.

I guess code in https://github.com/openshift/insights-operator/blob/fac81f249a23886f976a8869e38978b2a2355275/pkg/controller/status/status.go#L346 might be a good example on how to use it.

test/integration/main_test.go
```diff
@@ -39,6 +40,33 @@ func KubeClient() (result *kubernetes.Clientset) {
...
+	data, err := k.RESTClient().Get().AbsPath("/apis/config.openshift.io/v1/clusteroperators/insights").DoRaw()
+	obj := map[string]interface{}{}
+	err = json.Unmarshal(data, &obj)
+	if err != nil {
+		panic(err.Error())
+	}
+	return obj
+}
+
+func IsOperatorDegraded(t *testing.T, config map[string]interface{}) bool {
```
tisnik: Probably a nitpick, but it would be IMHO better to have a docstring (comment) for all exported function. In this case, it is possible we don't have to export these functions outside current test package?

test/integration/main_test.go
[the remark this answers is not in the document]
lina-nikiforova: Agreed but I couldn't find it easily. I will take a look and also at Ivan's comment about this part.

test/integration/bugs_test.go
```diff
@@ -22,3 +25,56 @@ func TestDefaultUploadFrequency(t *testing.T) {
 	// check logs for "Gathering cluster info every 2h0m0s"
 	CheckPodsLogs(t, kubeClient, "Gathering cluster info every 2h0m0s")
 }
+
+// https://bugzilla.redhat.com/show_bug.cgi?id=1745973
+// Check if insights operator reports "degraded" after 5 unsuccessful upload attempts
```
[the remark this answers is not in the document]
lina-nikiforova: thanks! :)

test/integration/main_test.go
lina-nikiforova: well, I see I was wrong about exported functions and it's not needed here as everything is in one package (at least now).

test/integration/main_test.go
lina-nikiforova: thank you! I found a way to remove this ugly code :)

lina-nikiforova:
I implemented the suggestions but pretty sure it will fail in the CI. I will work on it a bit later.

lina-nikiforova:
Fails because of the known issue https://bugzilla.redhat.com/show_bug.cgi?id=1793051
Tests should be restarted when it's resolved.

test/integration/main_test.go
```diff
@@ -32,24 +35,84 @@ func KubeClient() (result *kubernetes.Clientset) {
...
+		}
+	}
+	t.Log("Operator is degraded")
+	return true
+}
+
+func isOperatorDisabled(t *testing.T, operator *configv1.ClusterOperator) bool {
+	statusConditions := operator.Status.Conditions
+
+	for _, condition := range statusConditions {
```
tisnik: this might be refactored as the code is almost the same as in `isOperatorDegraded`.

tisnik: but ok for tests :)

PR #66, opened 2020-01-09
insightsclient: include error message when we are unable to build request

mfojtik:
We see 'Unable to report: unable to build request to connect to Insights server' message reported as reason for degraded insights operator, but we don't know why (probably bad token?).

@iNecas @smarterclayton is there a reason we muted the error we use for degraded message?


iNecas:
The only reason I could think of is if the error contained some sensitive data, but it doesn't look like that's the case

PR #70, opened 2020-01-27
Collect certificates

martinkunc:
The task should add collection of Certificate features, which could be used to verify their validity in further stages.


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -223,6 +226,20 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
 			}
 			return []record.Record{{Name: "config/proxy", Item: ProxyAnonymizer{config}}}, nil
 		},
+		func() ([]record.Record, []error) {
+			requests, err := i.certClient.CertificateSigningRequests().List(metav1.ListOptions{})
+			if errors.IsNotFound(err) {
+				return nil, nil
```
tisnik: I know this is used in previous gathering functions, but is the `nil` handled properly later?

pkg/gather/clusterconfig/csr.go
```diff
@@ -0,0 +1,185 @@
...
+	certificatesv1b1api "k8s.io/api/certificates/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/json"
+)
+
+type CSRAnonymizer struct {
+	*certificatesv1b1api.CertificateSigningRequest
+}
+
+func (a CSRAnonymizer) Marshal(_ context.Context) ([]byte, error) {
```
tisnik: nitpick - all public methods should have proper docstring

pkg/gather/clusterconfig/csr.go
```diff
@@ -0,0 +1,185 @@
...
+	"time"
+
+	"github.com/openshift/insights-operator/pkg/utils"
+	"k8s.io/api/certificates/v1beta1"
+	certificatesv1b1api "k8s.io/api/certificates/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/json"
+)
+
+type CSRAnonymizer struct {
```
tisnik: nitpick - all public data types should have proper docstring

pkg/gather/clusterconfig/csr.go
```diff
@@ -0,0 +1,185 @@
...
+	// CSR in a PEM
+	// parse only first PEM block
+	block, _ := pem.Decode(r.Spec.Request)
+	if block == nil {
+		// unable to decode CSR: missing block
+		return
+	}
+	csr, err := x509.ParseCertificateRequest(block.Bytes)
+	if err != nil {
+		return
```
tisnik: don't we need to log these issues?

pkg/gather/clusterconfig/clusterconfig.go
martinkunc: Fortunately returning nil in a slice will return empty slice and wont fail. It looks even empty Record wont fail as its just a structure. I have just verified both.

pkg/gather/clusterconfig/csr.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	c.Spec.Request = &CsrFeatures{}
+	c.Spec.Request.ValidSignature = err == nil
+	c.Spec.Request.Subject = anonymizePkxName(csr.Subject)
+
+	c.Spec.Request.SignatureAlgorithm = csr.SignatureAlgorithm.String()
+	c.Spec.Request.PublicKeyAlgorithm = csr.PublicKeyAlgorithm.String()
+	c.Spec.Request.DNSNames = utils.Map(csr.DNSNames, anonymizeURL)
+	c.Spec.Request.EmailAddresses = utils.Map(csr.EmailAddresses, anonymizeURL)
+	ipsl := []string{}
```
tisnik: if we know `len(csr.IPAddresses)`, you can allocate an array, but it's up to you

pkg/utils/strings.go
```diff
@@ -0,0 +1,10 @@
+package utils
+
+// Map applies each of functions to passed slice
+func Map(it []string, fn func(string) string) []string {
```
tisnik: in the future, we might create a common utils package outside operator

alexandrevicenzi: in the future, Go may add map-reduce functions, but there already libraries like this available.

iNecas:
Great job on adding tests as well.

I'm a bit concerned about the possible amounts of the records, so I would suggest filtering out `Approved` CSRs, as they would probably not be very valuable for troubleshooting.

@smarterclayton any suggestion for somebody close to CSRs to give us some hinds on what kinds of CSRs types are useful for troubleshooting?

methlock:
Well, we don't need the actual certificate, we are just using that to extract expiration date. If the cert is expired or will be expired on week or so.

I'm totally fine if the data extraction will be part of the collection mechanism.

So basically we just need info if the cert is approved and expiration date.

If it is still too much data, we can move all logic to the operator and collect just unapproved, expired and those, which will expire in 7 (we can discuss this number) days.

martinkunc:
I can either filter all, or just gather certain data from Approved. Do you have any idea about numbers here ?

One idea was that we might start thinking about maybe different profiles of small and bigger size.
While one could be really statistical, the other could be all-in-one for detailed investigation.

Which way would be the best for now ?

iNecas:
My main question is, whether we care (have a use-case) for the approved ones, or we are interested into some specific cases?

methlock:
As far as I know, we don't care about approved ones (at least from the Insights perspective).

smarterclayton:
Can you describe in more detail why we are collecting this?  What scenarios will this enable us to detect (or which bugs are we seeing in the field that this info would return)?

methlock:
The unapproved or expired certificates were a root issue in some customer cases. 
Hence it seems useful to collect this info in advance and possibly show it to the customer.

Are we missing something?

pkg/gather/clusterconfig/csr.go
```diff
@@ -0,0 +1,264 @@
...
+func WithTime(t time.Time) FilterOptFunc {
+	return func(o *FilterOpt) {
+		o.time = t
+	}
+}
+
+func IncludeCSR(c *CSRAnonymizedFeatures, opts ...FilterOptFunc) bool {
+	opt := &FilterOpt{time: time.Now()}
+	for _, o := range opts {
+		o(opt)
```
alexandrevicenzi: these variables names are not good, e, t, o, i, etc, it could be a little more descriptive. In this case it could be `optFn`.

pkg/gather/clusterconfig/csr.go
alexandrevicenzi: these comments are still pending, would be nice to add some docs in the whole file.

pkg/utils/strings.go
```diff
@@ -0,0 +1,10 @@
+package utils
+
```
smarterclayton: In general in go we do not create utility packages like this.  This is a trivial method and would inline privately in a package.  Utility packages cause dependency creep and attract unrelated use cases, which cause coupling.

Please move this to a private method at the package call site.

martinkunc: Thank you for the comment. It makes sense, I will remove this in some further PR.

PR #71, opened 2020-02-04
Add alexandrevicenzi as code owner

alexandrevicenzi:


alexandrevicenzi:
@iNecas @tisnik one check is pending, is there a way to fix or ignore it?

PR #81, opened 2020-02-25
add new test TestClusterDefaultNodeSelector

lina-nikiforova:
New test to cover https://bugzilla.redhat.com/show_bug.cgi?id=1782151

### DIFF
--- a/test/integration/bugs_test.go
+++ b/test/integration/bugs_test.go
@@ -81,3 +81,44 @@ func TestUnreachableHost(t *testing.T) {
 	})
 	t.Log(errDegraded)
 }
+
+// https://bugzilla.redhat.com/show_bug.cgi?id=1782151
+func TestClusterDefaultNodeSelector(t *testing.T) {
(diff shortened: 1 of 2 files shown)


test/integration/bugs_test.go
```diff
@@ -81,3 +81,45 @@ func TestUnreachableHost(t *testing.T) {
 	})
 	t.Log(errDegraded)
 }
+
+// https://bugzilla.redhat.com/show_bug.cgi?id=1782151
+func TestClusterDefaultNodeSelector(t *testing.T) {
+	// set default selctor of node-role.kubernetes.io/worker
+	schedulers, err := configClient.Schedulers().List(metav1.ListOptions{})
+	t.Log(schedulers)
```
alexandrevicenzi: this should probably be after the if err, otherwise, if failed you are going to log nil or an empty list.

lina-nikiforova: I am removing it, it will not be very useful message anyway.

PR #84, opened 2020-03-04
Fix reporting duration error and add tests

martinkunc:
The retrieve config was doing checks for interval, but the actual error, but it declared new err variable, which was shadowing the original err.
That means any set inside of the block, when the block is left was discarded and err returned to original value from outer scope.
In case of too short interval (under 1 m) the outer 'err' was nil, even thought inner scope set it to "too short" it was discarded and function returned err = nil.


alexandrevicenzi:
@martinkunc where is the BZ related to this?

pkg/config/configobserver/configobserver_test.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+			if tt.expConfig != nil && !reflect.DeepEqual(tt.expConfig, c.config) {
+				t.Fatalf("The test expected config doesn't fit actual config.\nExpected: %v Actual: %v", tt.expConfig, c.config)
+			}
+		})
+	}
+
+}
+
+var (
```
alexandrevicenzi: const perhaps?

pkg/config/configobserver/configobserver_test.go
```diff
@@ -0,0 +1,146 @@
...
+			secs := tt.config
+			// setup mock responses for secretes by secret name
+			kube.CoreV1().(*corefake.FakeCoreV1).Fake.AddReactor("get", "secrets",
+				func(action clienttesting.Action) (handled bool, ret runtime.Object, err error) {
+					actionName := ""
+					if getAction, ok := action.(clienttesting.GetAction); ok {
+						actionName = getAction.GetName()
+					}
+
+					//log.Printf("namespace %s resource: %s verb %s Name %s", action.GetNamespace(), action.GetResource(), action.GetVerb(), actionName)
```
alexandrevicenzi: remove or uncomment

pkg/config/configobserver/configobserver_test.go
```diff
@@ -0,0 +1,146 @@
...
+			expErrS := ""
+			if tt.expErr != nil {
+				expErrS = tt.expErr.Error()
+			}
+			errS := ""
+			if err != nil {
+				errS = err.Error()
+			}
+			if expErrS != errS {
+				t.Fatalf("The test expected error doesn't fit actual error.\nExpected: %s Actual: %s", tt.expErr, err)
```
alexandrevicenzi: match is a better word then fit.

pkg/config/configobserver/configobserver_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/client-go/kubernetes"
+	clsetfake "k8s.io/client-go/kubernetes/fake"
+	corefake "k8s.io/client-go/kubernetes/typed/core/v1/fake"
+	"k8s.io/klog"
+
+	clienttesting "k8s.io/client-go/testing"
+)
+
+func TestChangeSupportConfig(t *testing.T) {
+	var cases = []struct {
```
alexandrevicenzi: I would rather create 2 tests, each test with only one purpose. It's more clear to read and understand. But up to you.

PR #86, opened 2020-03-19
Collecting config maps

martinkunc:
This is adding gathering of the configmaps from the openshift-config namespace.
Would something like this be reasonable ?
It is collecting the values, except in case it is a PEM, which replaces with dummy PEM structure.


pkg/utils/strings.go
```diff
@@ -1,10 +0,0 @@
-package utils
```
alexandrevicenzi: What is the relation of this change to config maps?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -543,3 +562,34 @@ func (i *Gatherer) ClusterVersion() *configv1.ClusterVersion {
...
+// ConfigMapAnonymizer implements serialization of configmap
+// and potentially anonymizes if it is a certificate
+type ConfigMapAnonymizer struct{ v []byte }
+
+// Marshal implements serialization of Node with anonymization
+func (a ConfigMapAnonymizer) Marshal(_ context.Context) ([]byte, error) {
+	return []byte(anonymizeConfigMap(a.v)), nil
+}
+
+func anonymizeConfigMap(dv []byte) string {
```
alexandrevicenzi: What if there's another type of sensitive data in the config map? Like a user/password?

I'm asking because I'm not sure what could be inside `openshift-config`.

martinkunc: I havent noticed anything except the certificates, which are being anonymized. If there is anything, I would need to handle it case by case, but so far I dont know about anything else.

pkg/utils/strings.go
martinkunc: Clayton wanted to remove the package in a comment to previous review.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -543,3 +562,34 @@ func (i *Gatherer) ClusterVersion() *configv1.ClusterVersion {
...
+	anonymizedPemBlock := `-----BEGIN CERTIFICATE-----
+YW5vbnltaXplZA==
```
iNecas: Why specifically this string?

iNecas:
@martinkunc do you plan adding some tests as well?

iNecas:
One thing I'm not sure about is how valuable the anonymized ca files are? If nothing else, I would recommend making it more explicit that they are anonymized (as it's not very obvious right now).

I've also tested this with adding some additoina config-map (example taken from https://docs.openshift.com/container-platform/4.3/nodes/scheduling/nodes-scheduler-default.html and I can confirm the config-map got collected).

Another idea to start thinking about, as more things are getting added to connect, is some meta/configuration about the kinds of data being collected, and allowing to tweak the configuration if needed. You can have a look at https://github.com/redhataccess/insights-client/blob/master/docs/insights-client.conf.5 to see, how insights for rhel is dealing with that. - not in scope of this PR though

pkg/gather/clusterconfig/clusterconfig.go
martinkunc: I wanted it to still be a valid PEM, the "YW5vbnltaXplZA==" is base64 of  "anonymized" in case anyone would be trying to decode it.

martinkunc:
I was hoping that I could add some tests later. Should I bring some ?

martinkunc:
By saving those ca/pem files I was hoping only to indicate that they exists and are parseable in the cluster. Just in some edge case the value of configmap would be invalid (not a pem). But maybe it doesn't make sense ?

pkg/gather/clusterconfig/clusterconfig.go
iNecas: My first reaction was to try using `openssl` against it and it failed terribly. So just mentioning `ANONYMIZED`, or `XXXXXXXXXXXXXX` would be feel intuitive to me

pkg/utils/strings.go
alexandrevicenzi: But this could be another PR as it's not related to this change.

pkg/gather/clusterconfig/clusterconfig.go
alexandrevicenzi: Not sure if someone else could help us here because if we end up collecting a password it will be too late to fix something.

alexandrevicenzi:
@iNecas @martinkunc any updates on the comments? Otherwise we can merge it.

lina-nikiforova:
I tested this by provisioning cluster using this PR and config maps got collected.
All CAs are anonymized like
```
-----BEGIN CERTIFICATE-----
ANONYMIZED
-----END CERTIFICATE-----
```

martinkunc:
I just wanted to add some basic unit tests, or I can also do it later. This was the last thing here.

alexandrevicenzi:
@martinkunc add some unit test as @lina-nikiforova and @iNecas suggested.

PR #87, opened 2020-03-19
Support for specific http proxy for the service

martinkunc:
This change allows to override proxy passed with Env using support secret.
It uses golang build-in proxy settings, including noproxy logic.

### DIFF
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -12,11 +12,11 @@ import (
 	"net"
 	"net/http"
 	"net/textproto"
+	"net/url"
 	"os"
 	"strconv"
 	"time"
 
-	knet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/client-go/pkg/version"
 	"k8s.io/client-go/transport"
 	"k8s.io/component-base/metrics"
@@ -40,6 +40,7 @@ type Client struct {
 
 type Authorizer interface {
 	Authorize(req *http.Request) error
+	NewSystemOrConfiguredProxy() func(*http.Request) (*url.URL, error)
 }
 
 type ClusterVersionInfo interface {
@@ -88,10 +89,10 @@ func getTrustedCABundle() (*x509.CertPool, error) {
 	return certs, nil
 }
 
-func clientTransport() http.RoundTripper {
-	// default transport, proxy from env
+// clientTransport creates new http.Transport with either system or configured Proxy
+func clientTransport(authorizer Authorizer) http.RoundTripper {
 	clientTransport := &http.Transport{
-		Proxy: knet.NewProxierWithNoProxyCIDR(http.ProxyFromEnvironment),
+		Proxy: authorizer.NewSystemOrConfiguredProxy(),
 		DialContext: (&net.Dialer{
 			Timeout:   30 * time.Second,
 			KeepAlive: 30 * time.Second,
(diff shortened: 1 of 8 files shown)


pkg/insights/insightsclient/insightsclient_test.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+			if (tc.ProxyURL == "" && url != nil) ||
+				(len(tc.ProxyURL) > 0 && (url == nil || tc.ProxyURL != url.String())) {
+				t.Fatalf("Unexpected value of Proxy Url. Test %s Expected Url %s Received Url %s", tc.Name, tc.ProxyURL, url)
+			}
+		})
+	}
+}
+
+// func TestEnvNoSpecProxy(t *testing.T) {
```
alexandrevicenzi: Any reason to leave commented?

pkg/insights/insightsclient/insightsclient_test.go
```diff
@@ -0,0 +1,143 @@
...
+		{
+			Name:       "Env set, specific proxy set https",
+			EnvValues:  map[string]interface{}{"HTTPS_PROXY": "envsecproxy.to"},
+			RequestURL: "https://google.com",
+			HttpConfig: config.HTTPConfig{HTTPProxy: "specsecproxy.to"},
+			ProxyURL:   "http://specsecproxy.to",
+		},
+		{
+			Name:       "Env set, specific proxy set noproxy, request without noproxy",
+			EnvValues:  map[string]interface{}{"HTTPS_PROXY": "envsecproxy.to", "NO_PROXY": "envnoproxy.to"},
```
tremes: Not sure about `HTTP/S_PROXY` values, but shouldn't the `NO_PROXY` support multiple values?

martinkunc: Hi, it should, there is just no test for that..

pkg/insights/insightsclient/insightsclient_test.go
martinkunc: I forgot the comment here, I will remove it. The test is moved above.

martinkunc:
I was hoping that I could add some tests later. Should I bring some ?

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -113,6 +122,39 @@ func clientTransport() http.RoundTripper {
...
+		NoProxy:    c.NoProxy,
+	}
+	// The golang ProxyFunc seems to have NoProxy already built in
+	return func(req *http.Request) (*url.URL, error) {
+		return proxyConfig.ProxyFunc()(req.URL)
+	}
+}
+
+// FromConfig is setting HttpProxy from HttpConfig in support secret, if it is used
+func FromConfig(c config.HTTPConfig) func(req *http.Request) (*url.URL, error) {
```
smarterclayton: `ProxyFromConfig` as name

smarterclayton: Actually, these two methods are confusingly named in general.  This should probably be `NewProxyFromConfig()` in the line above.

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -40,6 +43,7 @@ type Client struct {
 
 type Authorizer interface {
 	Authorize(req *http.Request) error
+	HTTPConfig() config.HTTPConfig
```
smarterclayton: Why is the authorizer not handling the proxy setup?  It already has all the config info.

pkg/insights/insightsclient/insightsclient.go
martinkunc: I removed them when I moved the logic to authorizer and removed options to simplify a bit.

pkg/insights/insightsclient/insightsclient.go
martinkunc: I moved it there. If you ment something like a Roundtripper with dynamic injectable Proxy, I can prepare it as well, but at first it seemed a bit too much code.

alexandrevicenzi:
@martinkunc we need this in 4.3 as well.

martinkunc:
I tried to squeeze the changes, which invalidated the lgtms, sorry.

PR #89, opened 2020-04-03
Bug 1820633: Cherry-pick release 4.4 Support for specific http proxy for the service

martinkunc:
Cherry-pick support for specific http proxy
Obviously I wasn't successful with squashing the commits here, so I had to cherry pick them manually this time.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,11 @@ require (
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5
 	github.com/tmc/grpc-websocket-proxy v0.0.0-20190109142713-0ad062ec5ee5 // indirect
+<<<<<<< HEAD
 	golang.org/x/net v0.0.0-20200114155413-6afb5195e5aa // indirect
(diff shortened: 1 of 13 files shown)


alexandrevicenzi:
It seems to me that there's more code than it supposed to be, is this correct?

If you try to apply the [patch](https://patch-diff.githubusercontent.com/raw/openshift/insights-operator/pull/87.patch) directly does it work?

Also, the title of this PR is misleading, we are backporting proxy to 4.4.

martinkunc:
@jwforres Could I ask for approving the cherry-pick with cherry-pick-approved label please ?

eparis:
/lgtm cancel
why are the insights test failing? seems that needs to be addressed.

eparis:
So far I see base_test.go failing 4 times in the history and passing 3 times.  how are things in master? Is your flake rate 50% there?

martinkunc:
@eparis We have the same situation in master with this test. I already have an internal case to investigate it.

PR #99, opened 2020-05-04
Bug 1838969: pkg/record/diskrecorder/diskrecorder: add json extension 

vrutkovs:
`Marshalable` interface is extended to provide a file extension. It 
is being appended to the files when these are written to disk. Openshift/kube object serializers are writing json data, configmap / id serializers are writing plain bytes etc.
If marshaller returns empty string no extension is appended.

Ref: CCXDEV-1366

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -527,12 +527,22 @@ func (r RawByte) Marshal(_ context.Context) ([]byte, error) {
 	return r, nil
 }
 
+// GetExtension returns extension for "id" file - none
+func (r RawByte) GetExtension() string {
+	return ""
+}
+
 // Raw is another simplification of marshalling from string
 type Raw struct{ string }
 
 // Marshal returns raw bytes
 func (r Raw) Marshal(_ context.Context) ([]byte, error) {
 	return []byte(r.string), nil
+}
+
+// GetExtension returns extension for raw marshaller
+func (r Raw) GetExtension() string {
+	return ""
 }
 
 // Anonymizer returns serialized runtime.Object without change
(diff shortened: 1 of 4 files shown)


iNecas:
Since this is a backward incompatible chnage that could break the processing tooling, I would like to ask @mbacovsky to discuss possibility to handle this change in the processing tool to prepare it for this change.

martinkunc:
I think because we don't have everyting just json, maybe it would be easier to name data record with suffix. Instead of setting it for each Flush in diskrecorder, change names of files in Gather methods - at places where we create actual json content, for example this line:
https://github.com/openshift/insights-operator/blob/master/pkg/gather/clusterconfig/clusterconfig.go#L138

martinkunc:
Would you mind if we would in scope of this case change:
/config/id file from pure guid content to something like { "cluster_id":"guid" }
and in gathering config maps, instead of saving them to individual files collect one /config/configmaps in a json way (with PEM blocks in json as well).

If you wouldn't mind the change in processing logic. I can prepare the change and then everything should be just json.

vrutkovs:
>change:
/config/id file from pure guid content to something like { "cluster_id":"guid" }

Just `"guid"` is a valid JSON though, just needs to be quoted. Would that work for you?

>collect one /config/configmaps in a json way

I've never seen configmaps being collected - mind pointing me to the code which does that?

martinkunc:
Hi, the string of cluster id guild in config/id makes sense for me as its valid.
The example of configmaps is

For when it contains PEM certificate here: https://github.com/openshift/insights-operator/blob/master/docs/insights-archive-sample/config/configmaps/admin-kubeconfig-client-ca/ca-bundle.crt

For when it contains other data here: https://github.com/openshift/insights-operator/tree/master/docs/insights-archive-sample/config/configmaps/openshift-install

Code collecting it is at: https://github.com/openshift/insights-operator/blob/master/pkg/gather/clusterconfig/clusterconfig.go#L229

But the main question is whether @mbacovsky is fine with changes.

martinkunc:
Ok, Martin is good with the change. I can prepare the change. Ideally if we could wait with this until it will be prepared to have everything just json.

vrutkovs:
/cc @martinkunc

Reworked this to have per-marshaller granularity - openshift/kube serializers write json data, others don't have extension appended (could have `txt` though), PTAL

vrutkovs:
/retest

Flake

martinkunc:
@vrutkovs Wouldn't you like something based on options better ? Like https://play.golang.org/p/w4g7ImJmIRN

vrutkovs:
Options seems interesting, but do we plan to pass more options there? Seems a bit too complex for this usecase though.

I don't mind rewriting this to use variadic vars though

martinkunc:
Ok, if Pavel is also fine with this, lets do it this way.
/lgtm

martinkunc:
@vrutkovs Could you please also add Bug: with a BZ link (create one) for 4.5,4.4.4.3 and 4.2. The process is described here: https://url.corp.redhat.com/ocp-backport-process. We need to merge it to old versions to ease us automatic merges of future cases. Thank you!

martinkunc:
@vrutkovs Would it be possible also to create BZ issues for backporting ? IT should be just clones to Bug 1838969. Issue for 4.4. should depend on this (issue for 4.5), issue for 4.3 should depend on issue for 4.4 and issue for 4.3 should depend on issue for 4.2.

/lgtm

PR #100, opened 2020-05-05
Bug 1832232: Gather image registry config

mfojtik:
The config is valuable as we can drive deprecation of some fields inside it, but also valuable to get insights into how image registry is configured in clusters.

/cc @bparees @dmage


bparees:
images is more about the external registries they are using.  we probably also want `configs.imageregistry.operator.openshift.io` which is the operator config resource for how the internal registry is configured (e.g. what storage it uses) if we're not already getting that somehow.

mfojtik:
@bparees where is the client/api for that resource?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -528,6 +550,30 @@ func (a FeatureGateAnonymizer) Marshal(_ context.Context) ([]byte, error) {
...
+type ImageRegistryAnonymizer struct {
+	*registryv1.Config
+}
+
+// Marshal implements serialization of Ingres.Spec.Domain with anonymization
+func (a ImageRegistryAnonymizer) Marshal(_ context.Context) ([]byte, error) {
+	a.Spec.HTTPSecret = anonymizeString(a.Spec.HTTPSecret)
+	if a.Spec.Storage.S3 != nil {
+		a.Spec.Storage.S3.Bucket = anonymizeString(a.Spec.Storage.S3.Bucket)
+		a.Spec.Storage.S3.KeyID = anonymizeString(a.Spec.Storage.S3.KeyID)
```
dmage, line 563: I don't know what's the purpose of anonymizeString, should it be used for anything that may identify the customer? RegionEndpoint may contain address of an in-house S3-compatible storage.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -528,6 +550,30 @@ func (a FeatureGateAnonymizer) Marshal(_ context.Context) ([]byte, error) {
...
+	}
+	if a.Spec.Storage.Azure != nil {
+		a.Spec.Storage.Azure.AccountName = anonymizeString(a.Spec.Storage.Azure.AccountName)
+		a.Spec.Storage.Azure.Container = anonymizeString(a.Spec.Storage.Azure.Container)
+	}
+	if a.Spec.Storage.GCS != nil {
+		a.Spec.Storage.GCS.Bucket = anonymizeString(a.Spec.Storage.GCS.Bucket)
+		a.Spec.Storage.GCS.ProjectID = anonymizeString(a.Spec.Storage.GCS.ProjectID)
+		a.Spec.Storage.GCS.KeyID = anonymizeString(a.Spec.Storage.GCS.KeyID)
+	}
```
dmage, line 573: Swift:
* AuthURL
* Container
* Domain
* DomainID
* Tenant
* TenantID
* probably RegionName

bparees:
> @bparees where is the client/api for that resource?

https://github.com/openshift/api/blob/master/imageregistry/v1/types.go
https://github.com/openshift/client-go/tree/master/imageregistry

dmage:
@mfojtik you have a typo in the commit message, you grab config.imageregistry.operator.openshift.io/cluster

pkg/gather/clusterconfig/clusterconfig.go
mfojtik: yes, anynymize will hide the data (replace with `x`)

mfojtik:
@dmage fixed

mfojtik:
/hold 

for CCX team approval

iNecas:
Going to play a bad :policeman: here, but mind adding some tests to https://github.com/openshift/insights-operator/blob/master/pkg/gather/clusterconfig/clusterconfig_test.go, having some example input and anonymized output would make me sleep better at night :)

martinkunc:
@mfojtik Could you you please add a bugzilla case to know to which versions it would have to be backported ?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -528,6 +550,40 @@ func (a FeatureGateAnonymizer) Marshal(_ context.Context) ([]byte, error) {
...
 }
 
+// IngressAnonymizer implements serialization with marshalling
+type ImageRegistryAnonymizer struct {
+	*registryv1.Config
+}
+
+// Marshal implements serialization of Ingres.Spec.Domain with anonymization
+func (a ImageRegistryAnonymizer) Marshal(_ context.Context) ([]byte, error) {
+	a.Spec.HTTPSecret = anonymizeString(a.Spec.HTTPSecret)
```
martinkunc, line 560: What if we would just remove all Storage ? Can here be any other types, like pvc ?
And what about HttpClaim, is it something to remove as well ?

dmage: If we remove Storage, we won't be able to know if someone uses a storage parameter.

mfojtik:
@martinkunc I expect this to be backported to 4.4 at least.

mfojtik:
@iNecas test pushed. PTAL.

pkg/gather/clusterconfig/clusterconfig_test.go
```diff
@@ -88,6 +92,147 @@ func TestConfigMapAnonymizer(t *testing.T) {
...
+							Bucket:    "bucket",
+							Region:    "region",
+							ProjectID: "foo",
+							KeyID:     "bar",
+						},
+					},
+				},
+			},
+			evalOutput: func(t *testing.T, obj *imageregistryv1.Config) {
+				if obj.Spec.Storage.GCS.Bucket == "foo" {
```
iNecas, line 191: `s/foo/bucket`, or maybe just checking it contains noething but `/x*/` would be better (and reusable for all keys to prevent this kind of copy/paste problems)

mfojtik: it does not matter :-) this just checks the Bucket value was modified by anonymyzer.

iNecas: And if you read it carefully, you will find this test actually would catch even if the anonymizer was not there, becuase the strings are different (`foo` vs. `bucket`)

mfojtik: updated

mfojtik:
@iNecas @martinkunc all updated, are we fine here?

iNecas:
Letting @martinkunc to have the final word, but not other comments from me. Thanks

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -329,6 +344,22 @@ func GatherClusterAuthentication(i *Gatherer) func() ([]record.Record, []error)
 	}
 }
 
+// GatherClusterImageRegistry fetches the cluster Image Registry configuration
```
martinkunc, line 347: Could you please also run `make gen-doc` ? That should update the `doc/gathered-data.md` with this new method.

mfojtik: done

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -2,61 +2,75 @@ package clusterconfig
...
 	// maxEventTimeInterval represents the "only keep events that are maximum 1h old"
 	// TODO: make this dynamic like the reporting window based on configured interval
 	maxEventTimeInterval = 1 * time.Hour
+
+	registrySerializer serializer.CodecFactory
+	registryScheme     = runtime.NewScheme()
 )
 
+func init() {
+	utilruntime.Must(registryv1.AddToScheme(registryScheme))
```
martinkunc, line 50: Is is neccessary to register the registryv1 explicitly ? I would think it would be registered automatically by importing the package, like it is happening with openshift/kubernetes types. Or am I missing something ?

mfojtik: it is always better to make your own local scheme than poisoning or depending on some "automatic" behavior...

deads2k: > Is is neccessary to register the registryv1 explicitly ? I would think it would be registered automatically by importing the package, like it is happening with openshift/kubernetes types. Or am I missing something ?

Automatic registration is a bug, not a feature because it assumes a single unified scheme.  As we evolved our API, we developed disjoint schemes focused on their individual purposes.  This improves reliability because it avoids action at a distance.  It is more appropriate to install the types you're using in a local scheme like this.

martinkunc: Ok, nice, thanks for the explanation both !

PR #101, opened 2020-05-11
Bug 1820633: [release-4.4] Gather image config

mfojtik:
This picks https://github.com/openshift/insights-operator/pull/100 on top of the refactoring changes.

This is needed so the changes from master are still pickable to release-4.4.

/cc @iNecas @martinkunc


martinkunc:
Hi @mfojtik and sorry for the mess. Let me please understand. You basically merged all in 4.4 to allow future automated cherry-picks ? I think I would need to do the same for old releases as well.

martinkunc:
The CLs looks good as far as I can say. It includes some changes from #96, which is still pending because of freeze in 4.4., but I think it wont cause anything bad.
It would require adding BZ case, which would target 4.4.z release, so I guess it would still be hold to merge unless I am missing something.

martinkunc:
/retitle Bug 1820633: [release-4.4] Gather image config

martinkunc:
Oh wait, you are merging this to master ? Shouldn't it be to release-4.4 branch ?
/hold

PR #105, opened 2020-05-13
[release-4.3] Bug 1825758: Collecting config maps

martinkunc:
Picks changes from #86 and adds whole lot of missing historical commits to ease future merges into 4.3


ashcrow:
There is a significant difference in commits here compared to what went into master. Can you give me a tl;dr as to why that is before I review as patch manager?

martinkunc:
> There is a significant difference in commits here compared to what went into master and 4.4. Can you give me a tl;dr as to why that is before I review as patch manager?

That is right. I found that in 4.3 was missing significant amount of commits from history. I took this step and tried to reconcillieate it all in this commit. There is similar work for branch 4.4 in #101 (which should differ in adding collecting image registry).
For 4.2 the effort is in #110 

The whole effort it so allow automatic merges in the old versions. Changes in IO are often required to be merged and without this next developer would have to carefully hand pick his change. I hope after this, situation will be much more manageable.

PR #107, opened 2020-05-15
[release-4.2] Bug 1825770: Collecting config maps

martinkunc:
This PR collects all missing history into 4.2 and picks #86


tremes:
OCP 4.2 is EOL so we can close this PR (I don't have permissions to do so)

PR #103, opened 2020-05-12
[release-4.4] Bug 1834677: Gather image registry  config

mfojtik:
This picks #100 on top of the refactoring changes.

This is needed so the changes from master are still pickable to release-4.4.

/cc @iNecas @martinkunc

PR #113, opened 2020-05-19
Bug 1837967: Gathering Image Pruner configuration

ricardomaraschini:
Adding cluster's Image Registry pruner configuration to gather.

### DIFF
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -44,6 +44,13 @@ Location in archive: config/id/
 See: docs/insights-archive-sample/config/id
 
 
+## ClusterImagePruner
+
+fetches the image pruner configuration
+
+Location in archive: config/imagepruner/
+
+
 ## ClusterImageRegistry
 
 fetches the cluster Image Registry configuration
--- a/manifests/03-clusterrole.yaml
+++ b/manifests/03-clusterrole.yaml
@@ -99,6 +99,7 @@ rules:
     - imageregistry.operator.openshift.io
   resources:
     - configs
+    - imagepruners
   verbs:
     - get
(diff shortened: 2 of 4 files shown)


dmage:
/lgtm

/cc @iNecas @martinkunc
This is a new config for the automated image pruner that was added in 4.4.

martinkunc:
Hi Ricardo, should this be explicitly gathered only for 4.5, or does image-pruner exist in 4.4,4.3 or 4.2 ? Do you know ?
For better maintainability we should try to merge in older versions. I have other PRs which should be resolved before this will go smoothly. I could do it afterwards, if that would make sense to collect even in older versions.

ricardomaraschini:
> Hi Ricardo, can it be merged into 4.4 as well ?
> For better maintainability we should try to merge in older versions. There is #101 which should be merged before this will go smoothly. I could do it afterwards, if that would make sense to collect even there.

@martinkunc We need this backported to 4.4 as well. As you mentioned we need to wait for https://github.com/openshift/insights-operator/pull/113 to be merged first but that one seems to be closed.

Could you clarify?

martinkunc:
> > Hi Ricardo, can it be merged into 4.4 as well ?
> > For better maintainability we should try to merge in older versions. There is #101 which should be merged before this will go smoothly. I could do it afterwards, if that would make sense to collect even there.
> 
> @martinkunc We need this backported to 4.4 as well. As you mentioned we need to wait for #113 to be merged first but that one seems to be closed.
> 
> Could you clarify?

@ricardomaraschini Sorry, we need to wait for #103 to be merged to 4.4 before this one.

PR #114, opened 2020-05-21
pkg/controller/status/status: Set reasons for conditions

wking:
We've been light on reasons for expected conditions like `Progressing=False` since 80ab9237bc.  But if we feel like we have a message we want to set to help humans understand the condition, we should be setting a reason string for machines too.  `AsExpected` follows [the library-go precedent][1].

The `Degraded` reason I'm adding here isn't a great fit for a progressing reason, but does match the current logic used to set `Progressing=True`.  The insights operator doesn't actually supply any in-cluster APIs beyond ClusterOperator as far as I can tell, so it's not clear to me what sort of progressing it would do except its internal intialization (which I give a new `Initializing` reason).  But adding an odd reason seemed easy enough, and we can always circle back later and remove this `Progressing=True` case if we decide we don't need it.

[1]: https://github.com/openshift/library-go/blob/94c59dec54be25c8527e51e8c0a885712aeb01b5/pkg/operator/status/condition.go#L67

### DIFF
--- a/docs/insights-archive-sample/config/clusteroperator/insights
+++ b/docs/insights-archive-sample/config/clusteroperator/insights
@@ -15,22 +15,26 @@
       {
         "type": "Degraded",
         "status": "False",
+        "reason": "AsExpected",
         "lastTransitionTime": "2020-03-18T12:14:52Z"
       },
       {
         "type": "Available",
         "status": "True",
+        "reason": "AsExpected",
         "lastTransitionTime": "2020-03-18T12:14:52Z"
       },
       {
         "type": "Progressing",
         "status": "False",
         "lastTransitionTime": "2020-03-18T12:16:52Z",
+        "reason": "AsExpected",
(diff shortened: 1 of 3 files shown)


martinkunc:
@wking, could you please add BZ Bug prefix, which would allow us to merge it historically. Is it ok to merge all the way to 4.2 ?

wking:
> ...could you please add BZ Bug prefix...

Is this serious enough to backport?  4.2 at least is maintenance mode, and I'm pretty sure this does not meet that bar.  I was expecting it to land next week once 4.5 forks off master and never be backported at all.

martinkunc:
Ok, I agree this is a minor change. The reasoning behind this was to allow future collaborators effortless automatic merge as much as possible.

martinkunc:
@wking Actually, I could backport this for you if you dont mind. I am sure being proactive will save future mergers some, even minimal effort. There are many people committing to repo nowadays.

PR #115, opened 2020-05-21
Bug 1838973: GatherClusterOperators: store pod logs

vrutkovs:
If the pod is not healthy Insights operator should collect
previous/current logs for the pod containers. Only last 100 lines
is being kept to prevent the archive from growing large.

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -1,10 +1,12 @@
 package clusterconfig
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/pem"
 	"fmt"
+	"io"
 	"regexp"
 	"sort"
 	"strings"
@@ -44,6 +46,9 @@ var (
 
 	registrySerializer serializer.CodecFactory
 	registryScheme     = runtime.NewScheme()
+
+	// logTailLines defines a number lines to keep when fetching pod logs
+	logTailLines = int64(100)
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


martinkunc:
@vrutkovs Thank you.
Do you have any guess how much data it collects for one pod and how many pods we typically have in a cluster ?

vrutkovs:
>Do you have any guess how much data it collects for one pod and how many pods we typically have in a cluster ?

This would collect logs for pods, related to a failing operator. Usually its 1-5 pods for failing operator. In my tests this increased archive size from 12kB to 14kB in some scenarios, so it doesn't affect the size too much

martinkunc:
/lgtm
Thank you @vrutkovs !

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -164,11 +169,40 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
...
+							}
+
+							req := i.coreClient.Pods(namespace).GetLogs(pod.Name, &corev1.PodLogOptions{Previous: previous, TailLines: &logTailLines, Container: container.Name})
+							readCloser, err := req.Stream()
+							if err != nil {
+								klog.V(2).Infof("Failed to fetch %s for %s pod in namespace %s for failing operator %s: %q", currentOrPrevious, pod.Name, namespace, item.Name, err)
+								continue
+							}
+
+							defer readCloser.Close()
```
smarterclayton, line 194: this leaks connection streams, you can't do this :)

You need to warp the inner loop in a 

```
func() {
  ...
  defer readCloser.Close()
}()
```

construct if you want to do this.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -164,11 +169,40 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
...
+
+							buf := bytes.NewBufferString("")
```
smarterclayton, line 196: Log lines can be multimegabytes.  We had a bug recently where a single line was 15MB.  This means that you could hold potentially unlimited memory in the operator, causing a crash that would be unrecoverable.

You need to limit the maximum bytes the log can gather (MaxBytes field) and reuse the buffer instead of initializing it for every pod.

You also need to consider what the maximum number of pods could be read this way and limit the size based on number of total containers you are gathering logs for.  If you are gathering logs for 100 failing containers, 500k might be acceptable.  If this is a huge cluster (5k nodes) and you have a crashlooping daemonset, it might be 20k containers, in which case 500k is way too big.  You MUST bound total memory consumption in this operator to some reasonable limit (50mb is about what i might allow).

The real solution here is that you are misuing Record - you want a record type that is backed by a file (and you should be streaming these logs to a temporary file location and then cleaning them up after they are copied to the tar, or moving them from temp dir to final location at the time record is committed).  Record is not supposed to be used for unbounded large objects.

vrutkovs: >You need to limit the maximum bytes the log can gather (MaxBytes field) and reuse the buffer instead of initializing it for every pod.

Implemented in https://github.com/openshift/insights-operator/pull/126

>The real solution here is that you are misuing Record - you want a record type that is backed by a file

Agree, we're storing arbitrary user certificates from config maps too, so a file-backed storage might be useful in other places. @martinkunc do you think this is worth a jira story?

smarterclayton: I would probably say any one gatherer should never exceed 1-5mb in memory at any one time, and so for instance if you read multiple things we need records that can be lazily put on temp disk and then moved to the upload / record location on demand

PR #117, opened 2020-05-28
Bug 1841057: Skip the initial upload delay

martinkunc:
If the operator is started for the first time, or if it is not degraded, it will upload the collected data immediately after they are collected.

This is removing the original initial jitter delay between periods. Consequent runs will still be jittered. The 15 seconds detection of data to send is still there even for initial run, so the data should arrive in 15 seconds after they are collected (few seconds).

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -43,10 +43,11 @@ type Controller struct {
 	statusCh     chan struct{}
 	configurator Configurator
 
-	lock     sync.Mutex
-	sources  []controllerstatus.Interface
-	reported Reported
-	start    time.Time
+	lock             sync.Mutex
+	sources          []controllerstatus.Interface
+	reported         Reported
+	start            time.Time
+	safeInitialStart bool
 }
 
 func NewController(client configv1client.ConfigV1Interface, configurator Configurator, namespace string) *Controller {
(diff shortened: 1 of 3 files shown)


martinkunc:
@iNecas Would you please have some time to look at this PR ? Thank you..

smarterclayton:
The reason we did this was so that if the operator went into a crashloop it would not overwhelm the upstream server.  How will you mitigate that now after this change went in?

martinkunc:
> The reason we did this was so that if the operator went into a crashloop it would not overwhelm the upstream server. How will you mitigate that now after this change went in?

@smarterclayton I was hoping that skipping fast upload on Operator's degrated status would be enough. Are you suggesting that I should add detection for POD status as well ? I was thinking about something like this:
```
pod, _  := gatherKubeClient.CoreV1().Pods(os.Getenv("POD_NAMESPACE")).Get(os.Getenv("POD_NAME"))
	allHealthy := true
	for _, c := range pod.Status.ContainerStatuses {
		allHealthy = allHealthy && c.RestartCount == 0
	}
	if allHealthy // fast upload
```
Do you think it would be better ?

Also I found that I should turn off the consequent fast upload after first successfull upload, rather that based only on non initial call of updateStatus, because Gathering might not be finished for larger clusters at time when updateStatus is called next time.

martinkunc:
@smarterclayton I realized that Restarts can be something good as well. Would there be a better way, maybe checking LastTerminationState?=State.ExitCode != 0 
Would it be good enough ?

martinkunc:
Trying to address the better status check in https://github.com/openshift/insights-operator/pull/132

PR #119, opened 2020-06-04
pkg/gather/clusterconfig/clusterconfig: include node information in every archive

vrutkovs:
A lot of scenarios could use node information even if these are reporting
as healthy - e.g. custom taints may prevent pods from scheduling or
single master installs. This commit ensures anonymized node information
is included in every archive.


martinkunc:
/lgtm
/approve
Thank you @vrutkovs ! Very nice,

PR #123, opened 2020-06-05
Bug 1844419: [release-4.4] GatherClusterOperators: store pod logs

vrutkovs:
If the pod is not healthy Insights operator should collect
previous/current logs for the pod containers. Only last 100 lines
is being kept to prevent the archive from growing large.

This also backports #128 so that resulting tarball doesn't exceed 8Mb

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -42,6 +42,7 @@ import (
 	networkv1client "github.com/openshift/client-go/network/clientset/versioned/typed/network/v1"
 
 	"github.com/openshift/insights-operator/pkg/record"
+	"github.com/openshift/insights-operator/pkg/record/diskrecorder"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


tremes:
@vrutkovs Can you please rebase?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -202,19 +205,48 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
...
 					}
-					records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pods.Items[i].Namespace, pods.Items[i].Name), Item: PodAnonymizer{&pods.Items[i]}})
-				}
-				if namespaceEventsCollected.Has(namespace) {
-					continue
+					records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pod.Namespace, pod.Name), Item: PodAnonymizer{pod}})
+
+					// Fetch previous and current pod logs
+					for _, previous := range []bool{true, false} {
+						for _, container := range pod.Spec.Containers {
```
tremes, line 225: I was looking at this in 4.5 and master and there we use `pod.Spec.InitContainers`

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -202,19 +205,48 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
...
+							}
+
+							defer readCloser.Close()
+
+							buf := bytes.NewBufferString("")
+							_, err = io.Copy(buf, readCloser)
+							if err != nil {
+								klog.V(2).Infof("Failed to write %s for %s pod in namespace %s for failing operator %s: %q", currentOrPrevious, pod.Name, namespace, item.Name, err)
+								continue
+							}
```
tremes, line 245: Would it be possible to move this to a new function pls? Same `collectContainerLogs` as we have in 4.5 and master. There's also some extra logic calculating the buffer size in 4.5 and master.

Just for the sake of smoother backports.

pkg/gather/clusterconfig/clusterconfig.go
vrutkovs: Good idea, backported release-4.5 parts here as well

pkg/gather/clusterconfig/clusterconfig.go
vrutkovs: Backported `collectContainerLogs` w/ size limitation

pkg/gather/clusterconfig/clusterconfig.go
tremes: Thanks for quick update. I'll check ASAP.

tremes:
Ohh this needs a rebase again pls.

tremes:
Verified against cluster.
/lgtm

PR #124, opened 2020-06-10
Make e2e tests more stable

martinkunc:
The e2e tests were occasionally failing, I think there were few problems:
- Tests in go run in indefined order. Normal unit tests should not change environment for consequent tests. For tests which were modifying the cluster I added cleanup methods, which should revert caused changes.
- Added extra test to make sure pull-secret exists
- Shortened upload interval for OptIn test, which makes sure we upload as fast as possible and not wait for random Jitter period. (Not waiting on startup skips the initial delay only when upload is executed in first 30 seconds (Because it is based on initial call to status.update) - maybe this could be improved).
- Fixed IsDisabled condition
- Added check for disabled to OptIn test beginning

### DIFF
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -2,90 +2,157 @@ package integration
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/insights-operator/pkg/controller/status"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// Check if opt-in/opt-out works
-func TestOptOutOptIn(t *testing.T) {
-	// Backup pull secret from openshift-config namespace.
-	// oc extract secret/pull-secret -n openshift-config --to=.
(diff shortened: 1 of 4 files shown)


test/integration/basic_test.go
```diff
@@ -2,90 +2,157 @@ package integration
...
+	if secretConfig, ok = pullSecret.Data[".dockerconfigjson"]; !ok {
+		t.Fatalf("The pull-secret didn't contain .dockerconfigjson key: %s", err)
+	}
+	obj := map[string]interface{}{}
+	errUnmarshal := json.Unmarshal(secretConfig, &obj)
+	if errUnmarshal != nil {
+		t.Fatal(errUnmarshal.Error())
+	}
+	creds := obj["auths"].(map[string]interface{})
+	if _, ok := creds["cloud.openshift.com"]; !ok {
```
tisnik, line 38: prob. worth to move this string into constantant?

test/integration/basic_test.go
```diff
@@ -2,90 +2,157 @@ package integration
...
+		t.Fatalf("The pull-secret read failed: %s", err)
+	}
+	// Backup support secret from openshift-config namespace.
+	// oc extract secret/support -n openshift-config --to=.
+	supportSecret, err := clientset.CoreV1().Secrets("openshift-config").Get("support", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("The pull-secret read failed: %s", err)
+	}
+	resetSecrets := func() {
+		err := forceUpdateSecret("openshift-config", "pull-secret", pullSecret)
```
tisnik, line 70: dtto?

PR #125, opened 2020-06-17
gather: Collect namespace level cpu and memory metrics

smarterclayton:
These metrics help identify distribution of CPU and memory use
within a cluster at infrequent intervals. All labels on the new
metrics are present in ALERTS and are also returned in telemetry,
so do not represent a larger source of gathered data.

The data will be used to identify regressions in core and extension
operator usage across the fleet, as well as to ensure that over time
OpenShift operators continue to reduce resource consumption.


martinkunc:
/test images
/lgtm

Thank you Clayton :)

PR #126, opened 2020-06-18
clusterconfig.go: limit the size of collected logs

vrutkovs:
This commit ensures clusterconfig won't collect large logs.
GatherClusterOperators would use maximum available compressed tarball
size (set in `diskrecorder`), multiply it by `logCompressionRatio`
(set to 2 initially) and split this budget between unhealthy pods.
`collectContainerLogs` would limit the amount of fetched bytes so
that log lines would not exceed container log budget.

Addresses https://github.com/openshift/insights-operator/pull/115#discussion_r441858786

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -34,6 +34,14 @@ import (
 	imageregistryv1 "github.com/openshift/client-go/imageregistry/clientset/versioned/typed/imageregistry/v1"
 
 	"github.com/openshift/insights-operator/pkg/record"
+	"github.com/openshift/insights-operator/pkg/record/diskrecorder"
+)
+
+const (
+	// Log compression ratio is defining a multiplier for uncompressed logs
+	// diskrecorder would refuse to write files larger than MaxLogSize, so GatherClusterOperators
+	// has to limit the expected size of the buffer for logs
+	logCompressionRatio = 2
 )
 
 var (
@@ -47,7 +55,7 @@ var (
 	registrySerializer serializer.CodecFactory
 	registryScheme     = runtime.NewScheme()
 
-	// logTailLines defines a number lines to keep when fetching pod logs
+	// logTailLines sets maximum number of lines to fetch from pod logs
 	logTailLines = int64(100)
 )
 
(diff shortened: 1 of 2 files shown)


vrutkovs:
From https://github.com/openshift/insights-operator/pull/115#discussion_r442657086

>What if we would add these additional limits to the logic in #126.

>    Collecting logs from maximally 200 failing pods
>    Collect at max 5000 bytes of log
>    200 * 5000 = at max 1Mb

I don't think we should be tweaking log collection by hardcoding parameters. Currently we're limited by two factors:

* collection time (gathering must succeed in interval/2). 
* final tarball size should not exceed 8 MBs.

Not much we can do about the first one (this param is bound to API performance), so this PR focuses on the second one. When pods are failing it is critical to collect as many logs as possible - 5000 bytes might or might not be enough.

Limit on a number of processed pods makes sense as it ensures collection time is sane. The code is sequentially requesting each container's logs, so I guess it would make sense to limit `totalUnhealthyContainers` to 100 (so that we'd have no more than 200 logs for current and previous containers)

martinkunc:
I was thinking to use the 5K as the maximal hard limit, to limit the case one of archives would be still in allowed range, but because of excessive logging It might be something we are not interested. So the final limit would be min(by_archive, 5000).
Also, I think the calculation should know that there are other data in archive, which we want. maybe logs could be 90% of max archive size ?.

martinkunc:
Should we worry about logs anonymization somehow in here ?

vrutkovs:
> Also, I think the calculation should know that there are other data in archive, which we want. maybe logs could be 90% of max archive size ?.

Low `logCompressionRatio` setting leaves quite a lot of space for other data. In my tests (log all pods for all operators) logs used ~95% of available log bundle quota and reached 500Kb compressed (up to 125Kb of uncompressed container logs was collected), so there's plenty of space for other Insights data there.

> Should we worry about logs anonymization somehow in here ?

Lets file a Jira for that and track it separately. Most of the container logs use internal endpoints / certificates etc., so anonymization may render logs useless.

martinkunc:
@vrutkovs Ok, I forgot this is just for Pods from clusteroperators, in that case I guess anonymization is probably not that important.

vrutkovs:
@martinkunc needs lgtm again due to merge conflict (in a comment for GatherNodes/GatherUnhealthyNodes)

PR #128, opened 2020-06-26
Bug 1851492: [release-4.5] clusterconfig.go: limit the size of collected logs

vrutkovs:
This commit ensures clusterconfig won't collect large logs.
GatherClusterOperators would use maximum available compressed tarball
size (set in `diskrecorder`), multiply it by `logCompressionRatio`
(set to 2 initially) and split this budget between unhealthy pods.
`collectContainerLogs` would limit the amount of fetched bytes so
that log lines would not exceed container log budget.

Cherry-pick of https://github.com/openshift/insights-operator/pull/126

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -34,6 +34,14 @@ import (
 	imageregistryv1 "github.com/openshift/client-go/imageregistry/clientset/versioned/typed/imageregistry/v1"
 
 	"github.com/openshift/insights-operator/pkg/record"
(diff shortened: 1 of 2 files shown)


martinkunc:
Thank you @vrutkovs 
/lgtm

sdodson:
@vrutkovs why the force push, is this now out of sync with master?

vrutkovs:
@sdodson comment update from https://github.com/openshift/insights-operator/pull/127/files#diff-397c7d21d587358413d334b87528a7cbL224-R224 caused merge conflict

PR #131, opened 2020-06-29
[release-4.4] Bug 1834677: Gather image registry config (refresh)

martinkunc:
This picks #100 on top of the refactoring changes and superseeds #103

This is needed so the changes from master are still pickable to release-4.4.

/cc @mfojtik


martinkunc:
/retitle [release-4.4] Bug 1834677: Gather image registry config (refresh)

sdodson:
Is this another of the big cleanup efforts? I'm surprised by how many commits there are, many of them look like they're cleaning up interim state. If this is a clean pick from already merged commits upstream that's fine but in the future lets make sure that we don't have interim commits that produce broken state.

See https://github.com/openshift/enhancements/blob/master/guidelines/commit_and_pr_text.md#number-of-commits

martinkunc:
> Is this another of the big cleanup efforts? I'm surprised by how many commits there are, many of them look like they're cleaning up interim state. If this is a clean pick from already merged commits upstream that's fine but in the future lets make sure that we don't have interim commits that produce broken state.
> 
> See https://github.com/openshift/enhancements/blob/master/guidelines/commit_and_pr_text.md#number-of-commits

@sdodson This is indeed one. We did some refactoring, plus had some missing commits in past on which this PR depends. Resovling it now would help us automate the merges much more often. I will try to limit the number of commits in the future.

PR #134, opened 2020-07-07
Test pods logs collected - Automate BZ1838973

psimovec:
This is my first contribution in golang, any suggestions are welcome!

automates https://bugzilla.redhat.com/show_bug.cgi?id=1838973

Changes to previous code:
upgraded checkPodsLogs to accept regex instead of a string / added possibility to check only new logs... I don't know if the solution I have done is alright, I just did not want to change old function's parameters..

Added:
ExecCmd - execute bash command on a pod
TestPodLogsCollected test case

(In recent PR https://github.com/openshift/insights-operator/pull/124 , waiting time for logs was reduced from 30 mins to 5 mins... But I wait for a fresh archive, and it takes 6 minutes. Is it okay to set it to 10 minutes?) solved

(I will squash commits after review if everything is ok)


test/integration/main_test.go
```diff
@@ -79,6 +84,16 @@ func clusterOperatorInsights() *configv1.ClusterOperator {
 	return operator
 }
 
+
+func clusterOperatorMonitoring() *configv1.ClusterOperator {
```
lina-nikiforova, line 88: I'd suggest to merge it with `clusterOperatorInsights` and create one function that can check any operator

test/integration/main_test.go
```diff
@@ -169,19 +184,24 @@ func restartInsightsOperator(t *testing.T) {
 	t.Log(errPod)
 }
 
-func checkPodsLogs(t *testing.T, kubeClient *kubernetes.Clientset, message string) {
+func _checkPodsLogs(t *testing.T, kubeClient *kubernetes.Clientset, message string, newLogs bool) {
```
lina-nikiforova, line 187: I'm not sure how underscore function name is used in go, @martinkunc ?

test/integration/main_test.go
```diff
@@ -169,19 +184,24 @@ func restartInsightsOperator(t *testing.T) {
...
+	r, _ := regexp.Compile(message)
 	newPods, err := kubeClient.CoreV1().Pods("openshift-insights").List(metav1.ListOptions{})
 	if err != nil {
 		t.Fatal(err.Error())
 	}
-
+	tajm := metav1.NewTime(time.Now())
```
lina-nikiforova, line 193: tajm - is it supposed to be `time`? :)

test/integration/main_test.go
martinkunc: You are right, go devs are a bit stricter about naming. Please use camelCase for private and PascalCase for public methods. Underscores should be only in file names, not in variables. See here: https://golang.org/doc/effective_go.html#mixed-caps

test/integration/main_test.go
```diff
@@ -79,6 +84,16 @@ func clusterOperatorInsights() *configv1.ClusterOperator {
 	return operator
 }
 
+
+func clusterOperatorMonitoring() *configv1.ClusterOperator {
+	// get info about insights cluster operator
+	operator, err := configClient.ClusterOperators().Get("monitoring", metav1.GetOptions{})
+	if err != nil {
+		panic(err.Error())
```
martinkunc, line 92: I think generally its better not to use panic in tests. would it be possible to add t *testing.T and use t.Fatal(err) instead ? The reason is to distinguish valid test "failure" from panic when something really unexepected happens.

test/integration/main_test.go
```diff
@@ -236,6 +256,100 @@ func forceUpdateSecret(ns string, secretName string, secret *v1.Secret) error {
...
+	}
+	for _, newPod := range newPods.Items {
+		if strings.HasPrefix(newPod.Name, prefix) {
+			return &newPod
+		}
+	}
+	return nil
+}
+
+func PodsLogsExist(t *testing.T, kubeClient *kubernetes.Clientset) bool {
```
martinkunc, line 326: At first I was not exacly sure what this method does just by its name. Pod logs is quite specific, but its not related actually. What about calling it something like ArchiveContainsPodLogs, or anything else :)
Coudln't you also please take out code to degrate in separate function, maybe degradeMonitoring, also this would be both neede to be reverted in the end. For that I would call it alrady from TestPodLogsCollected.
You could use return defer function, something like:
```
func action() func() {
	fmt.Println("doing action")
	return func() {
		fmt.Println("reverting action")
	}
}

func main() {
	defer action()()

	fmt.Println("main")
}
```

martinkunc:
Ahh, also it looks e2e tests failed on the build machine. I am guessing that you have used a package, which is not yet in vendor folder in github. When you bring a new package, you need to commit the content of vendor, go.mod and go.sum as well. Please be Please run 
```
make vendor
```
to keep it clean.

test/integration/main_test.go
psimovec: I have extracted function that can check any operator, it looks cleaner, but kept the functions, so the strings of operator names don't have to be used in code, but keeping the functions might be useless (for now I at least don't touch other tests)

test/integration/main_test.go
psimovec: By this naming, it looks like most of the main_test.go functions should also be renamed - most functions are in camelCase but are used as public ones

test/integration/main_test.go
psimovec: oh :D the last variable that made my life happier, forgot to rename it. It was time, but it's already taken, unfortunately :D

test/integration/main_test.go
psimovec: This will be also done in follow-up cleanup PR, so I don't touch other test functions in this PR

test/integration/main_test.go
```diff
@@ -236,6 +256,100 @@ func forceUpdateSecret(ns string, secretName string, secret *v1.Secret) error {
...
+	err = kubeClient.CoreV1().Pods(pod.Namespace).Delete(pod.Name, &metav1.DeleteOptions{})
+	e(t, err, "Failed to delete Pod")
+	operator := clusterOperatorMonitoring()
+	if !isOperatorDegraded(t, operator) {
+		t.Log("Operator is not degraded!!!")
+	}
+	// wait for fresh report
+	checkNewPodsLogs(t, kubeClient, "Writing \\d+ records to")
+	insightsPod := findPod(t, kubeClient, "openshift-insights", "insights-operator")
+	hasLatestArchiveLogs := "tar tf $(ls -dtr /var/lib/insights-operator/* | tail -1)|grep -c \"\\.log$\""
```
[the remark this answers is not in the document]
psimovec: Great to know about multi-line string, thanks!
And the specific matching makes more sense.

test/integration/main_test.go
```diff
@@ -236,6 +256,100 @@ func forceUpdateSecret(ns string, secretName string, secret *v1.Secret) error {
 	return nil
 }
 
+func checkPodsLogs(t *testing.T, kubeClient *kubernetes.Clientset, message string) {
```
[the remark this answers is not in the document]
psimovec: nice pattern! I cheated it a little bit with ...bool for one optional parameter only, I think it is way simpler than using options struct with one bool value, which gets set up by function. If you don't like this, I can change it to Functional Options pattern

test/integration/main_test.go
martinkunc: Ok. And regarding the public functions. Public in Go means that they are accessible from other packages, which is also making a contract with other packages that they will keep working, which is why we tend not to use them if we don't have to.
In case of tests, I think we are in same packge, so we don't need to expose functions for someone else, except for Test... functions, executed by test runner.

PR #116, opened 2020-05-26
Bug 1840012: [release-4.4] pkg: write files with extension

vrutkovs:
`Marshalable` interface is extended to provide a file extension. It
is being appended to the files when these are written to disk.
If marshaller returns empty string no extension is appended.

Cherry-pick of https://github.com/openshift/insights-operator/pull/99 on 4.4

PR #140, opened 2020-07-09
Bug 1855313: [release-4.3] pkg: write files with extension

vrutkovs:
`Marshalable` interface is extended to provide a file extension. It
is being appended to the files when these are written to disk.
If marshaller returns empty string no extension is appended.

Cherry-pick of https://github.com/openshift/insights-operator/pull/116 in release-4.3


tremes:
@vrutkovs Can we close this one?

PR #141, opened 2020-07-10
Info how to retrieve key and certificate and simple script to do so

tisnik:
Info how to retrieve key and certificate and simple script to do so


martinkunc:
@tisnik Thank you, very interestring

PR #142, opened 2020-07-13
Report the returned response body to log the error detail from cloud.redhat.com

martinkunc:
This should return (for log) the http body, for example if ingress returns http errors.

### DIFF
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -30,6 +30,10 @@ import (
 	"github.com/openshift/insights-operator/pkg/authorizer"
 )
 
+const (
+	responseBodyLogLen = 1024
+)
+
 type Client struct {
 	client      *http.Client
 	maxBytes    int64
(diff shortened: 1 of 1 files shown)


pkg/insights/insightsclient/insightsclient.go
```diff
@@ -229,6 +225,17 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
 	return nil
 }
 
+func responseBody(r *http.Response) string {
+	if r == nil {
+		return ""
+	}
+	body, _ := ioutil.ReadAll(r.Body)
```
tremes, line 232: Is it OK to ignore the error here?

martinkunc: I think so, see the original version. We are trying to log what we received from response. If we cannot, we will log empty string.

tremes: Ahh yes...sorry I was looking rather at the green fields than the red ones.

PR #145, opened 2020-07-16
[WIP] E2e tests small cleanup/refactor

psimovec:
1. removing clusterOperatorInsights() as this function was generalised in previous PR, but did not want to touch other tests in the PR
2. simplifying error check pattern.
```
	if err != nil {
		t.Fatal(err.Error(), "error message")
	}
```
this pattern was appearing very often, so I have simplified it to
```
	e(t, err, "error message")
```
Now there is possible to add any number of error messages to e function, so
```
	e(t, err, "error", "message")
```
or
```
	e(t, err)
```
works too

The e function also prints stack trace inside integration tests
example:
```=== RUN   TestPodLogsCollected
    main_test.go:292: [error message] mesage_inside_err
        github.com/openshift/insights-operator/test/integration.checkPodsLogs(0xc000582f00, 0xc00029b760, 0x13f0de6, 0x20, 0xc000615f6f, 0x1, 0x1)
        	/home/psimovec/w/insights-operator/test/integration/main_test.go:175 +0x214
        github.com/openshift/insights-operator/test/integration.TestPodLogsCollected(0xc000582f00)
        	/home/psimovec/w/insights-operator/test/integration/bugs_test.go:139 +0x72
--- FAIL: TestPodLogsCollected (0.16s)
```

We discussed with Martin, that in future we could use some assert library instead

### DIFF
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -44,7 +44,7 @@ func TestPullSecretExists(t *testing.T) {
 func TestOptOutOptIn(t *testing.T) {
 	// initially IO should be running
 	errDisabled := wait.PollImmediate(1*time.Second, 30*time.Second, func() (bool, error) {
-		insightsNotDisabled := !operatorStatus(t, clusterOperatorInsights(), status.OperatorDisabled, configv1.ConditionTrue)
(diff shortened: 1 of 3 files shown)


psimovec:
This is not relevant anymore, as the tests are being ported to different repo, in python

PR #146, opened 2020-07-21
PoC: Insights Operator retrieves latest report for the cluster from Smart Proxy

joselsegura:
This commit is a prove of concept.

It enables the Insights Operator to retrieve the latest report for the cluster from the Insights Results Smart Proxy service


joselsegura:
> It looks fine, just for the final version I thing that devels won't be happy with dependency with Aggregator :) But perfectly ok for PoC

Probably I should use `insights-operator-utils` instead of aggregator...

joselsegura:
@tisnik at the end I needed to use `insights-results-smart-proxy` because we have some useful structs defined there. It is just a PoC, but maybe we will need to move to a different place for a final version of this PR

PR #147, opened 2020-07-31
Limit the metric alerts to at max 1137 lines

martinkunc:
This downloads original metrics, but limits potentially huge alerts to 1137 lines. This is to handle situation when crashlooping service was generating thousands of new alert instances and the resulting metrics were huge.

PR #148, opened 2020-08-03
Bug 1862444: Limit collection of ALERTS metric to 1000 lines (~500KiB) to avoid unbearably large archives

natiiix:
Copy of #147 with a couple of additional fixes and tweaks.

The docs update also includes a forgotten update from #133.

## Description from BugZilla

> When an operator starts generating big amount of Alerts into the prometheus, Insights operator is downloading all the alerts and the resulting file might be big.
>
> The consequence is that internal data pipeline in CCX is failing, because we cannot store archive to Kafka, where maximal limit is 1MB a if the file is larger then 300 KB, services were failing during the Insights rule processing when it was extracting it all to memory.

### DIFF
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -159,36 +159,39 @@ Location in archive: config/configmaps/
 See: docs/insights-archive-sample/config/configmaps
 
 
+## ContainerImages
+
+collects essential information about running containers.
+Specifically, the age of pods, the set of running images and the container names are collected.
+
+Location in archive: config/running_containers.json
+
+
 ## MostRecentMetrics
 
 gathers cluster Federated Monitoring metrics.
 
 The GET REST query to URL /federate
 Gathered metrics:
-  ALERTS
   etcd_object_counts
   cluster_installer
   namespace CPU and memory usage
+  followed by at most 1000 lines of ALERTS metric
 
 Location in archive: config/metrics/
 See: docs/insights-archive-sample/config/metrics
 
 
-Output raw size: 108
+Output raw size: 120
 
 ### Examples
 
 #### MostRecentMetrics
-[{"Name":"config/metrics","Captured":"0001-01-01T00:00:00Z","Fingerprint":"","Item":"SGVsbG8sIGNsaWVudAo="}]
+[{"Name":"config/metrics","Captured":"0001-01-01T00:00:00Z","Fingerprint":"","Item":"SGVsbG8sIGNsaWVudAojIEFsZXJ0cyAK"}]
 
 ## Nodes
 
 collects all Nodes.
-
-The node is unhealthy when:
-the operator.Status.Conditions.Condition Type
-  is OperatorDegrated and Status is True or
-     OperatorAvailable and Status is False
 
 The Kubernetes api https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/node.go#L78
 Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#nodelist-v1core
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -7,6 +7,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/url"
 	"regexp"
 	"sort"
@@ -47,6 +48,12 @@ const (
 	// imageGatherPodLimit is the maximum number of pods that
 	// will be listed in a single request to reduce memory usage.
 	imageGatherPodLimit = 200
+
+	// metricsAlertsLinesLimit is the maximal number of lines read from monitoring Prometheus
+	// 500 KiB of alerts is limit, one alert line has typically 450 bytes => 1137 lines.
+	// This number has been rounded to 1000 for simplicity.
+	// Formerly, the `500 * 1024 / 450` expression was used instead.
+	metricsAlertsLinesLimit = 1000
 )
 
 var (
@@ -64,6 +71,9 @@ var (
 	logTailLines = int64(100)
 
 	imageHostRegex = regexp.MustCompile(`(^|\.)(openshift\.org|registry\.redhat\.io|registry\.access\.redhat\.com)$`)
+
+	// lineSep is the line separator used by the alerts metric
+	lineSep = []byte{'\n'}
 )
 
 func init() {
@@ -123,10 +133,10 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error 
 //
 // The GET REST query to URL /federate
 // Gathered metrics:
-//   ALERTS
 //   etcd_object_counts
 //   cluster_installer
 //   namespace CPU and memory usage
+//   followed by at most 1000 lines of ALERTS metric
 //
 // Location in archive: config/metrics/
 // See: docs/insights-archive-sample/config/metrics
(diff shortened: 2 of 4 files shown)


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -147,6 +151,24 @@ func GatherMostRecentMetrics(i *Gatherer) func() ([]record.Record, []error) {
...
+			return []record.Record{{Name: "config/metrics", Item: RawByte(fmt.Sprintf("# error: %v\n", err))}}, nil
+		}
+		r := NewLineLimitReader(rsp, metricsAlertsLinesLimit)
+		alerts, err := ioutil.ReadAll(r)
+		if err != nil {
+			klog.Errorf("Unable to read most recent alerts from metrics: %v", err)
+			return nil, []error{err}
+		}
+
+		data = append(data, []byte("# Alerts \n")...)
```
alexandrevicenzi, line 170: shouldn't this be uppercase?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -123,10 +128,10 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
 //
 // The GET REST query to URL /federate
 // Gathered metrics:
-//   ALERTS
 //   etcd_object_counts
 //   cluster_installer
 //   namespace CPU and memory usage
+//   followed by at most 1137 lines of ALERTS metric
```
alexandrevicenzi, line 134: is there any behavior for these 1137 lines? I would like to see the most recent alerts, and why not use a number like 1000?

alexandrevicenzi: another question, maybe for @iNecas, why not stop collecting alerts since we already have it on Thanos.

We will need to query Thanos anyway to get all the alerts because the archive does not ship all alerts, why bother sending it anyway if sending just in half, thinking of pipeline perspective, maybe for others this makes sense but for us, it does not, I would never rely on IO data if not complete.

natiiix: @alexandrevicenzi Please see the constant above. It's a bit of an odd number, but on average, 1137 alerts would add up to 500 KiB of data.

pkg/gather/clusterconfig/clusterconfig.go
natiiix: I'm not really sure because I haven't worked with this part of the collected data before. Should it?

pkg/gather/clusterconfig/clusterconfig.go
iNecas: I think it still makes sense to have them but one good point: can we indicate that the data are incomplete? We can then have a look on how % of clusters are hitting this situation and make a decision whether this is the way or not.

Especially for rules writing, depending on external service is always a bit problematic, as you know @alexandrevicenzi. Therefore it's good to understand what real impact we're talking about here. On side node, if a cluster is hitting 1000+ alerts, that's probably the first thing to fix before anything else, and out of those alerts, we can still ask ourselves, if we can help the customer fixing that.

pkg/gather/clusterconfig/clusterconfig.go
iNecas: What was there before this change (coming from prometheus directly):

```
# TYPE ALERTS untyped
```

pkg/gather/clusterconfig/clusterconfig.go
natiiix: @iNecas One thing I have discussed with Martin was that we could add the total number of alerts somewhere. WDYT? I've noticed there's a `# Alerts\n` label/comment above the alerts. Would it be possible to add a number after that or would it break some king of a pre-defined format?

pkg/gather/clusterconfig/clusterconfig.go
alexandrevicenzi: the Prometheus parser in Python reads that line, if something does not quite match it may not work, so keep what Ivan mentioned

pkg/gather/clusterconfig/clusterconfig.go
alexandrevicenzi: @natiiix I just find it an odd number, something even would be more interesting, if I get 1000 or 1137 is does not make much of a difference.

@iNecas, what if the first 1k alerts are from Watchdog and later on there's an alert from a real problem, this is the problem that I see, truncating it is as bad as not having at all, at least in my perspective.

I won't block anything, it was just a question, but as I mentioned, in the pipeline perspective we won't use or trust this data much, we will still query Thanos for the complete list of alerts.

I would be more interesting to parse the alerts or to get a list of unique alerts, I don't think we really need to know all the times that one issue happened, the last one seen is probably good enough, as well skipping Watchdog related alerts would be something to save some space.

natiiix: @alexandrevicenzi I don't know what this data is used for, but a large quantity of them clearly indicates an issue as far as I see. Maybe the metrics are used by external rules? I don't know. Someone wanted it there and now it's there, but it's not acceptable to have 70 MB of metrics, no matter who wants them.

iNecas: > what if the first 1k alerts are from Watchdog

Do you have real examples for this? I've not seen more than one for a cluster, if there were 100s of that, I think that would be also thing to fix.

Different approach could be  `sum without` promql, to eliminate labels we don't care and have more concise view, but that would mean we would be putting additional load on in-custer prometheus (might not be an issue, but raising it here). Anyway, I don't want to over-complicate this particular PR, limiting the lines seems like good enough to do the initial mitigation (bonus points for the number of all alerts somewhere)

natiiix: @iNecas When Martin tried to filter them out using the query, he wasn't very successful apparently. I don't know the specifics, but he said it probably wouldn't be a good way of doing it.

alexandrevicenzi: To put the fire off this is ok, but a better solution would be probably more interesting in the long term, as I said, it was questions regarding the usefulness of the data in the pipeline, this is my concern.

iNecas: Adding here just to complete the context: https://github.com/prometheus/client_python/blob/3ab0374d1548769748932b9316d86b088dc7712e/prometheus_client/parser.py#L177 <- this is the parser @alexandrevicenzi meant that relies on specific form of the comment, so we should keep that untouched

iNecas: As for detecting the limit hit, for now, we could probably just guess from the number of lines: not great not terrible.

@natiiix @alexandrevicenzi one thing orthogonal to this, but that would help going forward would be putting more metadata about different collectors execution: I can think of execution time, raw files size, any errors etc., that we could use to store metadata about the collected data, and in this case could capture the fact that the data are incomplete + how many more alerts are there. Would you be willing to put some task in jira about this and see if that could be something to prioritize among other things on the teams plate?

pkg/gather/clusterconfig/clusterconfig.go
natiiix: @iNecas @alexandrevicenzi I believe this change doesn't actually touch that at all. That line will still be there. It will just be prefixed by this line that's been added by Martin.
If that's invalid and if that would cause trouble with the parser, then please let me know where should I put the total number of alerts because there doesn't seem to be any other suitable place for it.

pkg/gather/clusterconfig/clusterconfig.go
natiiix: @iNecas Okay, that makes sense. I will create a task for it.

docs/gathered-data.md
```diff
@@ -159,37 +159,40 @@ Location in archive: config/configmaps/
...
 
 gathers cluster Federated Monitoring metrics.
 
 The GET REST query to URL /federate
 Gathered metrics:
-  ALERTS
   etcd_object_counts
   cluster_installer
   namespace CPU and memory usage
+  followed by at most 1137 lines of ALERTS metric
```
iNecas, line 179: Needs update after changing the limit (ideally together with the commit message)

natiiix: Good point; done.

PR #154, opened 2020-08-06
Add @natiiix to OWNERS

natiiix:
I have been working on the Insights Operator for a while now and having the permissions would be helpful. Right now I can't use many of the OpenShift bot commands. Ricardo will soon, hopefully, start working on the Insights Operator as well, so it may be for the best to add both of us at once.


natiiix:
/retitle Bug 1869312: Add @natiiix to OWNERS

PR #155, opened 2020-08-10
TestCSRCollected

psimovec:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1835090

Q: Is checking if CSR is collected enough? or we also need to check if the collected CSR is anonymized?

### DIFF
--- a/test/integration/bugs_test.go
+++ b/test/integration/bugs_test.go
@@ -2,6 +2,7 @@ package integration
 
 import (
 	"fmt"
+	"k8s.io/api/certificates/v1beta1"
 	"regexp"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


test/integration/bugs_test.go
```diff
@@ -201,6 +202,37 @@ func TestArchiveContains(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+		Spec:       v1beta1.CertificateSigningRequestSpec{Request: certificateRequest},
+		Status:     v1beta1.CertificateSigningRequestStatus{},
+	})
+	e(t, err, "Failed creating certificate signing request")
+	defer func() {
+		clientset.CertificatesV1beta1().CertificateSigningRequests().Delete(name, &metav1.DeleteOptions{})
+		restartInsightsOperator(t)
+	}()
+	defer ChangeReportTimeInterval(t, 1)()
```
lina-nikiforova, line 229: why do we need to change report time interval after the test? It wasn't changed during the test

psimovec: ChangeReportTimeInterval returns callback function which reverts the changes. The report time interval is changed on this line, only callback function is defered. It's equivalent to:
`revertTimeInterval := ChangeReportTimeInterval(t, 1)`
`defer revertTimeInterval()`

lina-nikiforova: ok, thanks!

lina-nikiforova:
lgtm

PR #157, opened 2020-08-11
TestUploadNotDelayedAfterStart

psimovec:
https://bugzilla.redhat.com/show_bug.cgi?id=1841057

### DIFF
--- a/test/integration/bugs_test.go
+++ b/test/integration/bugs_test.go
@@ -22,6 +22,7 @@ const knownFileSuffixesInsideArchiveRegex string = `(`+
 
 //https://bugzilla.redhat.com/show_bug.cgi?id=1841057
 func TestUploadNotDelayedAfterStart(t *testing.T) {
+	checkPodsLogs(t, clientset, `It is safe to use fast upload`)
 	time1:=logLineTime(t, `Reporting status periodically to .* every`)
 	time2:=logLineTime(t, `Successfully reported id=`)
 	delay := time2.Sub(time1)
--- a/test/integration/bugs_test.go
+++ b/test/integration/bugs_test.go
@@ -20,6 +20,18 @@ const knownFileSuffixesInsideArchiveRegex string = `(`+
 	`(\/|^)(config|id|invoker|metrics|version)` +
 `)$`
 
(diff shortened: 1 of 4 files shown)


test/integration/main_test.go
```diff
@@ -183,7 +183,16 @@ func deleteAllPods(t *testing.T, namespace string) {
 	t.Log(errPod)
 }
 
-func checkPodsLogs(t *testing.T, kubeClient *kubernetes.Clientset, message string, newLogsOnly ...bool) {
+func logLineTime(t *testing.T, pattern string) time.Time {
+	str := checkPodsLogs(t, clientset, `\d+:\d+:\d+\.\d+.*`+pattern)
```
martinkunc, line 187: Maybe it would be a bit safer to read from beginning of the line, like:
`^\S\d{2}\d{2}\s\d+:\d+:\d+\.\d+.*`
to parse pattern:
`I0721 13:37:20.340873`
This could help to skip potential lines which have time in the middle of log message

psimovec: nice catch!
unfortunately, there are no newlines in captured logs, which is quite weird.. so "^" does not work. I made it at least more specific with an additional check. Now there would be problem if log line contained whole different log line, which I don't expect to happen much...

psimovec: It should work now

test/integration/bugs_test.go
```diff
@@ -20,6 +20,18 @@ const knownFileSuffixesInsideArchiveRegex string = `(`+
 	`(\/|^)(config|id|invoker|metrics|version)` +
 `)$`
 
+//https://bugzilla.redhat.com/show_bug.cgi?id=1841057
+func TestUploadNotDelayedAfterStart(t *testing.T) {
+	time1:=logLineTime(t, `Reporting status periodically to .* every`)
```
martinkunc, line 25: Nice, Could we please just for sure check that IO operator is healthy at the beginning of TC, because that is precondition for fast upload mode.
Also, at the beginning of Log file should be `It is safe to use fast upload` and not `Not safe for fast upload`, but I guess checking health of operator should be good enough fot test. (To not to have investigate ccases when AWS or OCP itself failed).

PR #158, opened 2020-08-11
Decrease insights secret interval minimal duration

psimovec:
Decreasing the minimal duration of insights secret interval can significantly speed-up e2e tests


psimovec:
test insights-operator-e2e-tests

PR #159, opened 2020-08-11
Fix TestArchiveContains/File: Check if insights operator records an event

psimovec:
Checking if insights operator records an event file before checking that the records were written to the disk should prevent running tests with an older archive that doesn't have new records yet

### DIFF
--- a/test/integration/bugs_test.go
+++ b/test/integration/bugs_test.go
@@ -181,6 +181,7 @@ func TestArchiveContains(t *testing.T) {
 
 	defer ChangeReportTimeInterval(t, 1)()
 	defer degradeOperatorMonitoring(t)()
+	checkPodsLogs(t, clientset, `Recording events/openshift-monitoring`, true)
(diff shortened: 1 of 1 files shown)


test/integration/main_test.go
```diff
@@ -184,12 +184,13 @@ func deleteAllPods(t *testing.T, namespace string) {
 }
 
 func checkPodsLogs(t *testing.T, kubeClient *kubernetes.Clientset, message string, newLogsOnly ...bool) {
+	timeNow := metav1.NewTime(time.Now())
+	timeNow.Add(-2*time.Second)
```
martinkunc, line 188: If we use newLogsOnly=false, it could be safer then rely on (arbitrary) time before Now

psimovec: This would cause problems and probably break the tests even more.  If I use newLogsOnly=false, then it’s almost certain that that the check for `.*Wrote \d+ records to disk in \d+` will pass immediately, as it will be in logs already.. so tests of Logs and Event will get old archive again and fail. But I agree that it’s not the best solution now, I will try to find a different one.

PR #161, opened 2020-08-12
Bug 1871826: Updating archive and Generated doc

martinkunc:
To actualize currently gathered data, updating the sample.
Freshly collected data from a cluster are triggering lot of changes in samples


martinkunc:
/retitle Bug 1871826: Updating archive and Generated doc

PR #163, opened 2020-08-18
Bug 1882361: Get report from smart-proxy and expose overview as a metric

joselsegura:
Adding the ability of retrieving reports for the cluster from Insights Results Smart Proxy and expose an overview of the reported rules as Prometheus metrics


pkg/config/config.go
```diff
@@ -7,11 +7,15 @@ import (
...
-	StoragePath string `json:"storagePath"`
-	Interval    string `json:"interval"`
-	Endpoint    string `json:"endpoint"`
-	Impersonate string `json:"impersonate"`
+	Report           bool   `json:"report"`
+	StoragePath      string `json:"storagePath"`
+	Interval         string `json:"interval"`
+	Endpoint         string `json:"endpoint"`
+	Impersonate      string `json:"impersonate"`
+	SmartProxyConfig struct {
```
tisnik, line 15: probably worth to define this as a proper data type?

pkg/config/config.go
```diff
@@ -32,6 +39,19 @@ func (s *Serialized) ToController() (*Controller, error) {
 	if cfg.Interval <= 0 {
 		return nil, fmt.Errorf("interval must be a non-negative duration")
 	}
+
+	if len(s.SmartProxyConfig.PollTime) > 0 {
```
tisnik, line 43: or `s.SmartProxyConfig.PollTime != ""` if I understand correctly

pkg/controller/operator.go
```diff
@@ -167,6 +169,9 @@ func (s *Support) Run(ctx context.Context, controller *controllercmd.ControllerC
 	// know any previous last reported time
 	go uploader.Run(ctx)
 
+	reportUpdater := externalpipeline.New(s.Controller.SmartProxy, authorizer, configPeriodic)
+	go reportUpdater.Run()
```
tisnik, line 173: it will run w/o ability to stop this goroutine. Probably this is what we want to achieve (idk)

pkg/externalpipeline/externalpipeline.go
```diff
@@ -0,0 +1,165 @@
...
+	ClusterVersion() *configv1.ClusterVersion
+}
+
+type Authorizer interface {
+	Authorize(req *http.Request) error
+	NewSystemOrConfiguredProxy() func(*http.Request) (*url.URL, error)
+}
+
+var (
+	// ErrWaitingForVersion contains an error when the clusver version is not loaded yet
```
tisnik, line 45: clusver/cluster

pkg/externalpipeline/externalpipeline.go
```diff
@@ -0,0 +1,165 @@
...
+
+		reportResponse := InsightsReportResponse{}
+		if err = json.Unmarshal(body, &reportResponse); err != nil {
+			klog.Error("The report response cannot be parsed")
+		} else {
+			updateInsightsMetrics(reportResponse.Report)
+			r.LastReport = reportResponse.Report
+		}
+	}
+
```
tisnik, line 118: prob. log info about StatusCode that is different from StatusOK?

tisnik (approved):
LGTM, just have some comments...

(and the vendoring part is HUGE)

pkg/externalpipeline/externalpipeline.go
joselsegura: Agree. You will see it on the next commit

pkg/externalpipeline/externalpipeline.go
joselsegura: +1

pkg/externalpipeline/externalpipeline.go
```diff
@@ -0,0 +1,167 @@
...
+	klog.Infof("Endpoint: %s", endpoint)
+	if err != nil {
+		klog.Error(err)
+		return err
+	}
+
+	if req.Header == nil {
+		req.Header = make(http.Header)
+	}
+	req.Header.Set("User-Agent", fmt.Sprintf("insights-operator/%s cluster/%s", version.Get().GitCommit, cv.Spec.ClusterID))
```
iNecas, line 87: Shouldn't the https://github.com/openshift/insights-operator/blob/master/pkg/insights/insightsclient/insightsclient.go be made a bit more generic to handle other requests as well? Seems like quite a lot of common code, and things like configuring a proxy is not even handled here: I suppose it's right expectation for the same transport mechanism for upload&pull the data: the only thing that differs is the endpoint/verb, doesn't it?

pkg/config/config.go
joselsegura: Done

pkg/config/config.go
joselsegura: Yes, you are right, done

pkg/externalpipeline/externalpipeline.go
joselsegura: I didn't study the source code of that package, but probably both are Insights related, so it will make sense to extract "common factor" and use this common code from both packages. Probably it worths to move the new `externalpipeline` package into the `insights` one, so some of the interface and common code should be easily reused.

I will take a look into it and try my best to see what can be done.

Right now I'm just fighting against the malevolous CI

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -217,6 +217,50 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
...
+
+	req, err := http.NewRequest(http.MethodGet, endpoint, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	if req.Header == nil {
+		req.Header = make(http.Header)
+	}
+	req.Header.Set("User-Agent", fmt.Sprintf("insights-operator/%s cluster/%s", version.Get().GitCommit, cv.Spec.ClusterID))
```
iNecas, line 238: There is still quite a lot of common code between this and `Send` regarding preparation of the client and headers. 

What if we had a common method `doReq(ctx, req)` that would handle the client and headers preparation,  eliminating the code here to:

```go
// …
req, err := http.NewRequest(http.MethodGet, endpoint, nil)
if err != nil {
    return nil, err
}
resp, err := c.DoReq(ctx, req)
// …
```

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	klog.Info("Parsing report")
+	reportResponse := InsightsReportResponse{}
+	
+	if err = json.NewDecoder(*reportBody).Decode(&reportResponse); err != nil {
+		klog.Error("The report response cannot be parsed")
+		return
+	} else {
+		updateInsightsMetrics(reportResponse.Report)
```
iNecas, line 81: Common pattern in go is just return early from error checking  and continue without else, seems we can do this here as well.

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,124 @@
...
+
+	insightsStatus.WithLabelValues("low").Set(float64(low))
+	insightsStatus.WithLabelValues("moderate").Set(float64(moderate))
+	insightsStatus.WithLabelValues("important").Set(float64(important))
+	insightsStatus.WithLabelValues("critical").Set(float64(critical))
+	insightsStatus.WithLabelValues("total").Set(float64(total))
+}
+
+func init() {
+	if err := legacyregistry.Register(
```
iNecas, line 119: This would feel more go-idiomatic

```go
	err := legacyregistry.Register(insightsStatus)
        if err != nil {
	    fmt.Println(err)
	}
```

pkg/externalpipeline/externalpipeline.go
joselsegura: It is now handled by `insightsclient` package

pkg/insights/insightsreport/insightsreport.go
joselsegura: Sure, we have some linters in other projects to warn about it. TYVM!

pkg/insights/insightsreport/insightsreport.go
joselsegura: Done

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	klog.V(4).Info("Smart Proxy report correctly parsed")
+
+	if r.LastReport.Meta.LastCheckedAt == reportResponse.Report.Meta.LastCheckedAt {
+		klog.V(2).Info("Retrieved report is equal to previus one. Retrying...")
+		return
+	}
+
+	defer close(r.insightsReport)
```
iNecas, line 109: why deferring here, rather than just doing it as the last statement?

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,164 @@
...
+	klog.V(2).Info("Starting report retriever")
+
+	for {
+		// always wait for new uploaded archive or insights-operator ends
+		select {
+		case <-r.archiveUploadReporter:
+			// When a new archive is uploaded, try to get the report and repeat every 30s
+			// until the report is retrieved
+			r.insightsReport = make(chan struct{})
+			wait.Until(r.PullSmartProxy, time.Duration(30e9), r.insightsReport)
```
iNecas, line 127: I smell a problem here in case the report doesn't get processed in time for some reason. What is current time between upload/processing? What would happen when the archive doesn't get processed for some reasons?

So two things:

* 30s seems quite aggressive polling, peronsaly, I would start with higher intervals and making it shorter if needed: making it configurable similarly to other intervals is also a good idea
* top limit on amount of iterations to try
*

iNecas:
Where is the `github.com/RedHatInsights/insights-operator-utils` used in the latest version?

pkg/insights/insightsreport/insightsreport.go
iNecas: While on configuration: ability to turn this off is a must IMO

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,164 @@
+package insightsreport
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/RedHatInsights/insights-results-smart-proxy/types"
```
iNecas, line 9: I'm a bit ambivalent here: OTOH I see the motivation of having one place where the structure is defined, on the other, this brings quite a big package of vendors with it. What do you think @tisnik? 

An altnerative would be defining the type here, with only the fields we're intersted in here. The benefit would seeing immediately what data are used. The downside is type control not telling us that in case something changers on the definition side. OTOH once this gets to real clusters we would need more or less freeze the changes there anyway, as getting the changes though to the clusters would take some time.

pkg/insights/insightsuploader/insightsuploader.go
```diff
@@ -157,6 +159,7 @@ func (c *Controller) Run(ctx context.Context) {
 			}
 			c.reporter.SetSafeInitialStart(false)
 			klog.V(4).Infof("Uploaded report successfully in %s", time.Now().Sub(start))
+			c.archiveUploaded <- struct{}{}
```
iNecas, line 162: Would would happen when the gaherer was not consuming from this channel and should the uploader care?

pkg/insights/insightsreport/insightsreport.go
tisnik: @iNecas my personal feeling is that we might copy the structure into IO operator code. And Rob Pike seems not to disagree https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s

There are cons, of course, but the structure itself is pretty stable right now.

natiiix (commented):
Ivan made a couple of valid points that I'm not gonna repeat. The configuration certainly seems important. I don't really think that vendoring the Smart Proxy to get the types would be a big deal and I'd prefer it over copying code, but I can obviously see the problem with either solution. Other than that, the rest looks fine to me.

pkg/insights/insightsreport/insightsreport.go
joselsegura: Got it. I just copy/pasted it from the top of the function. Then I realized that I only want to close it when the function reach one specific point after some checks, so I just moved it

joselsegura: Obviously, I will fix this one on my next commit

pkg/insights/insightsreport/insightsreport.go
joselsegura: Answering your question, I found some examples and it will take more or less 3-4 seconds since the archive is successfully uploaded to Ingress, but it will always depend on the time it takes to process the archive. As the `wait.Until` will start as soon as the archive is uploaded, it is very likely that the first iteration will not find the expected report, but the second one should found it with the current 30s period between retries.

If the archive is not processed, the `wait.Until` will be retrying forever since a new archive is uploaded and correctly processed.

I agree with your suggestions, si I will continue working on that direction

pkg/insights/insightsreport/insightsreport.go
joselsegura: As a newcomer in the Golang world, I would like to follow the best practices possible in the Golang ecosystem, so if you think that copying the struct is better than referencing it, I will do it on that way :-D

pkg/insights/insightsuploader/insightsuploader.go
joselsegura: As far as I can tell about Golang channels, there is no problem if nobody reads the channel, but maybe I'm wrong

pkg/insights/insightsreport/insightsreport.go
martinkunc, line 127: It is better to use 10*time.Second than time.Duration(30e9).

joselsegura: Thank you Martin. I'm working currently on several updates for this PR and definitely it will on the boat :-)

pkg/insights/insightsclient/insightsclient.go
joselsegura: Can we consider this as "resolved"? I'm just reviewing all the comments that the PR already had to make some clean up

joselsegura:
I just resolved some of the comments because they are nitpicks already fixed in the new commits. Don't hesitate to reopen if I missed something

pkg/insights/insightsuploader/insightsuploader.go
iNecas: The problem is it would get stuck on trying to write

iNecas: https://gobyexample.com/non-blocking-channel-operations

iNecas:
Probably we don't need to touch vending as we're not using the dependency anymore

natiiix (approved):
/lgtm

I have not really tested this locally beucase I don't currently have a cluster provisioned.

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,187 @@
...
+
+	updateInsightsMetrics(reportResponse.Report)
+	r.LastReport = reportResponse.Report
+	// close(r.insightsReport)
+	return true, nil
+}
+
+// Run goroutine code for gathering the reports from Smart Proxy
+func (r *Gatherer) Run(ctx context.Context) {
+	r.Simple.UpdateStatus(controllerstatus.Summary{Healthy: true})
```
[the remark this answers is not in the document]
martinkunc: Sorry, I noticed the channel archiveUploadReporter is no in select, that would be better than checking LastReportedTime.

pkg/config/configobserver/configobserver.go
```diff
@@ -139,6 +139,27 @@ func (c *Controller) retrieveConfig() error {
 		if noproxy, ok := secret.Data["noProxy"]; ok {
 			nextConfig.HTTPConfig.NoProxy = string(noproxy)
 		}
+		if reportEndpoint, ok := secret.Data["reportEndpoint"]; ok {
+			nextConfig.ReportEndpoint = string(reportEndpoint)
+		}
+		if reportPullingDelay, ok := secret.Data["reportPullingDelay"]; ok {
+			nextConfig.ReportPullingDelay, err = time.ParseDuration(string(reportPullingDelay))
```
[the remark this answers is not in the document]
joselsegura: Good point. TYVM!

pkg/config/configobserver/configobserver.go
```diff
@@ -216,6 +237,18 @@ func (c *Controller) mergeConfigLocked() {
 		if len(c.secretConfig.Endpoint) > 0 {
 			cfg.Endpoint = c.secretConfig.Endpoint
 		}
+		if len(c.secretConfig.ReportEndpoint) > 0 {
+			cfg.ReportEndpoint = c.secretConfig.ReportEndpoint
+		}
+		if c.secretConfig.ReportPullingDelay > 0 {
```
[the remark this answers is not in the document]
joselsegura: Nice catch

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -114,22 +114,32 @@ func clientTransport(authorizer Authorizer) http.RoundTripper {
 	return transport.DebugWrappers(clientTransport)
 }
 
-func (c *Client) Send(ctx context.Context, endpoint string, source Source) error {
+func (c Client) prepareRequest(method string, endpoint string, ctx context.Context) (*http.Request, error) {
```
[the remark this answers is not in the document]
joselsegura: Sure! I was not aware of that :-)

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -114,22 +114,32 @@ func clientTransport(authorizer Authorizer) http.RoundTripper {
...
 	cv := c.clusterInfo.ClusterVersion()
```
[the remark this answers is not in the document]
joselsegura: Agree

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -216,6 +225,40 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
...
+
+	klog.V(4).Infof("Retrieving report from %s", req.URL.String())
+	resp, err := c.client.Do(req)
+
+	if err != nil {
+		klog.Errorf("Unable to retrieve latest report for %s: %v", cv.Spec.ClusterID, err)
+		return nil, err
+	}
+
+	if resp.StatusCode == http.StatusOK {
```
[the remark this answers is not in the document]
joselsegura: I will add same error handling for status codes != 2XX

Unmarshalling the response looks a bit out of context here, as we are only handling HTTP related stuff in the `insightsclient`.

I will add a new metric.

pkg/insights/insightsreport/types.go
```diff
@@ -0,0 +1,43 @@
...
+
+// ReportResponseMeta contains metadata about the report
+type ReportResponseMeta struct {
+	Count         int       `json:"count"`
+	LastCheckedAt Timestamp `json:"last_checked_at"`
+}
+
+// RuleWithContentResponse represents a single rule in the response of /report endpoint
+type RuleWithContentResponse struct {
+	RuleID       RuleID      `json:"rule_id"`
```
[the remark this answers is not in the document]
joselsegura: This would be a new endpoint into the Smart Proxy. I will discuss it with the team

pkg/insights/insightsreport/types.go
```diff
@@ -0,0 +1,43 @@
...
+	Description  string      `json:"description"`
+	Generic      string      `json:"details"`
+	Reason       string      `json:"reason"`
+	Resolution   string      `json:"resolution"`
+	TotalRisk    int         `json:"total_risk"`
+	RiskOfChange int         `json:"risk_of_change"`
+	Disabled     bool        `json:"disabled"`
+	Internal     bool        `json:"internal"`
+	UserVote     UserVote    `json:"user_vote"`
+	TemplateData interface{} `json:"extra_data"`
```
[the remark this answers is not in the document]
joselsegura: Probably I can remove almost all the unmarshalled values because the only one needed for this feature is the `TotalRisk`. I'll update this struct in my next commit

pkg/config/configobserver/configobserver.go
```diff
@@ -139,6 +139,36 @@ func (c *Controller) retrieveConfig(ctx context.Context) error {
...
+			nextConfig.ReportPullingDelay, err = time.ParseDuration(string(reportPullingDelay))
+			if err != nil {
+				klog.Warningf(
+					"reportPullingDelay secret contains an invalid value (%s). Using previous value",
+					reportPullingDelay,
+				)
+			}
+		}
+		if reportPullingTimeout, ok := secret.Data["reportPullingTimeout"]; ok {
+			nextConfig.ReportPullingTimeout, err = time.ParseDuration(string(reportPullingTimeout))
```
martinkunc, line 155: Ufortunatelly you would have to use it with if. time.ParseDuration when fails is not returning the original value, but 0. See https://play.golang.com/p/-WRrTb5Xg5s
Correct would be:
```
if v, err := time.ParseDuration(string(reportPullingTimeout); err == nil {
  nextConfig.ReportPullingTimeout = v
}
```

pkg/insights/insightsclient/insightsclient.go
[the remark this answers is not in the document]
martinkunc: Are we using the new metric I think this is missing: https://github.com/openshift/insights-operator/blob/master/pkg/insights/insightsclient/insightsclient.go#L197

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,184 @@
...
+	initialWait := config.ReportPullingDelay
+	minimalRetryTime := config.ReportMinRetryTime
+	reportEndpoint := config.ReportEndpoint
+
+	if len(reportEndpoint) == 0 {
+		klog.V(4).Info("Not downloading report because Smart Proxy client is not properly configured: missing report endpoint")
+		return true, nil
+	}
+
+	delay := initialWait - time.Duration(r.retries*minimalRetryTime.Nanoseconds()) - minimalRetryTime
```
martinkunc, line 84: Is it intentionally waiting 60s - (0 x 10s) - 10s, 60 - (1 x 10s) - 10s, 60 - (2x10s) - 10s etc ?
what about 
```
pause := wait(retry int) time.Duration { 
  if retry ==0 {
  return wait.Jitter(initialWait, 0.1)
}
  return wait.Jitter(minRetryTime, 0.3)
}
 if wait > 0 {
  time.Sleep(pause)
}
```
To wait for 60 + rand(6s), 10 + rand(3s), 10 + rand(3)

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,184 @@
...
+	for {
+		// always wait for new uploaded archive or insights-operator ends
+		select {
+		case <-r.archiveUploadReporter:
+			// When a new archive is uploaded, try to get the report after waiting
+			// Repeat until the report is retrieved or maxIterations is reached
+			config := r.configurator.Config()
+			if config.ReportPullingTimeout != 0 {
+				wait.Poll(config.ReportMinRetryTime, config.ReportPullingTimeout, r.PullSmartProxy)
+				r.retries = 0
```
martinkunc, line 142: Should't the retires be set to 0 before calling wait.Poll ?

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,184 @@
...
+
+	klog.V(4).Info("Retrieving report")
+	reportBody, err := r.client.RecvReport(ctx, reportEndpoint)
+	if authorizer.IsAuthorizationError(err) {
+		r.Simple.UpdateStatus(controllerstatus.Summary{
+			Operation: controllerstatus.DownloadingReport,
+			Reason:    "NotAuthorized",
+			Message:   fmt.Sprintf("Auth rejected for downloading latest report: %v", err),
+		})
+		return false, err
```
martinkunc, line 101: I believe auth failure should be signal to finish the retrying.

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,184 @@
...
+	} else if err != nil {
+		klog.Errorf("Error retrieving the report: %s", err)
+		return false, err
```
martinkunc, line 104: Any other error should also stop retrying immediatelly.

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	klog.V(4).Info("Report retrieved")
+	reportResponse := Response{}
+
+	if err = json.NewDecoder(*reportBody).Decode(&reportResponse); err != nil {
+		klog.Error("The report response cannot be parsed")
+		return false, err
```
martinkunc, line 112: When it returns invalid resp, it needs to finish retry as well

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -0,0 +1,184 @@
...
+
+// Run goroutine code for gathering the reports from Smart Proxy
+func (r *Gatherer) Run(ctx context.Context) {
+	r.Simple.UpdateStatus(controllerstatus.Summary{Healthy: true})
+	klog.V(2).Info("Starting report retriever")
+
+	for {
+		// always wait for new uploaded archive or insights-operator ends
+		select {
+		case <-r.archiveUploadReporter:
```
martinkunc, line 136: another reaction should be r.configurator.ConfigChanged(). Ideally it would keep some minimal refresh rate, which could be hard to implement using this pattern and insightsuploader has better tools for it.

pkg/insights/insightsclient/insightsclient.go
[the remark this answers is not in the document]
martinkunc: It is still not implemented I think

pkg/config/configobserver/configobserver.go
martinkunc: It looks this is still not implemented no ?

joselsegura: I implemented in a different way, because I didn't get the point of your comment. Nxt commit will fix that. TYVM

pkg/insights/insightsreport/insightsreport.go
joselsegura: I will use Jitter in next commit. TYVM

pkg/insights/insightsreport/insightsreport.go
joselsegura: Agree, will be updated in next commit

pkg/insights/insightsreport/insightsreport.go
joselsegura: Not sure about this. I'll review the insightsclient code and check what errors can be returned

pkg/insights/insightsreport/insightsreport.go
joselsegura: Agree

pkg/insights/insightsreport/insightsreport.go
joselsegura: I get it. I'll keep working on that.

pkg/insights/insightsclient/insightsclient.go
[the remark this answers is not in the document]
joselsegura: I just pushed a commit using the new metric. I forgot to update it!

pkg/insights/insightsreport/insightsreport.go
martinkunc: If any other error then unauthorized happens,, it needs to return true, err to indicate that this function shouldn't be called again.

pkg/insights/insightsreport/insightsreport.go
joselsegura: New version includes a total refactor of the "polling loop". Instead of using one of the commodities in `wait` package, I implemented a new way that is aware of configuration changes.
The problem with configuration changes is that almost every config change will affect to the current timers, that are already set using `wait` functions.

martinkunc:
Thank you Jose, it looks good.
/lgtm

PR #166, opened 2020-08-21
Bug 1873101: Gather VolumeSnapshot CRD

natiiix:
[Jira: CCXDEV-2393](https://projects.engineering.redhat.com/browse/CCXDEV-2393)

*~~Perhaps the vendoring of K8s API Extensions module commit will go into a separate PR, but I included it for now for the sake of completeness.~~ That does not seem to be possible since all the tests fail without the vendoring.*


natiiix:
Okay, it won't work without including the vendoring of the K8s API extensions module.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -665,6 +669,35 @@ func GatherCertificateSigningRequests(i *Gatherer) func() ([]record.Record, []er
...
+// The following CRDs are gathered:
+// - volumesnapshots.snapshot.storage.k8s.io (10745 bytes)
+// - volumesnapshotcontents.snapshot.storage.k8s.io (13149 bytes)
+//
+// The CRD sizes above are in the raw (uncompressed) state.
+//
+// Location in archive: config/crd/
+func GatherCRD(i *Gatherer) func() ([]record.Record, []error) {
+	return func() ([]record.Record, []error) {
+		crds, err := i.crdClient.CustomResourceDefinitions().List(metav1.ListOptions{})
```
martinkunc, line 683: Would it be possible to instead of getting whole List, just get concrete CRD by Name ? It could be a bit more efficient I think.
Maybe something like:
```
crds := []string{"one","two"}
  for _, c := range crds {
    i.crdClient.CustomResourceDefinitions().Get(c,metav1.ListOptions{})
//... append
  }
```

natiiix: Oh, thanks for pointing this out, I totally didn't realize we could do this.

martinkunc:
/lgtm

Thank you, very nice :)

martinkunc:
@natiiix It looks the CRDs for volume snapshots are missing on 4.3 we would need to skip missing CRD gracefully and continue collection.

PR #162, opened 2020-08-18
[release-4.4] Bug 1869197: Limit collection of ALERTS metric to 1000 lines (~500KiB) to avoid unbearably large archives

natiiix:
Backport of #148 (and #153)

PR #169, opened 2020-08-31
[release-4.3] Bug 1874018: Limit collection of ALERTS metric to 1000 lines (~500KiB) to avoid unbearably large archives

natiiix:
Backport of #148 (upstream implementation), #153 (4.5 backport), #162 (4.4 backport)

*The test is missing because the 4.3 Insights Operator executed gathering functions in a vastly different way from the 4.4+ versions.*


tisnik (approved):
LGTM

PR #172, opened 2020-09-04
Bug 1878664: Adding CRD and some example and documentation

joselsegura:


docs/custom-resource-sample/report_overview1.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: "external-pipeline.redhat.com/v1"
+kind: ReportOverview
+metadata:
+  name: my-new-report-overview-object
```
bond95, line 4: nit-pick: I'm not sure about this name maybe better `report-overview-object`

tisnik (approved):
LGTM

docs/custom-resource-sample/report_overview1.yaml
joselsegura: I'll update it, but it is just an example :-D

manifests/08-report_overview_crd.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: report-overviews.external-pipeline.redhat.com
+spec:
+  group: external-pipeline.redhat.com
+  versions:
+    - name: v1
```
iNecas, line 8: I would be more conservative with marking this `v1alpha1` at this point.

natiiix (commented):
I don't mind anything personally, but the version should probably be alpha/beta as Ivan suggested.

joselsegura:
Agree on @iNecas and @natiiix comments. I'll update it ASAP

manifests/08-report_overview_crd.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: report-overviews.external-pipeline.redhat.com
+spec:
+  group: external-pipeline.redhat.com
```
martinkunc, line 6: Woudln't it make better sense to call the group for example "insights.redhat.com" ? This looks a bit dependent on one particular place. What do you think ?

manifests/08-report_overview_crd.yaml
```diff
@@ -0,0 +1,32 @@
...
+                moderate:
+                  type: integer
+                important:
+                  type: integer
+                critical:
+                  type: integer
+  scope: Cluster
+  names:
+    plural: report-overviews
+    singular: report-overview
```
martinkunc, line 29: It looks like most of the multiword names doesn't use dash, would "reportoverview" make sense ? List crds for example here: http://pastebin.test.redhat.com/901861

manifests/08-report_overview_crd.yaml
```diff
@@ -0,0 +1,32 @@
...
+      storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                low:
+                  type: integer
```
martinkunc, line 19: Could we add Description: What does Low property mean ..

joselsegura:
@smarterclayton could you please take a look on this PR? It adds a new CRD to the clusters to be used by Insights Operator. 

Thank you in advance

docs/custom-resource-sample/report_overview1.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: "insights.redhat.com/v1alpha"
+kind: ReportOverview
```
tremes, line 2: Isn't this too general? Wouldn't be better e.g `InsightsReportOverview`? Just asking.... :-)

PR #173, opened 2020-09-07
Fix test: Increase allowed delay in TestUploadNotDelayedAfterStart

psimovec:
The allowed delay was too strict, it sometimes fails with older version if the delay happens to be ~1min 30 secs, which shouldn't cause test failure.


tisnik (approved):
LGTM (giving 3 minutes are ok :)

martinkunc:
Thanks Pavel,
/lgtm

PR #174, opened 2020-09-08
Improve container image collection

natiiix:
See #133 

- Don't filter container images based on URL.
- Since filtering isn't possible, only the top 1000 most common images are collected to limit the size.
- The sha256 is collected instead of the full image specs when possible.

### DIFF
--- a/pkg/gather/clusterconfig/container_images.go
+++ b/pkg/gather/clusterconfig/container_images.go
@@ -3,8 +3,8 @@ package clusterconfig
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
-	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -15,12 +15,20 @@ import (
 	_ "k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 
 	"github.com/openshift/insights-operator/pkg/record"
+	"github.com/openshift/library-go/pkg/image/reference"
 )
 
 const (
 	// imageGatherPodLimit is the maximum number of pods that
 	// will be listed in a single request to reduce memory usage.
 	imageGatherPodLimit = 200
+
+	// containerImageLimit is the maximum number of container images to collect.
+	// On average, information about one image takes up roughly 100 raw bytes.
+	containerImageLimit = 1000
+
+	// yyyyMmDateFormat is the date format used to get a YYYY-MM string.
+	yyyyMmDateFormat = "2006-01"
 )
 
 // GatherContainerImages collects essential information about running containers.
@@ -40,12 +48,8 @@ func GatherContainerImages(g *Gatherer) func() ([]record.Record, []error) {
 func gatherContainerImages(coreClient corev1client.CoreV1Interface, ctx context.Context) ([]record.Record, []error) {
 	records := []record.Record{}
 
-	contInfo := ContainerInfo{
-		Images:     ContainerImageSet{},
-		Containers: PodsWithAge{},
-	}
-	// Cache for image indices in the collected set of container image.
-	image2idx := map[string]int{}
+	// Cache for the temporary image count list.
+	img2month2count := img2Month2CountMap{}
 
 	// Use the Limit and Continue fields to request the pod information in chunks.
 	continueValue := ""
@@ -64,22 +68,11 @@ func gatherContainerImages(coreClient corev1client.CoreV1Interface, ctx context.
 			if strings.HasPrefix(pod.Namespace, "openshift") && hasContainerInCrashloop(podPtr) {
 				records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/%s", pod.Namespace, pod.Name), Item: PodAnonymizer{podPtr}})
 			} else if pod.Status.Phase == corev1.PodRunning {
-				for _, container := range pod.Spec.Containers {
-					imageURL := container.Image
-					urlHost, err := forceParseURLHost(imageURL)
-					if err != nil {
-						klog.Errorf("unable to parse container image URL: %v", err)
-						continue
-					}
-					if imageHostRegex.MatchString(urlHost) {
-						imgIndex, ok := image2idx[imageURL]
-						if !ok {
-							imgIndex = contInfo.Images.Add(imageURL)
-							image2idx[imageURL] = imgIndex
-						}
-						contInfo.Containers.Add(pod.CreationTimestamp.Time, imgIndex)
-					}
-				}
+				startMonth := pod.CreationTimestamp.Time.UTC().Format(yyyyMmDateFormat)
+
+				gatherImages(startMonth, img2month2count, pod.Status.ContainerStatuses)
+				gatherImages(startMonth, img2month2count, pod.Status.InitContainerStatuses)
+				gatherImages(startMonth, img2month2count, pod.Status.EphemeralContainerStatuses)
 			}
 		}
 
@@ -89,6 +82,43 @@ func gatherContainerImages(coreClient corev1client.CoreV1Interface, ctx context.
 			break
 		}
 		continueValue = pods.Continue
+	}
+
+	// Transform map into a list for sorting.
+	imageCounts := []tmpImageCountEntry{}
+	for img, countMap := range img2month2count {
+		totalCount := 0
+		for _, count := range countMap {
+			totalCount += count
+		}
+		imageCounts = append(imageCounts, tmpImageCountEntry{
+			Image:         img,
+			TotalCount:    totalCount,
+			CountPerMonth: countMap,
+		})
+	}
+
+	// Sort images from most common to least common.
+	sort.Slice(imageCounts, func(i, j int) bool {
+		return imageCounts[i].TotalCount > imageCounts[j].TotalCount
+	})
+
+	// Reconstruct the image information into the reported data structure.
+	contInfo := ContainerInfo{
+		Images:     ContainerImageSet{},
+		Containers: PodsWithAge{},
+	}
+	totalEntries := 0
+	for _, img := range imageCounts {
+		if totalEntries >= containerImageLimit {
+			break
+		}
+
+		imgIndex := contInfo.Images.Add(img.Image)
+		for month, count := range img.CountPerMonth {
+			contInfo.Containers.Add(month, imgIndex, count)
+			totalEntries++
+		}
 	}
 
 	return append(records, record.Record{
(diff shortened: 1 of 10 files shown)


tisnik (approved):
Seems legit

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -251,10 +270,45 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
 
+		// Transform map into a list for sorting.
+		type tmpImageCount struct {
+			Image      string
+			StartMonth string
+			Count      int
+		}
+		imageCounts := []tmpImageCount{}
+		for img, countMap := range img2month2count {
+			for month, count := range countMap {
```
martinkunc, line 281: Because we want to count all images by ID, shouldn't the Count be Sum of all counts from all months ?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -251,10 +270,45 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
+		sort.Slice(imageCounts, func(i, j int) bool {
+			return imageCounts[i].Count > imageCounts[j].Count
+		})
+
+		// Reconstruct the image information into the reported data structure.
+		contInfo := ContainerInfo{
+			Images:     ContainerImageSet{},
+			Containers: PodsWithAge{},
+		}
+		for i, img := range imageCounts {
```
martinkunc, line 300: Because we would collect all image IDs and all Counts (sum of all counts per months) we would iterate imageCounts, but write individual records from each month in img2month2count for this image iD.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -224,22 +224,41 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
-						if err != nil {
-							klog.Errorf("unable to parse container image URL: %v", err)
-							continue
-						}
-						if imageHostRegex.MatchString(urlHost) {
-							imgIndex, ok := image2idx[imageURL]
-							if !ok {
-								imgIndex = contInfo.Images.Add(imageURL)
-								image2idx[imageURL] = imgIndex
+					gatherImages := func(containers []corev1.ContainerStatus) {
```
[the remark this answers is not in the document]
natiiix: I know the readability suffers, but one of the variables (a property of the pod object) is being captured into the inline function. It could be done otherwise, but this results in the cleanest calls. Also, the output slice is captured in there. The calls don't look very nice when I put it out and considering that inline functions are very common in the I.O. code, I thought it wasn't that big of an issue.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -224,22 +224,41 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
+						for _, container := range containers {
+							dockerRef, err := reference.Parse(container.Image)
+							if err != nil {
+								klog.Errorf("unable to parse container image specification: %v", err)
```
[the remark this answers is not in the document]
natiiix: Do you mean that when a stdlib error happens, as opposed to an error from the `client-go`, we should just log it, but not return it?

pkg/gather/clusterconfig/clusterconfig.go
natiiix: If I do that, it will become even more difficult to split that information apart after the sorting is done and the temporary structure will become quite messy (name, total count and a dictionary of month:count, from the top of my head). I guess it's true, but neither way is optimal. Either way, we should put much more emphasis on the old images, but there isn't any nice way of doing it.

pkg/gather/clusterconfig/clusterconfig.go
natiiix: If I change it as mentioned above, then sure, that would be the way to do it, but I feel like the code will become even less readable as a result. I will try to implement it and show you what it would look like.

pkg/gather/clusterconfig/clusterconfig.go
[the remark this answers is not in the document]
martinkunc: Yes, basically I wouldnt fail unless it is life critical in the gathering and log it instead.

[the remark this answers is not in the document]
natiiix: So even if some data will be missing because we're not able to pre-process them, just log it and continue?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -237,22 +237,41 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
+								gatherErrors = append(gatherErrors, err)
```
martinkunc, line 245: I think I would ignore parsing errors at all and return nil instead down there and return only successfully collected list (the non nill error would affect status of operator I think).

natiiix: Okay, I remember we talked about this, but I forgot to implement it. It's a fair point.

pkg/gather/clusterconfig/clusterconfig.go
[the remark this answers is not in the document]
mfojtik: you can pass the variable as function parameter, I agree this is pretty long to be inline, but it is nit

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -279,22 +277,40 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
-						if imageHostRegex.MatchString(urlHost) {
-							imgIndex, ok := image2idx[imageURL]
-							if !ok {
-								imgIndex = contInfo.Images.Add(imageURL)
-								image2idx[imageURL] = imgIndex
+					gatherImages := func(containers []corev1.ContainerStatus) {
+						for _, container := range containers {
+							dockerRef, err := reference.Parse(container.Image)
+							if err != nil {
+								klog.Errorf("unable to parse container image specification: %v", err)
```
mfojtik, line 284: we usually don't use klog.Errof bug klog.Warningf for these... also klog messages should always start with upper case

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -279,22 +277,40 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
+								continue
+							}
+
+							// Use the sha256 hash ID if available, otherwise use the full image spec.
+							imgMinimal := dockerRef.ID
+							if imgMinimal == "" {
+								imgMinimal = container.Image
+							}
+
+							const yyyyMmFormat = "2006-01"
```
mfojtik, line 294: move this outside

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -279,22 +277,40 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
+							startMonth := pod.CreationTimestamp.Time.UTC().Format(yyyyMmFormat)
+
+							if countMap, ok := img2month2count[imgMinimal]; ok {
+								var count int
```
mfojtik, line 298: nit: `count := 0`

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -8,16 +8,15 @@ import (
...
 	"regexp"
 	"sort"
 	"strings"
 	"sync"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	apixv1beta1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	apixv1beta1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
```
mfojtik, line 19: use `v1`

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -306,6 +322,48 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
 			continueValue = pods.Continue
 		}
 
+		// Transform map into a list for sorting.
+		type tmpImageCount struct {
```
mfojtik, line 326: do not inline types like this, move this outside the func

pkg/gather/clusterconfig/clusterconfig.go
natiiix: Is it really a better way of declaring a variable that doesn't need a known default value? Of course, it will be initialized to `0` by default anyways, but to me, this syntax better indicates that it's supposed to be set later in the code and that the initial value shouldn't be used.

pkg/gather/clusterconfig/clusterconfig.go
natiiix: Fair enough, my logic with both the function and the struct were that I didn't like polluting the namespace with all these helper stuff that's only used once.

pkg/gather/clusterconfig/clusterconfig.go
natiiix: This change is completely unrelated to the PR; just a reshuffle of the import order, but I'll fix it anyways because it wasn't backported (yet).

pkg/gather/clusterconfig/clusterconfig.go
natiiix: Good point, I didn't notice that.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -334,6 +324,43 @@ func GatherContainerImages(i *Gatherer) func() ([]record.Record, []error) {
...
+			}
+			imageCounts = append(imageCounts, tmpImageCountEntry{
+				Image:         img,
+				TotalCount:    totalCount,
+				CountPerMonth: countMap,
+			})
+		}
+
+		// Sort images from most common to least common.
+		sort.Slice(imageCounts, func(i, j int) bool {
```
jasinner, line 342: It would be interesting to see the a sample of the data here, because I suspect the most common containers will be end-user application containers rather than the RH shipped containers which we are interested in writing rules for. If that sample shows that they are end-user application containers then we could perhaps inverse this. However better again would be to get the currently running containers in my opinion.

jasinner: Ok, my mistake, seems you are actually getting the current running containers. If that still exceeds the max image count on target clusters than we probably need to re-think this PR and stick with the current method of getting the image ref from the pod spec.

An alternative might be to lookup the container in Pyxis using the sha, and then use the returned data to filter based on image ref or build labels from there. It depends what's more important, time to gather data, or the size of data transmitted for post-processing. If time to gather is not an issue, using Pyxis to refine the collected data would seems like a valid option to me.

natiiix: As mentioned in my email, we have considered this, but to avoid any unnecessary load, we are trying to keep the collection as simple as possible, at least for the moment being.

jasinner: Let's see if this method will get us the containers we need. If not, we can adjust the filtering later.

tremes:
I experimented with it against my cluster and I think it works well and makes sense. Thanks @natiiix.

PR #175, opened 2020-09-08
Bug 1881907: Adds cluster version into the User-Agent header

0sewa0:


pkg/insights/insightsclient/insightsclient.go
```diff
@@ -128,7 +128,7 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
 	if req.Header == nil {
 		req.Header = make(http.Header)
 	}
-	req.Header.Set("User-Agent", fmt.Sprintf("insights-operator/%s cluster/%s", version.Get().GitCommit, cv.Spec.ClusterID))
+	req.Header.Set("User-Agent", fmt.Sprintf("insights-operator/%s cluster/%s version/%s", version.Get().GitCommit, cv.Spec.ClusterID, cv.Status.Desired.Version))
```
rluders, line 131: The `cv.Status.Desired.Version` doesn't give you the current version.

rluders (changes requested):

tisnik (approved):
It looks ok (giving the `?` string is expected by consumers)

rluders (approved):
lgtm

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -114,6 +114,15 @@ func clientTransport(authorizer Authorizer) http.RoundTripper {
 	return transport.DebugWrappers(clientTransport)
 }
 
+func getCurrentVersion(cv* configv1.ClusterVersion) string {
+	for _, history := range cv.Status.History {
+		if history.State == "Completed" {
+			return history.Version
```
iNecas, line 120: The cluster operator resoruce has `versions` field, that is probably more relevant for this (and doesn't rely on "completed" state), see https://docs.openshift.com/container-platform/4.4/rest_api/config_apis/clusteroperator-config-openshift-io-v1.html

Example from real cluster:

```
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  annotations:
    exclude.release.openshift.io/internal-openshift-hosted: 'true'
  creationTimestamp: '2020-09-06T01:27:05Z'
  generation: 1
  managedFields:
    - apiVersion: config.openshift.io/v1
      fieldsType: FieldsV1
      fieldsV1:
        'f:metadata':
          'f:annotations':
            .: {}
            'f:exclude.release.openshift.io/internal-openshift-hosted': {}
        'f:spec': {}
        'f:status': {}
      manager: cluster-version-operator
      operation: Update
      time: '2020-09-06T01:27:05Z'
    - apiVersion: config.openshift.io/v1
      fieldsType: FieldsV1
      fieldsV1:
        'f:status':
          'f:conditions': {}
          'f:extension': {}
          'f:relatedObjects': {}
          'f:versions': {}
      manager: insights-operator
      operation: Update
      time: '2020-09-06T03:04:27Z'
  name: insights
  resourceVersion: '24106'
  selfLink: /apis/config.openshift.io/v1/clusteroperators/insights
  uid: 77f6eb92-ad73-4fde-8eb8-f39edc96e033
spec: {}
status:
  conditions:
    - lastTransitionTime: '2020-09-06T03:04:11Z'
      status: 'False'
      type: Degraded
    - lastTransitionTime: '2020-09-06T03:04:11Z'
      status: 'True'
      type: Available
    - lastTransitionTime: '2020-09-06T03:04:27Z'
      message: Monitoring the cluster
      status: 'False'
      type: Progressing
    - lastTransitionTime: '2020-09-06T03:04:27Z'
      status: 'False'
      type: Disabled
  extension: {}
  relatedObjects:
    - group: ''
      name: openshift-insights
      resource: namespaces
    - group: apps
      name: insights-operator
      namespace: openshift-insights
      resource: deployments
  versions:
    - name: operator
      version: 4.5.7

```

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -128,7 +137,7 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
 	if req.Header == nil {
 		req.Header = make(http.Header)
 	}
-	req.Header.Set("User-Agent", fmt.Sprintf("insights-operator/%s cluster/%s", version.Get().GitCommit, cv.Spec.ClusterID))
+	req.Header.Set("User-Agent", fmt.Sprintf("insights-operator/%s cluster/%s version/%s", version.Get().GitCommit, cv.Spec.ClusterID, getCurrentVersion(cv)))
```
iNecas, line 140: @jhjaggars you mentioned some changed needed on ingress side before we can add additional fields to user-agent header. Mind adding more details here?

pkg/insights/insightsclient/insightsclient.go
0sewa0: But which version are we interested in ? 
The version of openshift running on the cluster or the version of the insights operator or they are the same thing ?

iNecas: It's usually the same, but even if it wasn't, the version of the operator is the thing we're interested in. The reason is to be able to do some backward compatibility conditions and version of the client/insights-operator is more useful than version of the cluster.

pkg/insights/insightsclient/insightsclient.go
jhjaggars: Yes, the code that parses the user-agent will need to be updated to support the additional field in a backward compatible way.  Have you considered putting the version in the `insights-operator/$VERSION` spot already?

pkg/insights/insightsclient/insightsclient.go
0sewa0: > The version of openshift running on the cluster or the version of the insights operator or they are the same thing ?
This seems to be true, for all base openshift operators, but it wouldn't be true about the user's own clusteroperators.

So what I'm trying to say is that getting this info from the clusteroperators would be actually more complicated(ie.: it would require more code changes).

0sewa0: :+1:

0sewa0: @iNecas Does it look right now ? :)

iNecas:
Looks good codewise but haven't a chance to try out: did anyone of the reviewers tried running this end2end?

0sewa0:
Don't know

iNecas:
Ok, I've tested via cluste-bot, as well as locally, and have seen the archvie getting though end2end, with the changes, so looks good from this perspective.

Code-wise, I know it follows an already established pattern, where the gatherer loads some info about the cluster and the client uses it when uploading the data. Said that, I don't like that approach a lot, as it couples the gathering and uploading in some unexpected way (what would happen if we had a configuration to not gather the cluster operators data).

Therefore, I'm good with acking this, as long as we have some refactoring task to decouple getting data for preparing the client from the generic gathering.

0sewa0:
The ClusterVersion is also from the gathering part, so I basically followed that pattern.  IMO probably most of insights operator should be refactored, or at least the gathering part. 
Few examples why:
- Working in a file that is more than 1200 lines long is not fun. (also its named `clusterconfig.go` why ??)
- Having `i` as the reference to the Gatherer object in methods is inconvenient.

martinkunc:
Good work, thank you!

Woudln't it be better to use RELEASE_VERSION env variable, which doesn't read value from cluster, but from Pod actually running ?

Maybe we could refactor the expression to method to return UserAgent value, together with waiting for ClusterVersion.
Also, I believe we could create the whole UserAgent string in semver form like:
```
func UserAgent(releaseVersionEnv, v apimachineryversion.Info, cv ClusterVersion) string
gitVersion := v.GitVersion
// Because we dont set GitVersion, just GitCommit, use kubernetes machinery, but fallback to RELEASE_VERSION envvar.
if gitVersion == version.GitVersion  { // default, not overwritten by build
  gitVersion = fmt.Sprintf("%s-%s", releaseVersionEnv, v.GitCommit)
}
return fmt.Sprintf("insights-operator/%s-%s cluster/%s",  gitVersion, cv.Spec.ClusterID)
}
```

0sewa0:
@iNecas what do you think about ^ ?

0sewa0:
@iNecas @martinkunc please review :)

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -114,6 +115,16 @@ func clientTransport(authorizer Authorizer) http.RoundTripper {
 	return transport.DebugWrappers(clientTransport)
 }
 
+func UserAgent(releaseVersionEnv string, v apimachineryversion.Info, cv *configv1.ClusterVersion) string {
```
iNecas, line 118: no need to export this function

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -114,6 +115,16 @@ func clientTransport(authorizer Authorizer) http.RoundTripper {
...
+	gitVersion := v.GitVersion
+	// If the RELEASE_VERSION is set in pod, use it
+	if releaseVersionEnv != "" {
+		gitVersion = releaseVersionEnv
+	}
+	gitVersion = fmt.Sprintf("%s-%s", gitVersion, v.GitCommit)
+	return fmt.Sprintf("insights-operator/%s cluster/%s", gitVersion, cv.Spec.ClusterID)
```
iNecas, line 125: Have you tried with real cluster, when trying against one, I'm getting `v0.0.0-master+$Format:%h$-$Format:%H$` in the `gitVersion`, which doesn't look right.

0sewa0: How can I try it on a 'real cluster'. I always run it locally while pointing at a cluster created by cluster-bot

0sewa0: So after a bit of investigation the build command doesn't seem to work as intended. The values the `version` returns are the defaults for that package and not the special ones that we should be setting during build.

`build:
	go build -mod=vendor -ldflags "-X vendor/k8s.io/client-go/pkg/version.gitCommit=$$(git rev-parse HEAD) -X vendor/k8s.io/client-go/pkg/version.gitVersion=v1.0.0+$$(git rev-parse --short=7 HEAD)" -o bin/insights-operator ./cmd/insights-operator`

martinkunc: Yes, that is right. That is why this change prefers RELEASE_VERSION envvar. If the var is set, it overrides the value built-in the package. Setting it should then affect Debug as well.
At some point later we would also like to resolve filling version in the ldflags, but I would do it in a separate CL.

0sewa0: I tested by using the envvar, which works but it still appends the `v.GitCommit` to the the value of the envvar and that will be`"$Format:%H$"`

martinkunc: When you build it with make build, it should set the gitVersion in the binary, when you run the binnary afterwards, will it have it set right ? It should.

0sewa0: conclusion: the `make build` had incorrect path to the packages where it tried to set the variables.

incorrect path: `github.com/openshift/insights-operator/vendor/k8s.io/client-go/pkg/version.gitCommit`
correct path: `k8s.io/client-go/pkg/version.gitCommit`

martinkunc:
Thank you, Marcell, good work.
/lgtm

PR #177, opened 2020-09-15
Bug 1879068: Collect hostsubnet information

tremes:
Gather host subnet information and anonymize all the IP addresses. The important/interesting part is the number of Egress IPs that can be obtained from this.


tremes:
@natiiix, @martinkunc, @iNecas can you please review?

iNecas:
What information exactly we're looking for here and why?

In general, we consider ip addresses and sensitive data and anonymize (see https://github.com/openshift/insights-operator/blob/3b7ba28926b16f5513ba8302613f7b2b31db76d8/pkg/gather/clusterconfig/clusterconfig.go#L1050) unless confirmed explicitly that it's not introducing security risk and justified that we really need the info.

tremes:
@iNecas Yep thanks. I realized I should anonymized that in sense of providing only count of egressIPs. I'll update.

tremes:
Anonymized and Updated.

natiiix (approved):
/lgtm

PR #181, opened 2020-09-18
[release-4.4] Bug 1880384: - collect hostsubnet information

tremes:


martinkunc:
@tremes To explain failing e2e: In 4.4 and earlier are missing this fix: https://github.com/openshift/insights-operator/commit/597d95839f637ea52ca80995e84ade77b0de3274#diff-84ddbb4357f406d84b9c0c51d88e985d

We were rerunning them few times to pass.

martinkunc:
@tremes the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1880352 is only in POST, not verified, after that our BZ could be refreshed to valid state.

PR #185, opened 2020-09-22
Bug 1881905: Adds gatherer for PodDistributionBudget

0sewa0:


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -1091,6 +1117,20 @@ func anonymizePod(pod *corev1.Pod) *corev1.Pod {
 	return pod
 }
 
+type PodDisruptionBudgetsAnonymizer struct {
+	*policyv1beta1.PodDisruptionBudget
+}
+
+// Marshal implements serialization of a Pod with anonymization
```
tremes, line 1124: Copy/paste comment. It's just minor but is there anything we could possibly anonymize (some IDs, IP addressess and similar)?

0sewa0: Yeah the comments are a mess but its WIP for a reason :D

About what to anonymize:  No idea really, the ones that I saw had nothing in them that I would consider confidential. maybe the `metadata.managedFields` can have something confidential info.

0sewa0:
currently some tests fail because:
`Sep 23 11:17:12.821: FAIL: Some cluster operators are not ready: insights (Degraded=True PeriodicGatherFailed: Source config could not be retrieved: poddisruptionbudgets.policy is forbidden: User "system:serviceaccount:openshift-insights:operator" cannot list resource "poddisruptionbudgets" in API group "policy" at the cluster scope)`

@iNecas @martinkunc @natiiix any thoughts ?

martinkunc:
Did this passed some local tests ? Basically you should get a cluster (quicklab) run IO and check data in /tmp folder.
Anyway, I think you need to add the Role policy according to this comment: https://github.com/rancher/rancher/issues/20325#issuecomment-496798677
in this file:
https://github.com/openshift/insights-operator/blob/master/manifests/03-clusterrole.yaml#L64

When you change file locally, you can try
```
oc apply -f 03-clusterrole.yaml
```
To apply the change to see if it works.

pkg/gather/clusterconfig/clusterconfig.go
martinkunc: Here is the type definition for PDB. To me it doesn't look like there could be anything worth anonymizing. What do you think ?
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/policy/types.go

0sewa0: that was my impression as well

natiiix (commented):
There are quite a few commits. It's usual to squash them to reduce the number of commits in the upstream branches. 1-3 is usually okay, but 6 seems like too many for a single enhancement.

0sewa0:
@natiiix Can't github do it ?? Gitlab made it all easy :(

natiiix:
> @natiiix Can't github do it ?? Gitlab made it all easy :(

@0sewa0 GitHub can do it, but the merge bot does as the name suggests, creates a merge commit, not a squash. Sometimes it's absolutely necessary for cherry-picking because you don't want to backport documentation or configuration changes in some scenarios.

martinkunc:
Nice, thank you @0sewa0 !
/lgtm

PR #186, opened 2020-09-22
Bug 1876799: Upgrade to k8s 0.18.9

martinkunc:
Updated Insights Operator to use 0.18.9 api where required. The new api is required by new openshift/client-go + to fix https://github.com/kubernetes/kubernetes/issues/91073
Update go.mod to use openshift/library-go to fix missing goautoneg 
Update vendors

Related update of openshift/release: https://github.com/openshift/release/pull/12162


natiiix (approved):
/lgtm

I just hope this won't break something somewhere else. :crossed_fingers:

PR #188, opened 2020-09-23
Bug 1881848: Do not return CRD not found error, just log it

natiiix:
Currently, if a CRD is missing from a cluster and the Insights Operator with unsucessfully attempt to collect it, it will crash. However, some CRDs are supposed to be missing, which would cause I.O. to repeatedly crash. Instead, these errors should only be logged and the error should not be reported.


tremes:
lgtm

PR #165, opened 2020-08-21
Bug 1882685: Log checker

psimovec:
Added log checker, which is more flexible than old function

PR #190, opened 2020-09-23
TestProxyOverride

psimovec:
https://bugzilla.redhat.com/show_bug.cgi?id=1820595
added TestProxyOverride - it runs for kinda long time ~10 mins
:beetle: [WIP, reset function doesn't always work properly] updated forceUpdateSecret to return "reset" function, added deleteSecret.. refactored some tests accordingly 
fixes bug with logChecker - logChecker did not fail when "failFast" was set, it just set .Err to the error

needs https://github.com/openshift/insights-operator/pull/165 merged first (OK)

forking the used tinyproxy container might be better option?

### DIFF
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -191,19 +191,60 @@ func duration(t *testing.T, start time.Time, end time.Time) float64 {
 }
 
 func LogChecker(t *testing.T) *LogCheck {
-	defaults := &LogCheck{
-		interval:   time.Second,
-		logOptions: corev1.PodLogOptions{},
-		timeout:    5 * time.Minute,
-		failFast:   true,
-		test:       t,
-		clientset:  clientset,
-	}
-	return defaults
+	return logChecker(t, clientset)
 }
 
 func checkPodsLogs(t *testing.T, message string) *LogCheck {
 	return LogChecker(t).Search(message)
+}
+
+func tinyproxy(t *testing.T) *TinyProxy {
+	proxy := &TinyProxy{}
+	err := proxy.create(t, clientset)
+	e(t, err, "failed to create tinyproxy")
+	proxy.waitUntilReady()
+	return proxy
+}
+func (proxy *TinyProxy) setClusterWideProxy(t *testing.T) {
+	// it will not work
+}
+func (proxy *TinyProxy) setIOProxyOverride(t *testing.T) {
+	modifiedSecret := corev1.Secret{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      Support,
+			Namespace: OpenShiftConfig,
+		},
+		Data: map[string][]byte{
+			"interval":   []byte("1m"), // for faster testing
+			"httpsProxy": []byte(proxy.adress),
+			"httpProxy":  []byte(proxy.adress),
+		},
+		Type: "Opaque",
+	}
+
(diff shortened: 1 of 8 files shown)


test/integration/main_test.go
```diff
@@ -191,44 +191,141 @@ func duration(t *testing.T, start time.Time, end time.Time) float64 {
...
+		NoProxy:            "example.com",
+		ReadinessEndpoints: []string{"ohno"},
+		TrustedCA: configv1.ConfigMapNameReference{
+			Name: "service.default",
+		},
+	},
+	}
+	cwproxy.Name = "cluster"
+	//version, _ := strconv.Atoi(oldProxy.ResourceVersion)
+	//version++
```
tremes, line 222: Some leftover.

test/integration/main_test.go
```diff
@@ -191,44 +191,141 @@ func duration(t *testing.T, start time.Time, end time.Time) float64 {
...
+	cwproxy.ObjectMeta.ResourceVersion = oldProxy.ResourceVersion //string(rune(version))
+
+	_, err := configV1Client().Proxies().Update(context.Background(), &cwproxy, metav1.UpdateOptions{})
+	e(t, err, "tak ne XDDDDDDD")
```
tremes, line 226: Please add some meaningful message here.

test/integration/main_test.go
```diff
@@ -191,44 +191,141 @@ func duration(t *testing.T, start time.Time, end time.Time) float64 {
...
+			"interval":   []byte("1m"), // for faster testing
+			"httpsProxy": []byte(proxy.address),
+			"httpProxy":  []byte(proxy.address),
+		},
+		Type: "Opaque",
+	}
+
+	clientset.CoreV1().Secrets(OpenShiftConfig).Delete(context.Background(), Support, metav1.DeleteOptions{})
+	_, err = clientset.CoreV1().Secrets(OpenShiftConfig).Create(context.Background(), &modifiedSecret, metav1.CreateOptions{})
+	e(t, err, "xd")
```
tremes, line 252: Same here.

test/integration/proxy.go
```diff
@@ -0,0 +1,89 @@
...
+	PORT       string
+	address    string
+	status     string
+	ready      bool
+	LogChecker *LogCheck
+	t          *testing.T
+}
+
+func (proxy *TinyProxy) create(t *testing.T, clientset *kubernetes.Clientset) error {
+	proxy.t = t
```
tremes, line 34: Do we need to do this? Can't we just pass the ref to `waitUntilReady`? Not sure what is better.....

test/integration/main_test.go
psimovec: Oh I see that I have totally missed some things, thanks for pointing out

test/integration/proxy.go
psimovec: Yes, you're right, thanks for pointing it out. The plan was to use goroutines to save some time, but this can be done even like you say. But it didn't really help in this case, it saved just a few seconds..

test/integration/main_test.go
psimovec: thanks

tremes:
@psimovec I think we can close it right?

tremes:
@psimovec I think we can close it right?

tremes:
Integration test have been moved to the internal repo.

PR #192, opened 2020-09-24
Bug 1883425: Gather installplans

martinkunc:
Installplans are part of operators-framework so we had to use dynamic client.
We are downloading TOP 100 installplans grouped by NS name and and CSV name.


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -772,6 +779,146 @@ func GatherMachineSet(i *Gatherer) func() ([]record.Record, []error) {
...
+					csv, _ = clusterServiceVersionNames[0].(string)
+				}
+
+				key := fmt.Sprintf("%s.%s.%s", ns, genName, csv)
+				m, ok := recs[key]
+				if !ok {
+					recs[key] = collectedPlan{Namespace: ns, Name: genName, CSV: csv, Count: 1}
+				} else {
+					m.Count++
+				}
```
psimovec, line 856: interesting PR overall :)

optional:
My personal preference would be to extract processing one item into a function with a descriptive name for better readability.
(if it makes sense for you in this case)

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -772,6 +779,148 @@ func GatherMachineSet(i *Gatherer) func() ([]record.Record, []error) {
...
+		return []error{fmt.Errorf("cannot cast item to map %v", item)}
+	}
+
+	err := failEarly(
+		func() error {
+			return parseJSONQuery(itemMap, "spec.clusterServiceVersionNames", &clusterServiceVersionNames)
+		},
+		func() error { return parseJSONQuery(itemMap, "metadata.namespace", &ns) },
+		func() error { return parseJSONQuery(itemMap, "metadata.generateName", &genName) },
+	)
```
tremes, line 851: Interesting technique :-) :+1:

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -772,6 +779,148 @@ func GatherMachineSet(i *Gatherer) func() ([]record.Record, []error) {
...
+			var items []interface{}
+			err = failEarly(
+				func() error { return parseJSONQuery(jsonMap, "metadata.continue?", &cont) },
+				func() error { return parseJSONQuery(jsonMap, "items", &items) },
+			)
+			if err != nil {
+				return nil, []error{err}
+			}
+
+			tc += len(items)
```
tremes, line 818: Is the `tc` used somewhere?

martinkunc: Aha, good point, I will remove this

pkg/gather/clusterconfig/clusterconfig_test.go
```diff
@@ -629,6 +629,47 @@ metadata:
...
+	if len(errs) > 0 {
+		t.Errorf("unexpected errors: %#v", errs)
+		return
+	}
+	if len(records) != 1 {
+		t.Fatalf("unexpected number or records %d", len(records))
+	}
+	b, _ := records[0].Item.Marshal(context.Background())
+	if string(b) != `{"items":[{"count":1,"csv":"lib-bucket-provisioner.v2.0.0","name":"install-","ns":"openshift-operators"}],"stats":{"TOTAL_COUNT":1,"TOTAL_NONUNIQ_COUNT":1}}` {
+		t.Fatalf("unexpected machineset name %s", records[0].Name)
```
tremes, line 669: Here should be `unexpected installplan name`, but let's fix it in another PR

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -809,6 +815,146 @@ func GatherMachineSet(i *Gatherer) func() ([]record.Record, []error) {
...
+//
+// The Kubernetes api https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/certificates/v1beta1/certificatesigningrequest.go#L78
+// Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#certificatesigningrequestlist-v1beta1certificates
+//
+// Location in archive: config/certificatesigningrequests/
+func GatherInstallPlans(i *Gatherer) func() ([]record.Record, []error) {
+	return func() ([]record.Record, []error) {
+		var plansBatchLimit int64 = 500
+		cont := ""
+		recs := map[string]collectedPlan{}
```
tremes, line 831: This should be probably `recs := map[string]*collectedPlan{}`. I tried to update the test by adding one same installplan (only with different name) and the corresponding count was only 1 in the gathered map (should be 2 in this case AFAIK).

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -809,6 +815,147 @@ func GatherMachineSet(i *Gatherer) func() ([]record.Record, []error) {
...
+// The Kubernetes api https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/certificates/v1beta1/certificatesigningrequest.go#L78
+// Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#certificatesigningrequestlist-v1beta1certificates
+//
+// Location in archive: config/certificatesigningrequests/
+func GatherInstallPlans(i *Gatherer) func() ([]record.Record, []error) {
+	return func() ([]record.Record, []error) {
+		var plansBatchLimit int64 = 500
+		cont := ""
+		recs := map[string]*collectedPlan{}
+		// oc get installplans -n=openshift-operators -v=6
```
tremes, line 832: Please remove.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -816,6 +822,146 @@ func GatherMachineSet(i *Gatherer) func() ([]record.Record, []error) {
...
+//
+// Location in archive: config/certificatesigningrequests/
+func GatherInstallPlans(i *Gatherer) func() ([]record.Record, []error) {
+	return func() ([]record.Record, []error) {
+		var plansBatchLimit int64 = 500
+		cont := ""
+		recs := map[string]*collectedPlan{}
+		total := 0
+		opResource := schema.GroupVersionResource{Group: "operators.coreos.com", Version: "v1alpha1", Resource: "installplans"}
+		resInterface := i.dynamicClient.Resource(opResource).Namespace("openshift-operators")
```
tremes, line 841: I just found that there are more installplans than just in `openshift-operators` namespace. I am wondering if this limitation makes sense. I did `oc get installplan -A` and there are more installplans in `openshift-storage`, `openshift-logging`, `default`, etc. (in shared4.5 cluster). WDYT?

martinkunc: It is a good idea, I will try to get all.

config/local.yaml
```diff
@@ -2,7 +2,7 @@ kind: GenericOperatorConfig
 apiVersion: operator.openshift.io/v1alpha1
 leaderElection:
   disable: true
-interval: "60s"
+interval: "5m"
 storagePath: /tmp/insights-operator
 endpoint: http://[::1]:8081
-#impersonate: system:serviceaccount:openshift-insights:gather
\ No newline at end of file
+impersonate: system:serviceaccount:openshift-insights:gather
```
martinkunc, line 8: This will make sure we will run under same account locally as we run in production

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -403,7 +409,7 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
 						logName = fmt.Sprintf("%s_previous.log", c.Name)
 					}
 					buf.Reset()
-					klog.V(2).Infof("Fetching logs for %s container %s pod in namespace %s (previous: %v): %q", c.Name, pod.Name, pod.Namespace, isPrevious, err)
+					klog.V(2).Infof("Fetching logs for %s container %s pod in namespace %s (previous: %v): %v", c.Name, pod.Name, pod.Namespace, isPrevious, err)
```
martinkunc, line 412: This it to display nil or error well, not like %q<nil>

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -904,6 +1065,25 @@ func GatherServiceAccounts(i *Gatherer) func() ([]record.Record, []error) {
 	}
 }
 
+func getAllNamespaces(i *Gatherer) (*corev1.NamespaceList, error) {
+	ns, ok := i.ctx.Value(contextKeyAllNamespaces).(*corev1.NamespaceList)
```
martinkunc, line 1069: Namespaces are used in two Gatherers. This will try to cache them into context (one run of Gather) to save some time.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -816,6 +822,161 @@ func GatherMachineSet(i *Gatherer) func() ([]record.Record, []error) {
...
+		if errors.IsNotFound(err) {
+			return nil, nil
+		}
+		if err != nil {
+			return nil, []error{err}
+		}
+
+		// collect from all openshift* namespaces
+		for _, ns := range config.Items {
+			if !strings.HasPrefix(ns.Name, "openshift") {
```
tremes, line 851: Do we want to filter installplans like that? I think user can install an OLM operator (which introduces new installplan AFAIK) to her/his own namespace.

martinkunc: This is mainly to collect statistics of openshift installplans, to limit them, not all of them. It is a bit wider then original idea collecting installplans from one openshift namespace.

tremes: OK. Thanks

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -64,13 +66,15 @@ const (
 	// Formerly, the `500 * 1024 / 450` expression was used instead.
 	metricsAlertsLinesLimit = 1000
 
+	// InstallPlansTopX is the Maximal number of Install plans by non-unique instances count
+	InstallPlansTopX = 100
```
tremes, line 70: I guess there's something worng with this limit. I tried to set it to 5 and got `{"items":[],"stats":{"TOTAL_COUNT":9,"TOTAL_NONUNIQ_COUNT":9}}` which means there was 9 installplans but can't see the five in the items array.

tremes:
I experimented with this locally several times and it seems to work good. Thanks
/lgtm

PR #193, opened 2020-09-24
Bug 1882699: Make gen-doc work outside of GOPATH

natiiix:


cmd/gendoc/main.go
```diff
@@ -169,32 +170,89 @@ func walkDir(cleanRoot string, md map[string]*DocBlock) error {
...
+	if err != nil {
+		return "", "", err
+	}
+
+	dirPath := absPkgFilePath
+	for {
+		goModPath = filepath.Join(dirPath, "go.mod")
+		if _, err := os.Stat(goModPath); os.IsNotExist(err) {
+			// This directory does not contain a go.mod file. Go to the parent directory.
+			dirPath = filepath.Dir(dirPath)
```
martinkunc, line 186: I think we could add a check for reaching root without finding go.mod. Without it, this would cycle indefinitively no ?

cmd/gendoc/main.go
```diff
@@ -169,32 +170,89 @@ func walkDir(cleanRoot string, md map[string]*DocBlock) error {
...
+		}
+		moduleName, err := getModuleNameFromGoMod(goModPath)
+		if err != nil {
+			log.Fatal(err)
+		}
+		importPath := filepath.Join(moduleName, relPkgPath)
+		return importPath
 	}
-	return gopath
+	return ""
```
martinkunc, line 255: This is maybe non-reachable no ?

natiiix: Well, theoretically yes, but Go doesn't seem to recognize the `Fatal` functions as a way of exiting a function, so it will complain that a return statement is missing. This is actually a common problem in pretty much every language when you're calling a function that terminates the program and don't return afterwards.

cmd/gendoc/main.go
natiiix: Good point. I'll have to check it to see what happens when it reaches root. I assumed it would just crash somewhere else at that point due to access violation or something.

martinkunc:
Thank you @natiiix good work
/lgtm

PR #196, opened 2020-09-30
Add pattern/patterns to TestArchiveContains

psimovec:
the task was split:
Something causes ci tests to fail.. extracted small part that causes fail into https://github.com/openshift/insights-operator/pull/234

Decided not to add default resources if it is not necessary.


test/integration/bugs_test.go
```diff
@@ -219,6 +226,32 @@ func TestArchiveContains(t *testing.T) {
...
+	t.Run("csr",
+		genLatestArchiveCheckPattern(
+			"csr", matchingFileExists,
+			`^config/certificatesigningrequests/.*\.json$`))
+
+	//https://bugzilla.redhat.com/show_bug.cgi?id=1879068
+	t.Run("HostsSubnet",
+		genLatestArchiveCheckPattern(
+			"hosts subnet", matchingFileExists,
+			`^config/hostsubnet/ip-.*\.json$`))
```
tremes, line 253: Wouldn't be easier to use this without the `ip` prefix there? I think there will be always some hostsubnet resource on a cluster.

tremes:
@psimovec please rebase

tremes:
Weird. I am getting `error: unable to recognize "./test/integration/resource_samples/csr.yaml": no matches for kind "CertificateSigningRequest" in version "certificates.k8s.io/v1"`

PR #198, opened 2020-10-01
Bug 1884221: IO becomes unhealthy due to a file change

rluders:
This patch should avoid that the operator ends up with an unhealthy status due to a possible file change.

### DIFF
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -14,9 +14,9 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
+	policyclient "k8s.io/client-go/kubernetes/typed/policy/v1beta1"
 	"k8s.io/client-go/pkg/version"
 	"k8s.io/client-go/rest"
-	policyclient "k8s.io/client-go/kubernetes/typed/policy/v1beta1"
 
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	networkv1client "github.com/openshift/client-go/network/clientset/versioned/typed/network/v1"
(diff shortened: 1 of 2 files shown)


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -169,7 +169,7 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
 // See: docs/insights-archive-sample/config/pdbs
 func GatherPodDisruptionBudgets(i *Gatherer) func() ([]record.Record, []error) {
 	return func() ([]record.Record, []error) {
-		pdbs, err := i.policyClient.PodDisruptionBudgets("").List(i.ctx, metav1.ListOptions{Limit: gatherPodDisruptionBudgetLimit,})
+		pdbs, err := i.policyClient.PodDisruptionBudgets("").List(i.ctx, metav1.ListOptions{Limit: gatherPodDisruptionBudgetLimit})
```
tremes, line 172: I have the impression that I've already seen these changes but the other way around. Is this the `go fmt` standard? Otherwise LGTM

rluders: It was formatted with `go fmt`. I guess that the recommendation for ending with `,` is only if you have multiple lines, then the last one should end with `,`.

martinkunc:
Nice, thank you Ricardo !

pkg/controller/operator.go
```diff
@@ -191,5 +191,5 @@ func (s *Support) Run(ctx context.Context, controller *controllercmd.ControllerC
 	go uploader.Run(ctx)
 
 	<-ctx.Done()
-	return fmt.Errorf("stopped")
+	return nil
```
martinkunc, line 194: Could I please ask you to add before this line log something like "stopped". I think it would be slight indication the logs in case we would investigate some crash or refresh (we are seeing that it is triggered, but this would help to see observer really finished).

rluders: Sure. I added it as a warning message to the log.

PR #200, opened 2020-10-05
Gather ServiceAccounts stats from cluster namespaces

martinkunc:
The change is collecting numbers of service accounts from kubernetes default namespaces and openshift built-in namespaces (openshift*).
It would be used to detect that a particular SA exists in ns, specifically to target https://access.redhat.com/solutions/5221881 and find lib-bucket-provisioner from NS openshift-storage

### DIFF
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -248,3 +248,17 @@ Response see https://docs.okd.io/latest/rest_api/policy_apis/poddisruptionbudget
 
 Location in archive: config/pdbs/
 See: docs/insights-archive-sample/config/pdbs
+
+
+## ServiceAccounts
+
+collects ServiceAccount stats
+from kubernetes default and namespaces starting with openshift.
+
+The Kubernetes api https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/serviceaccount.go#L83
+Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#serviceaccount-v1-core
+
+Location of serviceaccounts in archive: config/serviceaccounts
+See: docs/insights-archive-sample/config/serviceaccounts
+
+
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -16,8 +16,8 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	apixv1beta1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	apixv1beta1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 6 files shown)


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -1433,3 +1494,40 @@ func countLines(r io.Reader) (int, error) {
...
+// Marshal implements serialization of ServiceAccount
+func (a ServiceAccountsMarshaller) Marshal(_ context.Context) ([]byte, error) {
+	// Creates map for marshal
+	sr := map[string]interface{}{}
+	st := map[string]interface{}{}
+	st["TOTAL_COUNT"] = a.totalServiceAccounts
+	sr["serviceAccounts"] = st
+	nss := map[string]interface{}{}
+	st["namespaces"] = nss
+	//it := []interface{}{}
```
rluders, line 1513: Can't this comment be removed?

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -1433,3 +1494,40 @@ func countLines(r io.Reader) (int, error) {
...
+	for _, v := range a.v {
+		//kvp := map[string]interface{}{}
```
rluders, line 1515: Same here.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -1433,3 +1494,40 @@ func countLines(r io.Reader) (int, error) {
 		}
 	}
 }
+
+// ServiceAccountsMarshaller implements serialization of Service Accounts
+type ServiceAccountsMarshaller struct {
+	v                    []corev1.ServiceAccount
```
rluders, line 1500: Can't this var have a better name? It is kind hard to figure out that `v` is a reference to `[]corev1.ServiceAccount`, IMHO.

docs/gathered-data.md
```diff
@@ -247,4 +247,18 @@ The Kubernetes api https://github.com/kubernetes/client-go/blob/v11.0.0/kubernet
...
 Location in archive: config/pdbs/
-See: docs/insights-archive-sample/config/pdbs
\ No newline at end of file
+See: docs/insights-archive-sample/config/pdbs
+
+
+## ServiceAccounts
+
+collects ServiceAccount stats
+from kubernetes default and namespaces starting with openshft.
```
tremes, line 256: typo `openshift`

rluders (commented):
Didn't run the code yet, but I'll do it later, but looks good.

pkg/gather/clusterconfig/clusterconfig.go
martinkunc: Actually if you look it is a pattern in this file, and furthermore golang prefers short names, see in here: https://github.com/golang/go/wiki/CodeReviewComments#variable-names

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -843,6 +850,60 @@ func (i *Gatherer) gatherNamespaceEvents(namespace string) ([]record.Record, []e
...
+		config, err := i.coreClient.Namespaces().List(i.ctx, metav1.ListOptions{Limit: maxServiceAccountNamespacesLimit})
+		if errors.IsNotFound(err) {
+			return nil, nil
+		}
+		if err != nil {
+			return nil, []error{err}
+		}
+		totalServiceAccounts := 0
+		serviceAccounts := make([]corev1.ServiceAccount, 0)
+		records := make([]record.Record, 0)
```
0sewa0, line 872: Maybe super dumb question:
Here you are creating an empty `record.Record`, wouldn't it look simpler if you just simply said `var records []record.Record` which does the same thing ?

pkg/gather/clusterconfig/clusterconfig.go
0sewa0: does `v` stand for `value` ?

rluders: I mean, fine with the shorter name, but `v` is a kinda weird name to `ServiceAccount` shorthand. But, could be that I'm just bikeshedding. :thinking:

pkg/gather/clusterconfig/clusterconfig.go
rluders: @0sewa0 check this: https://stackoverflow.com/questions/25543520/declare-slice-or-make-slice

martinkunc: You are right, the make with 0 doesn't make much sense, it equal to records := []record.Record{}

tremes:
It looks good to me in general. Maybe we should increase `interval: "30s"` in config/local.yaml, because I was getting
`Unable to read ServiceAccounts in namespace openshift-service-ca-operator error context deadline exceeded` and 
```
E1005 12:06:48.285236  145144 periodic.go:75] Unable to flush recorder: cancelled before all results could be written to disk
E1005 12:06:48.285265  145144 periodic.go:157] config failed after 15.002s with: context deadline exceeded
```

pkg/gather/clusterconfig/clusterconfig.go
martinkunc: I think so Marcell, its probably from value, but I can chance it to for example sa. But as I said it is used this way above.

PR #201, opened 2020-10-05
[release-4.4] Bug 1885177:  Adds PodDisruptionBudgets gather.

0sewa0:


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -126,6 +133,30 @@ func (i *Gatherer) Gather(ctx context.Context, recorder record.Interface) error
...
+func GatherPodDisruptionBudgets(i *Gatherer) func() ([]record.Record, []error) {
+	return func() ([]record.Record, []error) {
+		pdbs, err := i.policyClient.PodDisruptionBudgets("").List(metav1.ListOptions{Limit: gatherPodDisruptionBudgetLimit})
+		if err != nil {
+			return nil, []error{err}
+		}
+		records := []record.Record{}
+		for _, pdb := range pdbs.Items {
+			records = append(records, record.Record{
+				Name: fmt.Sprintf("config/pdbs/%s", pdb.GetName()),
```
tremes, line 152: I think we should include the corresponding namespace name same as we do in master or is that another backport? :)

0sewa0: I created this for 4.5 https://github.com/openshift/insights-operator/pull/219

tremes: Yep thanks. I think it should be OK how you did that. :)

0sewa0:
@tremes  yes the examples are missing here and in 4.5, and I also checked in 4.6 that it also doesn't have the namespacing so I will update that to :/

0sewa0:
@tremes I undid the namespacing part because it would break the backporting workflow, so I will backport the namespacing stuff from 4.7 back to 4.6, 4.5, 4.4.

Kinda similar thing in the example part, so the example is missing in the 4.5 backport which is already merged, and the example wasn't updated with the original namespacing PR in 4.7, so I will fix this when I backport the namespacing stuff. (and have a separate PR for updating the example in master)

tremes:
@0sewa0 OK np. Thanks.
/lgtm

PR #187, opened 2020-09-22
Bug 1881364: Add new team members to OWNERS

natiiix:
Several members of the Insights Operator team are missing from the `OWNERS` file in the Insights Operator repository. This PR adds the missing members to the `OWNERS` file. Hopefully no one has been forgotten.

PR #203, opened 2020-10-05
Bug 1885261: Add new team members to OWNERS

natiiix:
Backport of #194, #191 and #187

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -6,6 +6,11 @@ approvers:
 - tisnik
 - alexandrevicenzi
 - natiiix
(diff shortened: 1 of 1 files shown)


tisnik (approved):
LGTM

bparees:
@natiiix i appreciate you following the process here, but next time you've got a change to non-shipping code like this that is definitely not a bug, just ask a pillar lead to apply the valid-bz label manually.  This BZ is going to go out as part of the list of "fixes" seen by customers which is a little weird.

PR #204, opened 2020-10-06
[release 4.5] Bug 1885515: Fixes 'make build'

0sewa0:
The paths in the `-ldflags` part of build script were incorrect causing the client-go `version.gitCommit` and `version.gitVersion` remained on the default value.


tisnik (approved):
Seems sane

tremes:
`build flag -mod=vendor only valid when using modules` from CI....

0sewa0:
@tremes  thanks, it looks like the `-mod=vendor` doesn't work with older go version (1.12) so I removed it :/

martinkunc:
@0sewa0 It should be fine to use -mod=vendor in go 1.12. The error message was complaining that we didn't have modules enabled. I tried this combination in go 1.12 and it works:
```
GO111MODULE=on go build -mod=vendor -ldflags "-X github.com/openshift/insights-operator/vendor/k8s.io/client-go/pkg/version.gitCommit=$(git rev-parse HEAD) -X github.com/openshift/insights-operator/vendor/k8s.io/client-go/pkg/version.gitVersion=v1.0.0+$(git rev-parse --short=7 HEAD)" -o bin/insights-operator ./cmd/insights-operator
```

Please use it, in 4.5/4.4 it should help with the removed dependency issue. (4.6 has already upgraded dependencies).

tremes:
@martinkunc Ahh OK. Thanks

tremes:
Works for me.
/lgtm

PR #205, opened 2020-10-06
[release 4.4] Bug 1885516: Fixes 'make build'

0sewa0:
The paths in the `-ldflags` part of build script were incorrect causing the client-go `version.gitCommit` and `version.gitVersion` remained on the default value.


russellb:
(patch manager) Is this really needed in 4.4?  Typically only urgent/high priority BZs go into 4.4 now.

0sewa0:
We discussed it with the team, and as 4.4 shifts out of focus this change is not needed. So closing this PR.

PR #208, opened 2020-10-08
WIP Gather cluster operator configs

tremes:
There is a request to collect info about `managementState` field. 

I was searching for some operator related resource with this field and found this - e.g https://docs.openshift.com/container-platform/4.5/rest_api/operator_apis/config-operator-openshift-io-v1.html#config-operator-openshift-io-v1. So this can be probably called operator configs (correct me if I am wrong) and there are more of them each one having its own dedicated interface. 

This fact led me to this idea of putting it somewhere else then `clusterconfig.go` so that this file wouldn't grow so much.


tremes:
@iNecas, @martinkunc can you please check if this makes sense or not?

martinkunc:
I think I would rather prefer keep using gatherconfig.go then defining separate gatherer until we think out some good refactoring rules.
And I think gathering only two specific operators wouldnt not be enough. I think we need to gather also OLM operators and all other openshift operators. We should investigate how to do it dynamically if possible, like how `oc adm -A ` does it as described in KB.

tremes:
@martinkunc Thanks for review. Yes these two are just a proposal, because there is more of them. 

Wrt OLM operators - my understanding is that the problem is that there's a difference between `OperatorSpec` (defined here https://github.com/openshift/api/blob/master/operator/v1/types.go using the desired `managementState`) and the `Operator` type defined by OLM (https://github.com/operator-framework/api/blob/master/pkg/operators/v1/operator_types.go) which doesn't seem to have anything like that.

tremes:
Closing. I'll create a new PR.

PR #209, opened 2020-10-08
Bug 1886462: Add the namespace to the gatherers reports to avoid conflicts

rluders:
Fix collisions between pdbs with the same name but living in different namespaces.

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -175,8 +175,12 @@ func GatherPodDisruptionBudgets(i *Gatherer) func() ([]record.Record, []error) {
 		}
 		records := []record.Record{}
 		for _, pdb := range pdbs.Items {
+			recordName := fmt.Sprintf("config/pdbs/%s", pdb.GetName())
+			if pdb.GetNamespace() != "" {
+				recordName = fmt.Sprintf("config/pdbs/%s/%s", pdb.GetNamespace(), pdb.GetName())
+			}
 			records = append(records, record.Record{
-				Name: fmt.Sprintf("config/pdbs/%s/%s", pdb.GetNamespace(), pdb.GetName()),
+				Name: recordName,
 				Item: PodDisruptionBudgetsAnonymizer{&pdb},
 			})
 		}
@@ -333,12 +337,8 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
 		}
 		records := make([]record.Record, 0, len(config.Items))
 		for index := range config.Items {
-			recordName := fmt.Sprintf("config/clusteroperator/%s", config.Items[index].Name)
-			if config.Items[index].Namespace != "" {
-				recordName = fmt.Sprintf("config/clusteroperator/%s/%s", config.Items[index].Namespace, config.Items[index].Name)
-			}
 			records = append(records, record.Record{
-				Name: recordName,
+				Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[index].Name),
 				Item: ClusterOperatorAnonymizer{&config.Items[index]},
 			})
 		}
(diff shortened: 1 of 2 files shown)


tisnik (approved):
It looks ok!

rluders:
I went one step further in this one, 'cause I noticed that the same issue could be happening with other gatherers. So, I guess that know, it assure that whenever we have a namespace, the report should be stored under the namespace's folder.

It would need some changes in the tests, and also, probably we want to want to release notes for our users. 'Cause this change is a potential breaker.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -471,10 +478,16 @@ func GatherConfigMaps(i *Gatherer) func() ([]record.Record, []error) {
 		records := make([]record.Record, 0, len(cms.Items))
 		for i := range cms.Items {
 			for dk, dv := range cms.Items[i].Data {
-				records = append(records, record.Record{Name: fmt.Sprintf("config/configmaps/%s/%s", cms.Items[i].Name, dk), Item: ConfigMapAnonymizer{v: []byte(dv), encodeBase64: false}})
+				records = append(records, record.Record{
```
martinkunc, line 481: We are fetching configmaps from ns openshift-config, so I think it doesn't make sense to add the NS prefix

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -578,7 +591,10 @@ func GatherHostSubnet(i *Gatherer) func() ([]record.Record, []error) {
 		}
 		records := make([]record.Record, 0, len(hostSubnetList.Items))
 		for _, h := range hostSubnetList.Items {
-			records = append(records, record.Record{Name: fmt.Sprintf("config/hostsubnet/%s", h.Host), Item: HostSubnetAnonymizer{&h}})
+			records = append(records, record.Record{
```
martinkunc, line 594: hostsubnets doesn't have namespace, because they are in Cluster scope.
You can see the scope using:
```
oc get crd hostsubnets.network.openshift.io -o=yaml | yq .spec.scope
```

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -740,7 +756,10 @@ func GatherCertificateSigningRequests(i *Gatherer) func() ([]record.Record, []er
 		}
 		records := make([]record.Record, len(csrs))
 		for i, sr := range csrs {
-			records[i] = record.Record{Name: fmt.Sprintf("config/certificatesigningrequests/%s", sr.ObjectMeta.Name), Item: sr}
+			records[i] = record.Record{
```
martinkunc, line 759: CSRs are also Cluster scoped and have no Namespace.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -773,8 +792,12 @@ func GatherCRD(i *Gatherer) func() ([]record.Record, []error) {
 			if err != nil {
 				return []record.Record{}, []error{err}
 			}
+			recordName := fmt.Sprintf("config/crd/%s", crd.Name)
```
martinkunc, line 795: CRDs themselves are also Cluster scoped. The hint is also that Api doesn't offer an option to fetch it from namespace.

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -333,7 +333,14 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
 		}
 		records := make([]record.Record, 0, len(config.Items))
 		for index := range config.Items {
-			records = append(records, record.Record{Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[index].Name), Item: ClusterOperatorAnonymizer{&config.Items[index]}})
+			recordName := fmt.Sprintf("config/clusteroperator/%s", config.Items[index].Name)
```
martinkunc, line 336: clusteroperators are also not namespace specific, but they are related to namespaces, where the pods or events are NS specific.

martinkunc:
Thanks @rluders, nice work. It looks only PDBs and MachineSets are NS scoped. Maybe we could add investigation of this add to our "adding new data" document..

0sewa0:
Tested it locally, works as expected :)

/lgtm

PR #211, opened 2020-10-13
Take default support instead of rely on existence of config

martinkunc:
While loading config, serialized.ToController was supposing that all values will be explicitly set in serialized config (config/local.yaml file), instead of using default values from support.Controller.
This change will use default values, unless they are overwritten by serializable config, or later overwritten by from secrets.
Before this change, loading logic required the PullReport.Delay to be set, which leads to crash on serialized.ToController error with `delay must be a non-negative duration`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,14 +20,15 @@ type Serialized struct {
 	Impersonate string `json:"impersonate"`
 }
 
-func (s *Serialized) ToController() (*Controller, error) {
-	cfg := Controller{
-		Report:         s.Report,
-		StoragePath:    s.StoragePath,
-		Endpoint:       s.Endpoint,
-		ReportEndpoint: s.PullReport.Endpoint,
-		Impersonate:    s.Impersonate,
+func (s *Serialized) ToController(cfg *Controller) (*Controller, error) {
+	if cfg == nil {
+		cfg = &Controller{}
 	}
+	cfg.Report = s.Report
+	cfg.StoragePath = s.StoragePath
+	cfg.Endpoint = s.Endpoint
+	cfg.Impersonate = s.Impersonate
+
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -20,14 +20,15 @@ type Serialized struct {
...
 }
 
-func (s *Serialized) ToController() (*Controller, error) {
-	cfg := Controller{
-		Report:         s.Report,
-		StoragePath:    s.StoragePath,
-		Endpoint:       s.Endpoint,
-		ReportEndpoint: s.PullReport.Endpoint,
-		Impersonate:    s.Impersonate,
+func (s *Serialized) ToController(cfg *Controller) (*Controller, error) {
```
tremes, line 23: Do we need to send the `cfg` attribute? We can always create a new `Controller` as before or not? Otherwise it looks good.

rluders:
It seems a really straightforward solution, so just looking at the code it seems OK.

/lgtm

rluders (approved):

pkg/config/config.go
martinkunc: This is called by Load - so the steps are:
Set default to support.Controller (in code)
Try to load config to support.Controller

And unfortunatelly previously loading was resetting whole support.Controller with what was set in config file, so all the defaults had to be sed there. That was ignoring what was already set in support.Controller and creating a new one.

Now, there is an option to not to pass existing Controller and create failback to previous behaviour, but we use the way which passes in the existing support.Controller default values set before.
And there is an option to override the defaults from config. For this there are if len(...)>0 blocks below - if it is set in config file, use it.

tremes: Ahh I see. Sorry. I should have read the description.

PR #212, opened 2020-10-13
Fix error in default Smart proxy report endpoint

joselsegura:


rluders (approved):
/lgtm

rluders:
BTW, I'm surprised that these values are fixed in code and not coming from some configmap or somethig. :thinking:

joselsegura:
> BTW, I'm surprised that these values are fixed in code and not coming from some configmap or somethig. thinking

It can be read from the `configobserver` package too from the cluster secrets, or by the configuration parser in the configuration file... but the default one if all of them are missing is incorrect, that leads in an IO that cannot retrieve the report

rluders:
> all of them are missing is incorrect, that leads in an IO that cannot retrieve the

I see. That makes sense.

PR #216, opened 2020-10-13
Bug 1887778: Adds ContainerRuntimeConfig gatherer

0sewa0:
My cluster didn't have one by default so here is an example you can use to verify:
```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: ContainerRuntimeConfig
metadata:
 name: set-log-and-pid
spec:
 machineConfigPoolSelector:
   matchLabels:
     debug-crio: config-log-and-pid
 containerRuntimeConfig:
   pidsLimit: 2048
   logLevel: debug
```

### DIFF
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -272,3 +272,13 @@ Location of serviceaccounts in archive: config/serviceaccounts
 See: docs/insights-archive-sample/config/serviceaccounts
 
 
+## ContainerRuntimeConfig
+
+collects ContainerRuntimeConfig information
+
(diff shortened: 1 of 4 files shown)


tremes:
Verified on my cluster. @0sewa0 can you please check if it works on 4.3 (and or 4.4)? I guess we should explicitly add the new apiGroups to `manifests/03-clusterrole.yaml`.

0sewa0:
@tremes Tried it on 4.4 and it worked without issue.

tremes:
@0sewa0 OK thanks. Interesting....I tried it on 4.3 and it works as well.

PR #218, opened 2020-10-13
Gather StatefulSet configs from default & openshift namespaces

tremes:


0sewa0:
@tremes does this need an update to `manifests/03-clusterrole.yaml` ?

tremes:
@0sewa0 I hope it doesn't. I just tried it on 4.3 and was able to collect them.

PR #220, opened 2020-10-14
Bug 1888172: Updates the sample archive.

0sewa0:
https://github.com/openshift/insights-operator/pull/209 didn't update the sample archive, but machineset and pdbs are now namespaced


tremes:
OK. We should probably start using json formatting in the sample archive. I know it's very likely mixed now, but still....

martinkunc:
Actually, Tomas had a good point and it is not that hard.
@0sewa0 Could you please look into the [formatting archive json files](https://github.com/openshift/insights-operator#formatting-archive-json-files) If you execute the command in your docs folder, it should reformat the files which are not.

PR #223, opened 2020-10-15
Gather clusteroperator resources

tremes:
Gather info about clusteroperator resources. I added only basic test for `CluterOperator` (no related resource there) because it doesn't seem to be easy to use `discoveryClient.ServerPreferredResources()`. This cal requires to have some `APIGroups` available (see https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go#L287)), which is not our case.

I would suggest to test this properly in integration tests.

### DIFF
--- a/pkg/gather/clusterconfig/clusterconfig.go
+++ b/pkg/gather/clusterconfig/clusterconfig.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/json"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	kubescheme "k8s.io/client-go/kubernetes/scheme"
 	appsclient "k8s.io/client-go/kubernetes/typed/apps/v1"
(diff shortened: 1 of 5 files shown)


pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -427,6 +436,37 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
 	}
 }
 
+func collectClusterOperatorResources(i *Gatherer, co configv1.ClusterOperator) *clusterOperatorResource {
+	var relObj configv1.ObjectReference
```
martinkunc, line 440: What using *configv1.ObjectReference.
Later you could check if relObj == nil instead of creating empty struct

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -427,6 +436,37 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
...
+	for _, ro := range co.Status.RelatedObjects {
+		if ro.Group == "operator.openshift.io" {
+			relObj = ro
+		}
+	}
+	if relObj == (configv1.ObjectReference{}) {
+		return nil
+	}
+	gvr := schema.GroupVersionResource{Group: "operator.openshift.io", Version: "v1", Resource: strings.ToLower(relObj.Resource)}
+	clusterResource, err := i.dynamicClient.Resource(gvr).Get(i.ctx, "cluster", metav1.GetOptions{})
```
martinkunc, line 450: Instead of "cluster" should be name of related resource I think (even thought they are all cluster in our case), and I would like a bit more if Group would be also from relObj (even though its semantically the same).

pkg/gather/clusterconfig/clusterconfig.go
```diff
@@ -434,6 +448,70 @@ func GatherClusterOperators(i *Gatherer) func() ([]record.Record, []error) {
...
+
+func getOperatorResourcesVersions(i *Gatherer) (map[string][]string, error) {
+	resources, err := i.discoveryClient.ServerPreferredResources()
+	if err != nil {
+		return nil, err
+	}
+	resourceVersionMap := make(map[string][]string)
+	for _, v := range resources {
+		if strings.Contains(v.GroupVersion, "operator.openshift.io") {
+			gvA := strings.Split(v.GroupVersion, "/")
```
martinkunc, line 498: What about to use `ParseGroupVersion` ?

martinkunc:
Nice work Tomas

pkg/gather/clusterconfig/clusterconfig.go
tremes: How can I use that? Where is it available please?

martinkunc: It is in `schema.ParseGroupVersion` or here https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/schema/group_version.go#L211

tremes: Thanks

rluders (approved):
/lgtm

PR #225, opened 2020-10-15
Bug 1888601: Fixes policyClient and the corresponding config.

0sewa0:
After this change, when you locally change your config/local.yaml, to some invalid account
impersonate: system:serviceaccount:openshift-insights:some-invalid-account
All the gatherers should return error and no archive should be created in /tmp in the end (withtout this, only pdb was gathered, because it used operator account)


martinkunc:
Thank you Marcell!

PR #226, opened 2020-10-15
[release 4.6] Bug 1888602: Fixes policyClient and the corresponding config.

0sewa0:
After this change, when you locally change your config/local.yaml, to some invalid account
impersonate: system:serviceaccount:openshift-insights:some-invalid-account
All the gatherers should return error and no archive should be created in /tmp in the end (withtout this, only pdb was gathered, because it used operator account)

### DIFF
--- a/manifests/03-clusterrole.yaml
+++ b/manifests/03-clusterrole.yaml
@@ -62,14 +62,6 @@ rules:
   - namespaces
   verbs:
   - get
-- apiGroups:
-  - policy
-  resources:
-  - poddisruptionbudgets
-  verbs:
-  - list
-  - get
-  - watch
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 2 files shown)


psimovec:
failure there is not supposed to be a failure

martinkunc:
@0sewa0 Marcell, could you please also add the machinesets persmissions here, is it is missing in all versions I believe.

rluders:
Just reading the code seems fine.

/lgtm

sdodson:
(Patch Manager) This PR was reviewed for inclusion last week, however this adds feature work during the weeks immediately after release where we're focused on stability, upgrade success, and anything which blocks backports of critical customer impacting bugs. Since this doesn't meet those criteria it was not accepted last week and likely will not until we've been able to promote the 4.5 to 4.6 upgrade to the stable channel.

sdodson:
Looking at this again this appears to fix a bug in existing code, so accepting it versus the others which add net new collection.

tremes:
@sdodson Thanks

PR #234, opened 2020-10-22
Simplify/generalize host subnet pattern

psimovec:


quarckster:
I think we should find out why `HostsSubnet` is not created. We shouldn't add this resource by ourselves.

PR #241, opened 2020-10-28
Added TestArchiveUploadedAndResultReceived

quarckster:


psimovec (commented):
Thanks for the changes!
I was going to comment that logLineTime expects the line to be in the logs. But you did the necessary changes already.

It definitely makes sense to have such test.

Only thing that I'd point out is that the logged times can be incorrect as the logLineTime function captures just HH:MM:SS. _edit: explained in DMs that it happens in case when second logLineTime is in different day than first one._

Otherwise LGTM

PR #250, opened 2020-11-04
[release-4.4] Bug 1894415: Gather StatefulSet configs from default & …

tremes:
…openshift namespaces


tremes:
4.4 is approaching EOL.

PR #252, opened 2020-11-04
[release-4.5] Backport e2e tests

psimovec:
Commented out SnapshotsCRD as it's not backported yet


tremes:
I think this is the change in non-shipping code, so we should be able to backport it without having to create a bugzilla.

sdodson:
@psimovec If this is a backport which PR or commits does it backport?

psimovec:
@sdodson I wasn't doing any backporting of e2e-tests before, so this PR is backport of almost all e2e tests. This is the closest one, 4.6 backport https://github.com/openshift/insights-operator/pull/248
And relevant will be this PR with the dependencies, it was already in 4.6 without backporting https://github.com/openshift/insights-operator/pull/134

test/integration/main_test.go
```diff
@@ -70,42 +74,31 @@ func configV1Client() (result *configv1client.ConfigV1Client) {
...
 }
 
-func clusterOperatorInsights() *configv1.ClusterOperator {
-	// get info about insights cluster operator
-	operator, err := configClient.ClusterOperators().Get("insights", metav1.GetOptions{})
+func clusterOperator(clusterName string) *configv1.ClusterOperator {
+	// get info about given cluster operator
+	operator, err := configClient.ClusterOperators().Get(clusterName, metav1.GetOptions{})
 	if err != nil {
+		// TODO -> change to t.Fatal in follow-up PR
```
tremes, line 81: Do we really want a followup PR when backporting? :)

test/integration/main_test.go
```diff
@@ -236,6 +227,153 @@ func forceUpdateSecret(ns string, secretName string, secret *v1.Secret) error {
...
+	option := &corev1.PodExecOptions{
+		Command: cmd,
+		Stdin:   true,
+		Stdout:  true,
+		Stderr:  true,
+		TTY:     false,
+	}
+	if stdin == nil {
+		option.Stdin = false
+	}
```
tremes, line 248: Minor. Maybe it would be better to do it the other way around. Set `Stdin` to false by default and if there's some then set it to true.

test/integration/main_test.go
psimovec: I want to keep it as similar to the version above as possible.

test/integration/main_test.go
psimovec: Thanks for your comment, I should do that on master version, but I don't want to apply this change here, as it's not in 4.6 yet

tremes: OK then :)

test/integration/main_test.go
tremes: Yeah ok makes sense.

PR #254, opened 2020-11-05
Shortening the initial upload delay

tremes:
If you look at the wait.Jitter function https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L193 then you can see that the initial delay was (in case of 2h intervals) in [15m, 45m) interval.

This is to shorten the initial delay to [10m, 20m) in case of 2h intervals.


tremes:
I was thinking about this again and now I think it doesn't make much sense to do this. We probably need more robust logic to handle the first upload time (e.g waiting on CVO is available and ready)

PR #255, opened 2020-11-09
Diskrecorder simplify the Summary function

tremes:
This is just an idea how to simplify diskrecorder.go. I believe we don't want to merge existing tar files together and most importantly, there is nothing to merge IMO, because the length of merged slice (https://github.com/openshift/insights-operator/blob/master/pkg/record/diskrecorder/diskrecorder.go#L293) will always be 0 or 1 (due to https://github.com/openshift/insights-operator/blob/master/pkg/record/diskrecorder/diskrecorder.go#L301-L307 note the slice is initialized in the for loop) if I am not mistaken.


rluders:
Looks pretty nice, so far. It makes sense to me.

rluders (approved):
/lgtm

PR #257, opened 2020-11-10
Separating the gather logic into separate files

0sewa0:
BE AWARE THIS IS A FAT ONE.

So to save time and pain in the future there is a major refactor necessary for the `clusterconfig.go`. Its an unmaintainable mess of code thats 2000 lines long. This PR splits it up into individual files.

There are still some problematic parts like the `0-utils.go` which contains all the code that was shared between the gather functions and it should be refactored further. `versions.go` is also problematic because it has 2 gather functions in it, one of which relies on the other, which we don't want.

Also gather-clients are now created where they will be used and not passed down from three levels higher. Only the configs and context is provided to the gather-functions.


pkg/controller/operator.go
```diff
@@ -172,15 +118,15 @@ func (s *Support) Run(ctx context.Context, controller *controllercmd.ControllerC
 
 	// the gatherers periodically check the state of the cluster and report any
 	// config to the recorder
-	configPeriodic := clusterconfig.New(gatherConfigClient, gatherKubeClient.CoreV1(), gatherKubeClient.CertificatesV1beta1(), metricsClient, registryClient.ImageregistryV1(), crdClient, gatherNetworkClient, dynamicClient, gatherPolicyClient, appsClient, discoveryClient)
+	clusterConfigGatherer := clusterconfig.New(gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig)
 	periodic := periodic.New(configObserver, recorder, map[string]gather.Interface{
-		"config": configPeriodic,
+		"clusterconfig": clusterConfigGatherer,
```
0sewa0, line 123: This should be changed back so that we don't break the rules

pkg/gather/clusterconfig/1-examples_test.go
```diff
@@ -0,0 +1,35 @@
+package clusterconfig
+
+import (
+	"fmt"
+)
+
+func ExampleGatherMostRecentMetrics_Test() {
```
0sewa0, line 7: What is the purpose of this file ? Could it be removed ?

pkg/controller/operator.go
tremes: I guess I was wrong. The `config` string is really hardcoded in every gather function so probably no need to keep the original name.

pkg/gather/clusterconfig/1-examples_test.go
tremes: Let me try to rephrase Martin here. This is used by the doc generator. In general this generator takes `GatherXY` function and then executes `ExampleGatherXY` and saves the result to the doc as sample output.

rluders:
Tested locally. And... it works. :+1: 
/lgtm

PR #260, opened 2020-11-12
Fix TestProxy in clusterauthorizer_test.go

tremes:
I think we had this wrong. We discovered the problem in https://github.com/RedHatInsights/insights-ingress-http-client/commit/2a8f0c40beee7eed9a2fa03f67969b036ff0bfca


psimovec:
Makes sense to me, HTTPS is mentioned in one of the comments, and with this change, HttpConfig values match the EnvValues.
/lgtm

PR #261, opened 2020-11-13
[release-4.6] Bug 1894244: Fixes records index on diskrecorder

rluders:
This is a backport of #246 

[CCXDEV-3429](https://issues.redhat.com/browse/CCXDEV-3429)


markmc:
(patch manager) @rluders I tried hard, but I can't tell from the PR, commit message, bz, or JIRA just what the impact of not having this fix is, nor the risk that the backport will result in other issues

rluders:
@markmc basically it is a bug fix. It fixes a problem related to the report collection, where it wasn't considering the resource naming space and was overwriting some reports on the index of reports array. In this case, it wasn't only ignoring the fully report collection, as it wasn't cleaning up it from memory. This PR fixes both issues.

PR #264, opened 2020-11-16
Reuse archives & refactor archive checks + some fixes

psimovec:
Signed-off-by: Pavel Simovec <psimovec@redhat.com>

### DIFF
--- a/test/integration/bugs_test.go
+++ b/test/integration/bugs_test.go
@@ -3,11 +3,10 @@ package integration
 import (
 	"context"
 	"fmt"
-	"regexp"
+	"reflect"
 	"testing"
 	"time"
 
-	"k8s.io/api/certificates/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -23,6 +22,9 @@ const knownFileSuffixesInsideArchiveRegex string = `(` +
 
 //https://bugzilla.redhat.com/show_bug.cgi?id=1841057
 func TestUploadNotDelayedAfterStart(t *testing.T) {
+	/* TODO Result is irellevant as at this point IO was already restarted
+	   this test would most likely fail and it's known issue, better solution is needed, skipping now*/
+	t.Skip()
 	LogChecker(t).Timeout(30 * time.Second).Search(`It is safe to use fast upload`)
 	time1 := logLineTime(t, `Reporting status periodically to .* every`)
 	time2 := logLineTime(t, `Successfully reported id=`)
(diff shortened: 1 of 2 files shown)


test/integration/main_test.go
```diff
@@ -314,6 +317,9 @@ func degradeOperatorMonitoring(t *testing.T) func() {
 func changeReportTimeInterval(t *testing.T, newInterval []byte) []byte {
 	supportSecret, _ := clientset.CoreV1().Secrets(OpenShiftConfig).Get(context.Background(), Support, metav1.GetOptions{})
 	previousInterval := supportSecret.Data["interval"]
+	if previousInterval == nil {
+		t.Fatal("missing secret with name", Support)
+	}
```
tremes, line 322: Well `supportSecret` can be nil right so this should be updated IMO. We should also check the error and not ignore it IMO.

test/integration/bugs_test.go
```diff
@@ -160,148 +162,114 @@ func TestUnreachableHost(t *testing.T) {
...
+func genLatestArchiveCheckPattern(prettyName string, check archiveCheck, archive []string, patterns ...string) test {
 	return func(t *testing.T) {
-		err := latestArchiveCheckFiles(t, prettyName, check, pattern)
-		if err != nil {
-			t.Fatal(err)
+		if len(patterns) == 0 {
+			t.Fatal(prettyName, ": No patterns to check")
+		}
+		for _, pattern := range patterns {
+			err := checkArchiveFiles(t, prettyName, check, pattern, archive)
```
tremes, line 171: Just minor - would it make sense to move the `checkArchiveFiles` (and probably also the `latestArchiveFiles`) function to `bugs_test.go`? 
I am not very experienced with Go tests and testing, but the `main_test.go` looks more like some test utils file (if it is so then we can leave these functions there) .

psimovec: I understand it as test utils file now too, checkArchiveFiles could make sense to move to bugs.go, I don't think so about latestArchiveFiles (I could want to import it to different test file).. But it would make sense to move all the util functions from main.go to different files, as it starts getting unnecessarily big, an it's not very readable what main.go is

psimovec: (should be done in different PR)

test/integration/main_test.go
rluders: +1

test/integration/bugs_test.go
rluders: I agree with @psimovec on this one. For now, it is fine, but it is starting to become a huge and not readable file.

rluders (approved):
/lgtm

rluders (changes requested):
Just a few changes, but in general it seems solid.

/lgtm

rluders (approved):
/lgtm

PR #270, opened 2020-11-25
First check IO container status and optionally delay first gathering

tremes:
This is to replace initial upload delay (setting it to 0 actually) with gathering delay. What we are doing now is following:
1. Run gathering right after IO start
2. Check if IO is healthy (checked in https://github.com/openshift/insights-operator/blob/master/pkg/controller/status/status.go#L396)
   - if it's healthy then upload immediately
   - if it's not heatlhy then wait some time between (15min, 45min) - this happens quite often in CI clusters because of `E1104 14:09:52.697292       1 status.go:406] Couldn't get Insights Operator Pod to detect its status. Error: pods "insights-operator-6d5cf8b5ff-982dt" is forbidden: User "system:serviceaccount:openshift-insights:gather" cannot get resource "pods" in API group "" in the namespace "openshift-insights": RBAC: clusterrole.rbac.authorization.k8s.io "cluster-reader" not found` 
This doesn't make much sense IMHO. I mean to send something which is older than 20 minutes (and actually it may be from time when cluster was not up and ready. We can optionally add one more condition checking clusteroperators available=true, progressing= false)

This proposal is doing following:
1. Check we can read IO container status and the status is running. 
2. If step 1 is OK within 1 minute then run gathering and upload immediately
3. if step 1 is not OK then wait (`wait.Jitter(s.Controller.Interval/12, 1)`, which is (10min, 20min) for 2h interval) before first gathering (which in fact means delaying first upload as well AFAIK)

### DIFF
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -141,7 +141,7 @@ func (s *Support) Run(ctx context.Context, controller *controllercmd.ControllerC
 
 	// upload results to the provided client - if no client is configured reporting
 	// is permanently disabled, but if a client does exist the server may still disable reporting
-	uploader := insightsuploader.New(recorder, insightsClient, configObserver, statusReporter)
+	uploader := insightsuploader.New(recorder, insightsClient, configObserver, statusReporter, initialDelay)
 	statusReporter.AddSources(uploader)
 
 	// TODO: future ideas
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -5,11 +5,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"time"
 
 	"k8s.io/klog"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/pkg/version"
@@ -17,7 +20,8 @@ import (
 
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
-
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	"github.com/openshift/insights-operator/pkg/authorizer/clusterauthorizer"
 	"github.com/openshift/insights-operator/pkg/config"
@@ -56,7 +60,7 @@ func (s *Support) LoadConfig(obj map[string]interface{}) error {
 
 func (s *Support) Run(ctx context.Context, controller *controllercmd.ControllerContext) error {
 	klog.Infof("Starting insights-operator %s", version.Get().String())
-
+	initialDelay := 0 * time.Second
 	if err := s.LoadConfig(controller.ComponentConfig.Object); err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


tremes:
@smarterclayton WDYT pls ? Can you please check? Does it make sense or is it non-sense? :)

tremes:
Actually now I am thinking it would be better to check if it's in crash loop first (similarly as in https://github.com/openshift/insights-operator/blob/master/pkg/controller/status/status.go#L394-L408), if it is then delay gathering,  if it is not in crash loop then (optionally?) check clusteroperators and if they're OK then upload immediately.

tremes:
I reworked it to check the IO container status first and optionally delay first gathering and upload. We can optionally add one more condition checking clusteroperators available=true, progressing=false to eliminate some false positives, but it seems to me as an edge case.

0sewa0:
I like the proposal and it makes sense to me. Also it removes more stuff then it adds which makes me happy :)
Didn't have time to test it yet, will do as soon as I can.

pkg/controller/operator.go
```diff
@@ -123,7 +126,13 @@ func (s *Support) Run(ctx context.Context, controller *controllercmd.ControllerC
 		"clusterconfig": clusterConfigGatherer,
 	})
 	statusReporter.AddSources(periodic.Sources()...)
-	go periodic.Run(4, ctx.Done())
+
+	// check we can read IO container status and we are not in crash loop
+	err = wait.PollImmediate(5*time.Second, 1*time.Minute, isRunning(ctx, gatherKubeConfig))
+	if err != nil {
+		initialDelay = wait.Jitter(5*time.Minute, 1)
```
tremes, line 133: We should and some info log here. This will also delay every local run since we usually don't have the env vars (https://github.com/openshift/insights-operator/pull/270/files#diff-52732d3ec39ef3de250ba276b2f85c749aef6e8f9e34cd6066abaa5f165bdd1aR176) set.

tremes: So it is really better to use someting lik  `wait.Jitter(s.Controller.Interval/12, 1)`

pkg/controller/operator.go
```diff
@@ -157,3 +167,24 @@ func (s *Support) Run(ctx context.Context, controller *controllercmd.ControllerC
...
+
+func isRunning(ctx context.Context, config *rest.Config) wait.ConditionFunc {
+	return func() (bool, error) {
+		c, err := corev1client.NewForConfig(config)
+		if err != nil {
+			return false, err
+		}
+		pod, err := c.Pods(os.Getenv("POD_NAMESPACE")).Get(ctx, os.Getenv("POD_NAME"), metav1.GetOptions{})
+		if err != nil {
+			return false, err
```
tremes, line 179: Actually this is questionable and very likely wrong, because if we return err then the `wait.PollImmediate` won't try again.

tremes:
Hm I think there's still need to delay the upload (in case of some problems) as well, because now it can happen that it will upload some old existing archive file (which has been already uploaded).

pkg/controller/operator.go
```diff
@@ -123,14 +127,21 @@ func (s *Support) Run(ctx context.Context, controller *controllercmd.ControllerC
 		"clusterconfig": clusterConfigGatherer,
 	})
 	statusReporter.AddSources(periodic.Sources()...)
-	go periodic.Run(4, ctx.Done())
+
+	// check we can read IO container status and we are not in crash loop
+	err = wait.PollImmediate(20*time.Second, wait.Jitter(s.Controller.Interval/12, 0.1), isRunning(ctx, gatherKubeConfig))
```
tremes, line 132: This will delay first gathering when running locally, because we usually don't set `POD_NAMESPACE` and `POD_NAME` env vars. But it shouldn't be long given the fact that the `s.Controller.Interval` is locally 5min.

0sewa0:
/lgtm

Tested it locally using cluster-bot, worked as expected.

PR #271, opened 2020-11-26
Init health status metrics to distinguish no report state vs 0 problems

tremes:


tisnik (approved):
That's fast Tomas. TYVM!

PR #272, opened 2020-11-26
Fix clusteroperators serialization

tremes:


rluders (approved):
Very simple. Just looking at the code, seems good.
/lgtm

PR #273, opened 2020-11-28
[release 4.7] Bug 1913356: Implemented gathering specific logs from openshift apiserver operator

Sergey1011010:
## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

- `config/pod/{namespace-name}/logs/{pod-name}/errors.log`

## Documentation

- ` docs/gathered-data.md `

## Unit Tests

- `pkg/gather/clusterconfig/openshift_apiserver_operator_logs_test.go`

## Privacy

## References

https://issues.redhat.com/browse/CCXDEV-3519
https://bugzilla.redhat.com/show_bug.cgi?id=1913356
https://access.redhat.com/solutions/5448851


tisnik (approved):
LGTM

pkg/gather/clusterconfig/openshift_apiserver_operator_logs.go
```diff
@@ -0,0 +1,112 @@
...
+
+	pods, err := coreClient.Pods(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	var records []record.Record
+
+	for _, pod := range pods.Items {
+		request := coreClient.Pods(namespace).GetLogs(pod.Name, &corev1.PodLogOptions{
```
tremes, line 62: This relies on the fact that there's only one pod in the `penshift-apiserver-operator` namespace right? It's probably minor but I would suggest to narrow down the filtering to one given pod, because there can be more pods in the future.

pkg/gather/clusterconfig/openshift_apiserver_operator_logs.go
```diff
@@ -0,0 +1,112 @@
...
+			LimitBytes:   &limitBytes,
+		})
+
+		logs, err := filterLogs(ctx, request, messagesToSearch)
+		if err != nil {
+			return nil, err
+		}
+
+		records = append(records, record.Record{
+			Name:        "logs/openshift-api-server-operator",
```
tremes, line 73: I think we already store some pod logs elsewhere in the archive (I think in case of unhealthy pods). I would suggest to use the same or similar dir. What others think?

tremes: It would be better to also include the pod name.

pkg/gather/clusterconfig/openshift_apiserver_operator_logs.go
```diff
@@ -0,0 +1,112 @@
...
+			Item:        Raw{logs},
```
tremes, line 74: The `logs` can be empty string right? I think it's weird to have empty files in the IO archive, but maybe it's OK. We should discuss this with team creating CCX rules.

pkg/gather/clusterconfig/openshift_apiserver_operator_logs.go
tremes: See https://github.com/openshift/insights-operator/blob/master/pkg/gather/clusterconfig/operators.go#L164. We may already have the log, but we don't have to so the question is how to handle these situations.

pkg/gather/clusterconfig/openshift_apiserver_operator_logs.go
Sergey1011010: it actually collects logs from all the pods from the given namespace

pkg/gather/clusterconfig/openshift_apiserver_operator_logs.go
Sergey1011010: Yeah, it doesn't make sense to have an empty file, thanks, will fix it

pkg/gather/clusterconfig/openshift_apiserver_operator_logs.go
Sergey1011010: It seems strange to have logs under "config" folder, but if we already store logs somewhere, then, sure, I'll change it :)

pkg/gather/clusterconfig/openshift_apiserver_operator_logs.go
tremes: Yes but I think we are interested only in "openshift-apiserver-operator" pod log right?

Sergey1011010: Actually the pod's name is something like "openshift-apiserver-operator-6ddb679b87-4kn55" so we can only find all or some pods from namespace "openshift-apiserver-operator", I don't think there will be other unrelatable pods in this namespace

go.mod
```diff
@@ -3,34 +3,77 @@ module github.com/openshift/insights-operator
...
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0
 	github.com/openshift/library-go => github.com/openshift/library-go v0.0.0-20200421122923-c1de486c7d47
-	k8s.io/api => k8s.io/api v0.18.9
-	k8s.io/apimachinery => k8s.io/apimachinery v0.18.9
-	k8s.io/apiserver => k8s.io/apiserver v0.18.9
-	k8s.io/client-go => k8s.io/client-go v0.18.9
+	k8s.io/api => k8s.io/api v0.19.4
+	k8s.io/apimachinery => k8s.io/apimachinery v0.19.4
+	k8s.io/apiserver => k8s.io/apiserver v0.19.4
+	k8s.io/client-go => k8s.io/client-go v0.19.4
```
tremes, line 78: I don't think we want to update the K8s API in this PR.

go.mod
```diff
@@ -3,34 +3,77 @@ module github.com/openshift/insights-operator
...
 	github.com/grpc-ecosystem/go-grpc-middleware v1.1.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway v1.11.3 // indirect
+	github.com/imdario/mergo v0.3.7 // indirect
+	github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible // indirect
+	github.com/kubernetes-sigs/kube-storage-version-migrator v0.0.0-20191127225502-51849bc15f17 // indirect
+	github.com/morikuni/aec v1.0.0 // indirect
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/opencontainers/go-digest v1.0.0-rc1 // indirect
+	github.com/opencontainers/image-spec v1.0.1 // indirect
+	github.com/opencontainers/runc v0.0.0-20191031171055-b133feaeeb2e // indirect
```
tremes, line 41: I guess there's something wrong with the dependencies here. I believe we don't need so many new deps.

go.mod
Sergey1011010: Without that, there's no way to test it(streaming logs is not implemented in the previous version), so we need to either remove test and revert dependencies or keep it as is.

go.mod
Sergey1011010: They are used somewhere in new version of k8s api, so Go added them here.

go.mod
tremes: Well OK. Openshift API is using K8s API `v0.19.2` (see https://github.com/openshift/api/blob/master/go.mod) so we can probably update to the same version, but I would split it into 2 PRs - one doing the versions update and second the actual log gathering. 

@natiiix Any recommendations for the version updates? I know you've already done it.

tremes: I just found out that the API is about to be updated to 1.20 (not sure when and 1.20 looks like to be equal to v0.20) so if we don't want to wait we will need to do the update twice....but maybe we can just stay on the 0.19.2

Sergey1011010: I mean, streaming logs is in "k8s.io/client-go", but we need to update other dependencies as well since there will be a dependency conflict.

tremes:
@Sergey1011010 Please rebase

pkg/gather/clusterconfig/openshift_apiserver_operator_logs_test.go
```diff
@@ -0,0 +1,44 @@
...
+	ctx := context.Background()
+
+	_, err := coreClient.Pods(testPodNamespace).Create(
+		ctx,
+		&corev1.Pod{ObjectMeta: metav1.ObjectMeta{
+			Name: testPodNamespace,
+		}},
+		metav1.CreateOptions{},
+	)
+	helpers.FailOnError(t, err)
```
tremes, line 30: I am asking myself. Is it worth of adding this new dependency because of using one function in the tests?

tremes: and I tend to say: No it's not :)

Sergey1011010: Well, there are some other functions https://github.com/RedHatInsights/insights-operator-utils/tree/master/tests/helpers which we may need in the future, but I can remove it, no problems

tremes: When looking at the change in `go.sum` (toooo many new stuff) then I would remove it. I would keep the unit tests as simple as possible (we can improve the integration tests once we have the independent CI image). Thanks and sorry.

0sewa0:
/lgtm

tested it on clusterbot cluster, the logs weren't collected when the specified error wasn't present, and they were collected when the error was present

PR #279, opened 2020-12-03
Bug 1907876: Refactoring record and gatherer

0sewa0:
The `record` package is currently is the place where the gather-functions are actually run and it also responsible for storing the output. This is to much responsibility for one component and we have the `Gatherer` which only collects the gather-functions but not actually runs them.

So the `record`package should only care about the storing of the data and not actually creating it, and the `Gatherer` should be responsible for managing the the gather-functions. (by managing I mean: deciding which to run, handling errors for gathering, [in the future] syncing up the multi-threaded workloads of the gather-functions)

This will allow us to have greater control over how the gathering is done, and will simplify the code base.

Also insightsclient is somewhat linked to  GatherClusterVersion, it undoes this link.

Furthermore makes the gathering configurable via the yaml config.

## Categories
- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive
- Not necessary

## Documentation
- `docs/arch.md` and `docs/gathered-data.md`  have been updated

## Unit Tests
- No new unit tests, old ones have been moved and renamed. (`record/interface_test.go` --> `gather/clusterconfig/0_gatherer_test.go`)

## Privacy
No new data is collected.

## References
https://issues.redhat.com/browse/CCXDEV-3506

### DIFF
--- a/docs/arch.md
+++ b/docs/arch.md
@@ -284,3 +284,10 @@ health_statuses_insights{metric="low"} 1
 health_statuses_insights{metric="moderate"} 1
 health_statuses_insights{metric="total"} 2
 ```
+
+## Configuring what to gather
+In the yaml config there is a field named `gather` it expects a list of strings, each string is an id that is connected to a gather function.
+Adding such an id to the list means that that certain gather function needs to be run.
+If nothing is set in the `gather` list then no gathering will take place and an error will be raised.
+There is a special id named `ALL` which if in the list then every gather function will be run.
+The id of each gather function can be found in the `docs/gathered-data.md` beside the `Id in config:` text for each section.
--- a/docs/arch.md
+++ b/docs/arch.md
@@ -291,3 +291,40 @@ Adding such an id to the list means that that certain gather function needs to b
 If nothing is set in the `gather` list then no gathering will take place and an error will be raised.
 There is a special id named `ALL` which if in the list then every gather function will be run.
 The id of each gather function can be found in the `docs/gathered-data.md` beside the `Id in config:` text for each section.
+
+#### Example for using special id `ALL`
+```
+gather:
+  - ALL
+```
+
+#### Example for using individual ids
+```
+gather:
+ - pdbs
+ - metrics
+ - operators
+ - container_images
+ - nodes
+ - config_maps
+ - version
+ - id
+ - infrastructures
+ - networks
+ - authentication
+ - image_registries
+ - image_pruners
+ - feature_gates
+ - oauths
+ - ingress
+ - proxies
+ - certificate_signing_requests
+ - crds
+ - host_subnets
+ - machine_sets
+ - install_plans
+ - service_accounts
+ - machine_config_pools
+ - container_runtime_configs
+ - stateful_sets
+```
--- a/pkg/gather/clusterconfig/0_gatherer.go
+++ b/pkg/gather/clusterconfig/0_gatherer.go
@@ -34,6 +34,8 @@ type Gatherer struct {
 
 type gatherFunction func(g *Gatherer) ([]record.Record, []error)
 
+const gatherAll = "ALL"
+
 var gatherFunctions = map[string]gatherFunction{
 	"pdbs": GatherPodDisruptionBudgets,
 	"metrics": GatherMostRecentMetrics,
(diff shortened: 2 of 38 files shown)


pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -34,7 +36,7 @@ func New(gatherKubeConfig *rest.Config, gatherProtoKubeConfig *rest.Config, metr
 // Gather is hosting and calling all the recording functions
 func (g *Gatherer) Gather(ctx context.Context, recorder record.Interface) error {
 	g.ctx = ctx
-	return record.Collect(ctx, recorder,
+	bulkFns := []func() ([]record.Record, []error){
 		GatherPodDisruptionBudgets(g),
```
0sewa0, line 40: so the gather functions are closures that return a function to be run to do the actual gathering.

I think this is unnecessary if we do the gathering inside the gatherer and leaving them in just complicates things.
Of course it will complicate the backports a bit but not by terribly.

Thoughts ?

rluders: LGTM. I think that it would introduce a better readability to the code, removing a necessary layer of complexity.

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -61,19 +60,19 @@ type Source struct {
 var ErrWaitingForVersion = fmt.Errorf("waiting for the cluster version to be loaded")
 
 // New creates a Client
-func New(client *http.Client, maxBytes int64, metricsName string, authorizer Authorizer, clusterInfo ClusterVersionInfo) *Client {
+func New(client *http.Client, maxBytes int64, metricsName string, authorizer Authorizer, gatherKubeConfig *rest.Config) *Client {
```
tremes, line 63: Just out of curiosity - why is it called `gatherKubeConfig` ? Wouldn't just `kubeConfig` be suficient?

0sewa0: True, but the reason I named it `gatherKubeConfig` is that it uses the `gatherKubeConfig` from the `operator.go` which is the specific config for gathering the data that the `insightsclient` will need to gather, so I think its clearer this way (for me atleast :D ).

The same could be brought up in the gatherer, it also uses `gatherKubeConfig`, plus `gatherProtoKubeConfig` and `metricsGatherKubeConfig`. (`gatherKubeConfig` and `gatherProtoKubeConfig` differ very little)

So we have several configs and I would like to keep note which part should use which and I know no better way then this.

tremes: Ahh right. Thanks for explanation. Makes sense then....and I should probably check why we have more of these.

pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -1,25 +1,68 @@
...
-	lock                    sync.Mutex
-	lastVersion             *configv1.ClusterVersion
+}
+
+type gatherFunction func(g *Gatherer) ([]record.Record, []error)
+
+const gatherAll = "ALL"
+
+var gatherFunctions = map[string]gatherFunction{
+	"pdbs": GatherPodDisruptionBudgets,
```
Sergey1011010, line 40: maybe it would be easier if ids would match function names(e.g. gather_pod_distribution_budgets or pod_distribution_budgets)

Sergey1011010:
Thanks, it looks much more flexible now

pkg/gather/clusterconfig/0_gatherer.go
0sewa0: yeah, thought about that, currently I used either the name of the go file or the name of the folder inside the archive. (I've chosen based on personal preference :D )

how I think we should do it is have the ids match the name of the resource the could used in `oc get <resource_name>` (I think this is true for `pdbs`)

docs/arch.md
```diff
@@ -283,4 +283,11 @@ health_statuses_insights{metric="important"} 0
 health_statuses_insights{metric="low"} 1
 health_statuses_insights{metric="moderate"} 1
 health_statuses_insights{metric="total"} 2
-```
\ No newline at end of file
+```
+
+## Configuring what to gather
+In the yaml config there is a field named `gather` it expects a list of strings, each string is an id that is connected to a gather function.
+Adding such an id to the list means that that certain gather function needs to be run.
```
tremes, line 290: What about to add some example of the list here?

pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -32,50 +75,92 @@ func New(gatherKubeConfig *rest.Config, gatherProtoKubeConfig *rest.Config, metr
...
+		last++
+		if last != i {
+			arr[last] = arr[i]
+		}
+	}
+	if last < len(arr) {
+		last++
+	}
+	return arr[:last]
+}
```
tremes, line 157: Well I noticed that this is just copy/paste but does this function really work as expected? What is the expected return value for `uniqueStrings([]string{"b", "a", "a", "c", "a"}` ? I would expect `[b a c]` (or similar) but I am seeing `[b a c a]`. 

This https://github.com/openshift/insights-operator/blob/master/pkg/record/interface_test.go#L16 looks suspicious to me as well.

docs/arch.md
0sewa0: good idea

pkg/gather/clusterconfig/0_gatherer.go
0sewa0: yeah, this looks overly complicated for what it actually meant to do and that test case is just totally wrong (or its some moon logic that I don't understand)
should be easy to fix

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -130,6 +129,19 @@ func userAgent(releaseVersionEnv string, v apimachineryversion.Info, cv *configv
 	return fmt.Sprintf("insights-operator/%s cluster/%s", gitVersion, cv.Spec.ClusterID)
 }
 
+func (c *Client) getClusterVersion() (*configv1.ClusterVersion, error) {
+	if c.clusterVersion != nil {
+		return c.clusterVersion, nil
+	}
+	ctx := context.Background()
+	cv, err := clusterconfig.GetClusterVersion(ctx, c.gatherKubeConfig)
```
tremes, line 137: I am not sure I like this dependency to `clusterconfig`. Please take a look at this https://github.com/RedHatInsights/insights-ingress-http-client/blob/main/insights/insightsclient/insightsclient.go#L149. It would probably mean to pass the clusterversion in `operator.go`. Can you please try it?

0sewa0: yeah, but if I "just" pass in the version to the client it still has to come from somewhere (something needs to gather it), so that is why it uses the clusterconfig.GetClusterVersion that is just the base logic to get the cluster version which as independent as it gets,
if we plan to move the insightsclient into separate project then a copy of that gather logic will need to go with it

tremes:
I experimented it with this little bit and I think it works as expected. Thanks. 
/lgtm

pkg/insights/insightsclient/insightsclient.go
tremes: Yes good point. It would probably make more sense to move the cluster version gathering to a potential new client lib I created https://github.com/RedHatInsights/insights-ingress-http-client/issues/9

PR #280, opened 2020-12-03
[WIP] new parameters TestArchiveContains 

psimovec:
rn: Problems with creating a machineset resource. Unfortunately, it works just locally

CollectingStatefulSets failed for me locally. And it's not supposed to fail... (Didn't skip the test to see if it fails on CI too)


tremes:
Integration tests have been moved to the internal repository.

PR #281, opened 2020-12-04
Fix bug in statefulset gatherer & minor doc fix

tremes:


psimovec:
The same fix as in https://github.com/openshift/insights-operator/pull/274 (which helped)
/lgtm

rluders (approved):
/lgtm

PR #282, opened 2020-12-04
Adds github pull request template.

rluders:
This PR adds the Github PR template as [CCXDEV-3569](https://issues.redhat.com/browse/CCXDEV-3569) requests.

Sample repository:
https://github.com/rluders/github-templates-test

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,7 +15,7 @@ This PR implements a new data enhancement to...
 
 ## Documentation
 
-- `path/to/documentation.json`
+- `path/to/documentation.md`
 
 ## Tests
 <!-- If it includes new tests, list them down bellow -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,35 @@
+<!-- Short description of the PR. What does it do? -->
+This PR implements a new data enhancement to...
+
+## Categories
+<!-- Select the categories that your PR better fits on -->
+
+- [ ] Bugfix
+- [X] Enhancement
+- [ ] Backporting
+- [ ] Others (CI, Infrastructure, Documentation)
+
+## Sample archive
+
+- `path/to/sample_data.json`
+
+## Documentation
+
+- `path/to/documentation.json`
+
+## Tests
+<!-- If it includes new tests, list them down bellow -->
+
+- `path/to/file_test.go`
+
+## Privacy
+<!-- It has anonymization/privacity beein considered by CCX? (e.g. external IP addresses) -->
+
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
...
+- [ ] Backporting
+- [ ] Others (CI, Infrastructure, Documentation)
+
+## Sample archive
+
+- `path/to/sample_data.json`
+
+## Documentation
+
+- `path/to/documentation.json`
```
0sewa0, line 18: never seen docs in json before :)

0sewa0: looks good otherwise

rluders: Oh... Copy and Paste. Let me change it.

tremes:
I would change the order of the parts as in the proposal. The testing should be second and so on.

.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
...
+
+- `path/to/documentation.md`
+
+## Tests
+<!-- If it includes new tests, list them down bellow -->
+
+- `path/to/file_test.go`
+
+## Privacy
+<!-- It has anonymization/privacity beein considered by CCX? (e.g. external IP addresses) -->
```
tremes, line 26: Typo and I would suggest to formulate the question better.

tremes: Ahh it's not

.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Categories
+<!-- Select the categories that your PR better fits on -->
+
+- [ ] Bugfix
+- [X] Enhancement
+- [ ] Backporting
+- [ ] Others (CI, Infrastructure, Documentation)
+
+## Sample archive
```
tremes, line 12: This can be checkbox as well right.

tremes: I mean I would really put it as a question.

.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
...
+
+- `path/to/sample_data.json`
+
+## Documentation
+
+- `path/to/documentation.md`
+
+## Tests
```
tremes, line 20: Too general. We should explicitly say unit tests because we don't usually add any e2e tests

.github/pull_request_template.md
rluders: Hummm... If it doesn't have any sample archive, just put nothing there, if it has put the sample data path. Don't think that add a checkbox or just a question would be helpful. Otherwise, we should do it for all the other lists.

PR #288, opened 2020-12-07
update k8s api to 0.19.2

Sergey1011010:
It's needed for https://github.com/openshift/insights-operator/pull/273


tremes:
This is already adressed by https://github.com/openshift/insights-operator/pull/285

PR #289, opened 2020-12-08
Check context status when checking container is running OK

tremes:
This is additional commit for https://github.com/openshift/insights-operator/commit/b7da048d7375d1ebf5dc0750926637dcdf3b046f. 
We shouldn't check the condition if the context was cancelled or done meanwhile. This is unnecessary waiting and can lead to error state.


0sewa0:
Pretty simple :)

/lgtm

PR #291, opened 2020-12-08
Bug 1918351: Gather SAP configuration (SCC & ClusterRoleBinding)

natiiix:
This enhancements gathers `anyuid` and `privileged` SCCs, and the corresponding ClusterRoleBindings from clusters running a SAP payload.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

- ~~`config/sap_namespaces.json`~~ (removed because it's no longer necessary)
- `config/securitycontextconstraint/`
- `config/clusterrolebinding/`

## Documentation

- `docs/gathered-data.md` - "SAPConfig" section

## Unit Tests

- `pkg/gather/clusterconfig/sap_config_test.go`

## Privacy

## Changelog

No.

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-3485


pkg/gather/clusterconfig/sap_config.go
```diff
@@ -0,0 +1,113 @@
...
+	if errors.IsNotFound(err) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	sapNamespaces := map[string]struct{}{}
+	for _, item := range datahubsList.Items {
+		sapNamespaces[item.GetNamespace()] = struct{}{}
```
tremes, line 71: That looks a little unnecessary. Can't we have just a list/array of sapNamespaces names?

pkg/gather/clusterconfig/sap_config.go
```diff
@@ -0,0 +1,113 @@
...
+		return
+	}
+
+	records, errors := gatherSAPConfig(g.ctx, gatherDynamicClient, gatherKubeClient.CoreV1(), gatherSecurityClient, gatherAuthClient)
+	c <- gatherResult{records: records, errors: errors}
+}
+
+func gatherSAPConfig(ctx context.Context, dynamicClient dynamic.Interface, coreClient corev1client.CoreV1Interface, securityClient securityv1client.SecurityV1Interface, authClient authclient.AuthorizationV1Interface) ([]record.Record, []error) {
+	sccFilter := map[string]struct{}{"anyuid": {}, "privileged": {}}
+	crbFilter := map[string]struct{}{"system:openshift:scc:anyuid": {}, "system:openshift:scc:privileged": {}}
```
tremes, line 57: Do we need maps for these? Wouldn't be array of string sufficient?

docs/insights-archive-sample/config/securitycontextconstraint/privileged.json
```diff
@@ -0,0 +1 @@
...ernetes.io/description":{},"f:release.openshift.io/create-only":{}}},"f:priority":{},"f:readOnlyRootFilesystem":{},"f:requiredDropCapabilities":{},"f:runAsUser":{".":{},"f:type":{}},"f:seLinuxContext":{".":{},"f:type":{}},"f:seccompProfiles":{},"f:supplementalGroups":{".":{},"f:type":{}},"f:volumes":{}}},{"manager":"kubectl-edit","operation":"Update","apiVersion":"security.openshift.io/v1","time":"2020-11-26T12:53:56Z","fieldsType":"FieldsV1","fieldsV1":{"f:users":{}}}]},"priority":null,"allowPrivilegedContainer":true,"defaultAddCapabilities":null,"requiredDropCapabilities":null,"allowedCapabilities":["*"],"allowHostDirVolumePlugin":true,"volumes":["*"],"allowHostNetwork":true,"allowHostPorts":true,"allowHostPID":true,"allowHostIPC":true,"allowPrivilegeEscalation":true,"seLinuxContext":{"type":"RunAsAny"},"runAsUser":{"type":"RunAsAny"},"supplementalGroups":{"type":"RunAsAny"},"fsGroup":{"type":"RunAsAny"},"readOnlyRootFilesystem":false,"users":["system:admin","system:serviceaccount:openshift-infra:build-controller","system:serviceaccount:sdi:mlf-deployment-api"],"groups":["system:cluster-admins","system:nodes","system:masters"],"seccompProfiles":["*"],"allowedUnsafeSysctls":["*"]}
```
tremes, line 1: This is little bit weird. All these resources are gathered without their `kind` and `apiVersion`. It's probably not an issue, but it's not very consistent (given the format of all our resources)

natiiix: That's a very good catch, but we're not doing that anywhere intentionally. It just seems that this client doesn't return those fields when using the `Get` method for some strange reason.

pkg/gather/clusterconfig/sap_config.go
natiiix: You're right. It would work fine, but this syntax ensures that each resource is collected at most once. Originally, there was a `name` field check when iterating over a list of returned results instead of using `Get`, which checked if the name existed as a key in this map (which is actually the canonical Go syntax for a set) because it's faster and shorter than iterating over a list. I can change it if you want, but I think this should be fairly clear to anyone aware of this somewhat unusual syntax for sets).

pkg/gather/clusterconfig/sap_config.go
natiiix: Oh yeah, the reason behind this is historical. Now we no longer really care about the particular namespace, in which the resource exists, so we don't really need to collected this information at all. Originally, there was a loop iterating over the namespaces a looking for SCCs/CRBs that match them. It would be a problem to use a list if there were multiple `datahubs` resources in the namespace (not sure if that's even possible at all). I'll change this one to just a boolean value for simplicity. We don't need the actual namespaces anywhere in the code or in the archive. Wdyt?

tremes:
Unfortunately we don't have the option to verify this directly (because we don't have any cluster with SAP installed). I only checked that nothing is added when the SAP is not installed. We verified this approach in remote RH cluster with SAP installed (by sharing IO image including this change).
/lgtm

PR #292, opened 2020-12-08
Update initial waiting times and give TestIsIOHealthy more time

tremes:
`TestIsIOHealthy` was failing, because we were waiting too long (more than 10 min) in case of the previous container state was not healthy. 
This is to shorten the waiting time to max 330 seconds (five and a half minutes) and add more time to `TestIsIOHealthy` test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ test-unit:
 .PHONY: test-unit
 
 test-e2e:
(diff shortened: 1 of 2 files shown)


0sewa0:
/lgtm

Lets see if this can pass the CI

PR #300, opened 2020-12-18
[release 4.7] Bug 1913343: Added changelog file

Sergey1011010:
Added a changelog file not including minor changes and changes before October 27.  

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

No changes

## Documentation

No changes

## Unit Tests

No new unit tests

## Privacy

Not necessary

## References

https://issues.redhat.com/browse/CCXDEV-3621
https://bugzilla.redhat.com/show_bug.cgi?id=1913343


rluders:
I think that it is fine, but don't we want to autogenerate this CHANGELOG file? We could get this information from the PR template.

Sergey1011010:
@rluders, I think not all the changes should be in changelog file, but only sufficient ones. We probably need to discuss that after holidays :)

tremes:
It should be `CHANGELOG.md` and would be nice to have similar formatting (with links) as in https://github.com/openshift/cluster-monitoring-operator/blob/master/CHANGELOG.md. I would also add more for 4.6 ( at least `Adds gatherer for PodDisruptionBudget`)

tremes:
@rluders is there some way how to autogenerate it? I agree with Sergey that we don't want everything in the changelog, but maybe it would be still possible to generate it somehow.

rluders:
@tremes yep. It is possible to autogenete the changelog. I guess that we could even put some rules to WHAT should be included. Let me check it.

PR #302, opened 2021-01-06
Bug 1917367: Refactor periodic.go

0sewa0:
Refactors `periodic.go` in a way that it doesn't use overly complicated workload management (it should be the responsibility of the gatherers to manage the workloads) and as it was written ~2 years ago and wasn't changed since so it doesn't fit how the actual codebase developed since then. (we only have 1 Gatherer but that has multiple independent gather functions, yet `periodic.go` tries to manage multiple Gatherers)

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive
- Shouldn't change

## Documentation

- TBD

## Unit Tests

- TBD

## Privacy
Has no effect on privacy

## References
None

### DIFF
--- a/pkg/controller/periodic/periodic.go
+++ b/pkg/controller/periodic/periodic.go
@@ -27,8 +27,6 @@ type Controller struct {
 	recorder     record.FlushInterface
 	gatherers    map[string]gather.Interface
 	statuses     map[string]*controllerstatus.Simple
-
-	initialDelay time.Duration
 }
 
 func New(configurator Configurator, recorder record.FlushInterface, gatherers map[string]gather.Interface) *Controller {
@@ -61,7 +59,18 @@ func (c *Controller) Sources() []controllerstatus.Interface {
 func (c *Controller) Run(stopCh <-chan struct{}, initialDelay time.Duration) {
 	defer utilruntime.HandleCrash()
 	defer klog.Info("Shutting down")
-	c.initialDelay = initialDelay
+
+	// Runs a gather after startup
+	if initialDelay > 0 {
+		select {
+		case <-stopCh:
+			return
+		case <-time.After(initialDelay):
+			c.Gather()
+		}
+	} else {
+		c.Gather()
+	}
 
 	go wait.Until(func() { c.periodicTrigger(stopCh) }, time.Second, stopCh)
 
@@ -108,16 +117,6 @@ func (c *Controller) runGatherer(name string) error {
 func (c *Controller) periodicTrigger(stopCh <-chan struct{}) {
 	configCh, closeFn := c.configurator.ConfigChanged()
 	defer closeFn()
-
-	if c.initialDelay > 0 {
-		select {
-		case <-stopCh:
-			return
-		case <-time.After(c.initialDelay):
-			c.initialDelay = 0
-			c.Gather()
-		}
-	}
 
 	interval := c.configurator.Config().Interval
 	klog.Infof("Gathering cluster info every %s", interval)
--- a/pkg/controller/periodic/periodic.go
+++ b/pkg/controller/periodic/periodic.go
@@ -133,7 +133,7 @@ func (c *Controller) periodicTrigger(stopCh <-chan struct{}) {
 			interval = newInterval
 			klog.Infof("Gathering cluster info every %s", interval)
 
-		case <-time.After(wait.Jitter(interval/4, 2)):
+		case <-time.After(interval):
 			c.Gather()
 		}
 	}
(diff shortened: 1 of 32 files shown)


pkg/controller/periodic/periodic.go
```diff
@@ -24,101 +23,97 @@ type Configurator interface {
...
+		err := c.runGatherer(name)
+		if err == nil {
+			klog.V(4).Infof("Periodic gather %s completed in %s", name, time.Since(start).Truncate(time.Millisecond))
+			c.statuses[name].UpdateStatus(controllerstatus.Summary{Healthy: true})
+			return
+		}
+		utilruntime.HandleError(fmt.Errorf("%v failed after %s with: %v", name, time.Since(start).Truncate(time.Millisecond), err))
+		c.statuses[name].UpdateStatus(controllerstatus.Summary{Reason: "PeriodicGatherFailed", Message: fmt.Sprintf("Source %s could not be retrieved: %v", name, err)})
+	}
+}
```
tremes, line 83: I think these two methods would deserve some comment/documentation.

pkg/controller/periodic/periodic.go
```diff
@@ -24,101 +23,97 @@ type Configurator interface {
...
+	return gatherer.Gather(ctx, c.configurator.Config().Gather, c.recorder)
 }
 
 func (c *Controller) periodicTrigger(stopCh <-chan struct{}) {
-	configCh, closeFn := c.config.ConfigChanged()
+	configCh, closeFn := c.configurator.ConfigChanged()
 	defer closeFn()
 
-	interval := c.config.Config().Interval
+	if c.initialDelay > 0 {
```
tremes, line 106: What happens if the `initialDelay = 0`? Will the gathering be executed immediately? It doesn't seem so....but maybe I am missing something

pkg/controller/periodic/periodic.go
```diff
@@ -24,101 +23,97 @@ type Configurator interface {
...
+		case <-stopCh:
+			return
+		case <-time.After(c.initialDelay):
+			c.initialDelay = 0
+			c.Gather()
+		}
+	}
+
+	interval := c.configurator.Config().Interval
 	expireCh := time.After(wait.Jitter(interval, 0.5))
```
tremes, line 117: Do we still need the `expireCh`?

pkg/controller/periodic/periodic.go
rluders: Yep. It looks weird. AFIK, the Gather will not be executed if the `initialDelay` is less or eq to zero. I think that this behavior is not desired. Can we put some default value for it?

rluders (changes requested):

pkg/controller/periodic/periodic.go
tremes: `InitialDelay` has the default value 0 - https://github.com/openshift/insights-operator/blob/master/pkg/controller/operator.go#L63

pkg/controller/periodic/periodic.go
0sewa0: no idea, its something that was in the original and I didn't dare to remove it
will check what will happen if I remove it

pkg/controller/periodic/periodic.go
0sewa0: if `initialDelay == 0` then the gathering will start as normal at https://github.com/openshift/insights-operator/pull/302/files#diff-dec12f0cf28a91a837b46e755b46c081c9014121cdb770b2cd8ac916af29c8deR133

0sewa0: or do we want to start a gather immediately if no `initialDelay` is set

tremes: @0sewa0 yes I think so

pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -88,40 +93,40 @@ func (g *Gatherer) Gather(ctx context.Context, gatherList []string, recorder rec
...
+		for _, record := range gatherResults.records {
 			if err := recorder.Record(record); err != nil {
 				errors = append(errors, fmt.Sprintf("unable to record %s: %v", record.Name, err))
 				continue
 			}
 		}
-		if err := ctx.Err(); err != nil {
-			return err
-		}
+		klog.V(4).Infof("Read from %s's channel and received %s\n", gatherName, value.String())
```
tremes, line 129: I would log this in level 5 or higher.WDYT?

pkg/controller/periodic/periodic.go
0sewa0: updated it
it should run a gather immediately if `initialDelay` is 0 or less

tremes: thanks

tremes:
I experimented with that locally and it seems to work OK. Thanks
/lgtm

PR #303, opened 2021-01-06
Improve code removing some codesmells

rluders:
This PR improves the code by removing some codesmells based on linter tools feedback and adding some extra error checks.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

No.

## Documentation

No.

## Unit Tests

None.

## Privacy

## References

None.


cmd/gendoc/main.go
```diff
@@ -284,9 +301,15 @@ func execExampleMethod(methodFullPackage, methodPackage, methodName string) (str
 		fmt.Print(str)
 	}
 		`, methodFullPackage, methodPackage, methodName))
-	tf.Close()
+	if err != nil {
+		fmt.Print(err)
+	}
+
+	defer tf.Close()
```
0sewa0, line 308: IMO, closing something that we know for a fact that wont be used anymore shouldn't be a codesmell

0sewa0:
Looks good, only had 1 small issue.

Sergey1011010:
Looks great

cmd/gendoc/main.go
rluders: I guess that we can remove it. Normally GoLang GC should close `os.File` resources by default in this case, when it hits the end of the program.

pkg/controller/status/status.go
```diff
@@ -338,9 +338,15 @@ func (c *Controller) Start(ctx context.Context) error {
...
+				err := limiter.Wait(ctx)
+				if err != nil {
+					klog.Errorf("Limiter error: %v", err)
+				}
 			case <-c.statusCh:
-				limiter.Wait(ctx)
+				err := limiter.Wait(ctx)
+				if err != nil {
+					klog.Errorf("Limiter error: %v", err)
+				}
```
tremes, line 349: Maybe it would be good to distinguish these two errors.

pkg/record/diskrecorder/diskrecorder.go
```diff
@@ -245,6 +245,9 @@ func (r *Recorder) PeriodicallyPrune(ctx context.Context, reported AlreadyReport
 				}
 				return true, nil
 			})
+			if err != nil {
+				klog.V(2).Infof("Failed to wait for exponential backoff : %v", err)
```
tremes, line 249: This would deserve better message IMO. Is it only info level?

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -187,16 +187,16 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
...
 		}
 		r := &LimitedReader{R: source.Contents, N: c.maxBytes}
 		n, err := io.Copy(fw, r)
 		bytesRead = n
 		if err != nil {
-			pw.CloseWithError(err)
+			_ = pw.CloseWithError(err)
 		}
-		pw.CloseWithError(mw.Close())
+		_ = pw.CloseWithError(mw.Close())
```
tremes, line 199: Just a cosmetic change, but shouldn't we handle the error?

test/integration/log_checker.go
```diff
@@ -153,7 +153,7 @@ func (lc *LogCheck) Execute() *LogCheck {
 		lc.firstCheck = now
 	}
 	r := regexp.MustCompile(`(?m)` + lc.searching)
-	logOptions := &corev1.PodLogOptions{}
+	var logOptions *corev1.PodLogOptions
```
tremes, line 156: I don't get this change. I think we use the original declaration throughout our code base.

pkg/insights/insightsclient/insightsclient.go
0sewa0: as far as I know, its ok to not handle errors, but you should explicitly declare that you are not doing it.
in the original the fact that an error can happen is hidden which quite scary, here its clear that it was decided to not handle such an error

test/integration/log_checker.go
0sewa0: the fact that its used throughout the codebase doesn't mean its right :) (both is used semi randomly in our code)
I personally prefer to use `var logOptions *corev1.PodLogOptions` because its just visually clearer for me

pkg/insights/insightsclient/insightsclient.go
tremes: ok ok. I was looking at https://golang.org/pkg/io/#PipeWriter.CloseWithError and it's probably ok.

test/integration/log_checker.go
tremes: Sure. It's matter of preference. :)

pkg/insights/insightsclient/insightsclient.go
rluders: What @0sewa0 said. When you chose to commit an error handling it must be explicit in the code, otherwise, the linter (errcheck) will consider that you not verified the error/not know about its possibility, and leaving an error without being handled is, in some cases, a potential bug. So, it is not cosmetic, also, I must say that I didn't pay that much attention to the `insightsclient` part of the code, 'cause if I remember we gonna replace it with an external dependency. Isn't that correct?

test/integration/log_checker.go
rluders: It solves the `ineffectual assignments`. Let me try to explain it. Before this change we had:

`` logOptions := &corev1.PodLogOptions{}```

And right after that, we have an `if-else` statement, that will obligatorily instantiate a new value to `logOptions`, in this case, the first value will never be used, 'cause it will be replaced by the code inside the `if-else`. Did you get it? Is like, we define the variable, but do not initiate the struct, leaving it to the code inside the `if-else` statement.

pkg/record/diskrecorder/diskrecorder.go
rluders: @tremes any suggestion about the message text? How about changing it to `klog.Errorf`?

test/integration/log_checker.go
tremes: OK thanks for explanation.

pkg/record/diskrecorder/diskrecorder.go
```diff
@@ -245,6 +245,9 @@ func (r *Recorder) PeriodicallyPrune(ctx context.Context, reported AlreadyReport
 				}
 				return true, nil
 			})
+			if err != nil {
+				klog.V(4).Infof("Fail to properly prune last report within %s: %v", interval.Truncate(time.Second), err)
```
tremes, line 249: It's much better now thanks. However I think the interval is longer than 1s.

pkg/record/diskrecorder/diskrecorder.go
```diff
@@ -245,6 +245,9 @@ func (r *Recorder) PeriodicallyPrune(ctx context.Context, reported AlreadyReport
 				}
 				return true, nil
 			})
+			if err != nil {
+				klog.V(4).Infof("Fail to properly prune last report within %s: %v", interval.Truncate(time.Minute), err)
```
tremes, line 249: Is it a minute? wait.Backoff{Duration: time.Second, Steps: 4, Factor: 1.5} I am not sure how the backoff works but I guess the worst case (when the timeout exceeds) is 1s * 1,5 * 1.5 * 1.5 * 1.5 in this case or maybe just 1.5 * 4 ...not sure.

tremes:
I would suggest to reopen this against release-4.8 because I wouldn't create a new bugzilla issue for this.

pkg/record/diskrecorder/diskrecorder.go
rluders: As I told you on the channel, I tested it, and Minutes works fine.

0sewa0:
tested it locally, everything seems to work fine,
Nice cleanup
/lgtm

PR #307, opened 2021-01-08
Bug 1914161: Gather network OVS flows

tremes:
Gathering network OVS flows from every OVS pod in the "openshift-sdn" namespace.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

Sample archive updated. See `docs/insights-archive-sample/networking/ovs_flows/ovs-b9f5j ` and others.

## Documentation

Documentation updated in ` docs/gathered-data.md `

## Unit Tests

- No unit test added. Thiw requires a new integration test.

## Privacy

No sensitive data

## Changelog

Yes changelog updated.

## References

https://issues.redhat.com/browse/CCXDEV-3690
https://bugzilla.redhat.com/show_bug.cgi?id=1914161


pkg/gather/clusterconfig/ovs_flows.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+const sdnNamespace = "openshift-sdn"
+const ovsCommand = "ovs-ofctl -O OpenFlow13 dump-flows br0"
+
+// GatherOVSFlows collects OVS flow information for every OVS pod in the openshift-sdn namespace
+//
+// Location in archive: networking/ovs_flows/
```
tremes, line 23: Any thoughts about this location? I think it would be better to put it here and keep only pure Kubernetes/OpenShift stuff in the `config`.

tremes: But maybe it's more confusing....not sure

rluders: I think that makes sense to not store on the same path that we are storing k8s/os stuff. If we started to put some 'custom' gathers we may not want to mess up the data sources.

rluders (commented):
Just went through the code, pretty straightforward implementation. Isn't it a good idea to add some unit tests to it?

tremes:
@rluders Any idea for the unit tests? Note that there's no `openshift-sdn` namespace available and thus no OVS at all...we can test that it only returns an empty array of records, but I don't see much value in that.

iNecas:
Just a note from our discussion with Tomas: I'm concerned about role insights-operator starts playing in this PR: so far, it's been only collecting data from APIs exposed by someone else. What this PR does is lest insights-operator to run commands and parse it's output, in context of other pods. There is a risk of insights operator having to know more and more about producing the diagnostics data, which we tend to leave on component owners instead.

I would rather like to get to an architecture similar to Prometheus monitoring, where the components are responsible for providing the desired data via a common API endpoint, and insights-operator responsible for collecting them. What's missing right now is a this common place where this data could be exposed (I'm thinking about what node-exporter for Prometheus).

I would like to see a discussion with OpenShift architects on what options we might have here.

PR #315, opened 2021-01-17
Add a workload fingerprint gatherer

smarterclayton:
The rough structure and shape of workloads on the cluster provides information that is valuable for support scenarios as well as future development. The size, scale, and scope of pods as well as use of init containers is relevant for assessing how to prioritize
improvements to the project. As we see increasing amounts of supply chain security scrutiny, the image hashes running on a cluster can potentially be used to highly compromised images or to identify when certain workloads or known images are highly out of date. Finally, for a subset of workloads with known executables the info about the command being executed (in a hashed version) would allow for analysis of correlations between usage and failures for support.

This gatherer returns no directly identifiable workload information unless a hash collision is feasible. Only namespace names and the first values of command and argument are assessed for now, and command and argument are reduced from their full values to a partial identifier prior to hashing:

    spec.containers.args[0]: "-config.file=/etc/cache-config/cache-conf.yaml" to 
    stripped of values "-config.file" to 
    hashed as "V4JjWhyBxoZx"

or

    spec.containers.args[0]: "/bin/bash" to 
    stripped of path "bash" to 
    hashed as "N9KxLV2avCo2"

Images that have metadata in the cluster are associated with the requests and a running cache is kept to minimize accesses. The gatherer single threads image lookups at 10qps and will exit if the total gather time exceeds a reasonable constant, with subsequent calls potentially being faster due to caching.

The gatherer attempts to minimize the amount of info held in memory and should be reasonably resource efficient. If necessary, the gatherer could be executed at a less frequent interval.

The gatherer on a cluster with 6000 pods (p95 of the fleet) took about 1.8 core seconds to gather and used about 30mb peak during that gather, well within our normal output. The gathered data for that workload was 200kb, about 30k compressed.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

- `docs/insights-archive-sample/config/workload_info.json`

## Documentation

None yet

## Unit Tests

Integration test only currently.

## Privacy

The data gathered preserves deniability due to hashing and retrieves information identifiers that can be correlated to public images or detected via hash collision. No data is returned unmodified that would expose personal information and the subset of workload fingerprinting that is returned is deliberately limited.

## Changelog

Not updated

## References

None yet.


tremes:
Does it make sense to gather this for the OpenShift/Kubernetes namespaces? 

Is it possible to decode the hashed strings? It looks like it won't be possible. I originally thought we wanted to know the firstCommand, firstArguments value so what is the value of gathering those? I think I am missing something.

docs/insights-archive-sample/config/workload_info.json
```diff
@@ -0,0 +1,7166 @@
+{
+  "pods": 5156,
+  "imageCount": 171,
+  "images": {
```
tremes, line 4: I am failing to see why the number of images is so small (in this struct) in my testing clusters (not many external stuff deployments and usually only 3 items in the struct). 

Well the client can't find them in https://github.com/openshift/insights-operator/pull/315/files#diff-7f69d2c3dea3690849da339327364050f1b27a8de85fafc8ee832c0faad051c5R336. 
I am wondering if it's an intention, because it can't find for example the image of apiserver deployment/pods.

docs/insights-archive-sample/config/workload_info.json
```diff
@@ -0,0 +1,7166 @@
...
+              "imageID": "sha256:d4ea0e5042b386273d3121a6c35f6e85f93da6c6840fa29fa106276788c8e39f",
+              "firstCommand": "N9KxLV2avCo2"
+            }
+          ]
+        },
+        {
+          "duplicates": 2,
+          "containers": [
+            {
+              "imageID": "sha256:d4ea0e
```
tremes, line 7166: Is this from some real cluster? I guess so. The math related to the size of the archive doesn't seem so bad in this case. This file is ~180kB (compact json), which will increase the size of the archive by ~30kB.

docs/insights-archive-sample/config/workload_info.json
smarterclayton: These would only be images imported on the cluster, vs images directly referenced.  So yes, not all images would be included.  The goal is to use the image api to selectively get minimal info about the image - if we don't have the image metadata, that's ok.  Down the road, components like the image scanning operator might also report a subset of their info, enriching info about the images.  But for now, the simplest possible structure gathers data that can be correlated.

docs/insights-archive-sample/config/workload_info.json
smarterclayton: Yes this is from one of the CI clusters that has a lot of transient workloads.

docs/insights-archive-sample/config/workload_info.json
tremes: Thanks

pkg/gather/clusterconfig/workloads.go
```diff
@@ -0,0 +1,559 @@
...
+			podShape.Containers, ok = calculateWorkloadContainerShapes(h, pod.Spec.Containers, pod.Status.ContainerStatuses)
+			if !ok {
+				namespacePods.InvalidCount++
+				continue
+			}
+
+			if index := workloadPodShapeIndex(namespacePods.Shapes, podShape); index != -1 {
+				namespacePods.Shapes[index].Duplicates++
+			} else {
+				namespacePods.Shapes = append(namespacePods.Shapes, podShape)
```
tremes, line 214: Do we need the `containers` array to distinguish the pods (for each pod basically)? I mean, now I can get something like: 
```
....
                ,{
                      "containers": [
                        {
                            "imageID": "sha256:bb525a274fcc67b5a8f05882550a77e6586434cf9e78390f2632d6c558456fb2"
                        }
                    ]
                },
                {
                    "containers": [
                        {
                            "imageID": "sha256:3aeeb731c3254f5841cbd679e68ee409d1a1aaabc745c803ae2010f57dbe03cc",
                            "firstCommand": "zC8CUf1mHJ6g"
                        }
                    ]
                },
...
```
but maybe it could be only something like:
```
                {
                    "containers": [
                        {
                            "imageID": "sha256:bb525a274fcc67b5a8f05882550a77e6586434cf9e78390f2632d6c558456fb2"
                        },
                        {
                            "imageID": "sha256:3aeeb731c3254f5841cbd679e68ee409d1a1aaabc745c803ae2010f57dbe03cc",
                            "firstCommand": "zC8CUf1mHJ6g"
                        },
                    ]
```

smarterclayton: knowing the difference between two shapes is mostly so we can tell the difference between certain similar uses of images - the hypothesis is that .there is a fair amount of image reuse but it’s in different pods.  if we collapsed then we’d miss some examples

tremes: OK thanks. I would suggest the following. @smarterclayton can you please remove the WIP? We would merge it then and added some limits (it will run less often + some limit for the number of pods - I would go with 2500), which we can always remove in the future.

smarterclayton: Yeah, let me profile this really quick a few more times to see what i missed.

smarterclayton: So for a very large (p95 of the fleet, roughly) cluster iwth 6000 pods, this used about 30mb memory peak (allocated 150mb total), took 15s, and used about 2 core seconds, which is effectively "free".  Size of gathered was 30k.  

I'm not sure there's a need to limit or restrict size - even on a 600k pod cluster, that's roughly still "free".

smarterclayton:
Updated with two changes:

1. Added `restartAlways bool` to track whether this is a run once (job) or run forever (service) pod to the shape
2. Added firstCommand and firstArg to images consistent with pod shapes

Measured data.

I *think* this is ready for merge for 4.8, but Tomas you tell me :)

pkg/gather/clusterconfig/workloads.go
smarterclayton: I would probably say when we get to 30-50k pods we'd probably want to start limiting.  I can do a follow up which either runs less frequently or simply disables the capture at those levels

smarterclayton:
Verballing confirmed with ivan and tomas we are close - tagging and we'll see how this plays out (can always revert post freeze)

smarterclayton:
Overriding bug but will attach to card since this was in the queue.

pkg/gather/clusterconfig/workloads.go
tremes: Thanks for the analysis. Sounds good, but I am always a little careful. I mean I am not sure on the impact on our processing pipelines so I wanted to start low. Only 30k for the size of the data? It seems little....only the container sha256 is around 100 bytes

0sewa0: I tested it on a cluster-bot cluster where I deployed 20,000 sleeping pods, so its not a perfect test case but my findings are similar to Clayton's.
The increase in CPU usage is nothing extreme (~ 5 core seconds).
The RAM got a bit big if we look at total alloc of 240MB just for this, but I didn't see any major peak in memory usage. 
It took 22 seconds. (The size of the archive didn't change significantly, but that is because the 20,000 duplicates are not stored)
Its definitely the "heaviest" gathering we have currently (on large clusters), but it shouldn't cause trouble.

PR #318, opened 2021-01-21
Bug 1923895: Auto changelog

0sewa0:
Adds a script that can generate a changelog using the local git repo and a github API.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- NOPE

## Documentation

- Updates the README.md

## Unit Tests

- NOPE

## Privacy

## Changelog

NOPE

## References

NOPE


natiiix (commented):
Looks pretty interesting and honestly we probably don't even need anything terribly complicated for this purpose. However, you're missing a trailing newline. :disappointed:

tremes:
Do we need to add this to IO repository?

0sewa0:
> Do we need to add this to IO repository?

Yeah, it uses the git commands and so you need the IO .git where you run it
Its quite normal, openshift and other repos have it their changelog script "in" the repo.
We could move it under `tools` or somewhere if `cmd` is too high profile :)

README.md
```diff
@@ -152,4 +152,17 @@ Starting a web ui at `localhost:8080` to visualize/analyze the profiling data:
...
\ No newline at end of file
+[link](https://jvns.ca/blog/2017/09/24/profiling-go-with-pprof/)
+
+
+### Creating a changelog
+At `./cmd/changelog/main.go` there is a script that can generate a changelog for you.
+
+#### Prerequisites
+It uses both the local git and GitHub`s API to generate the file so:
+- To get info from GitHub you will need to set the `GITHUB_TOKEN` envvar to a GitHub access-token.
```
0sewa0, line 163: I'm a bit conflicted about using the GitHub API, because the only thing we get is the categories defined in the PRs description. That feels a bit of an overkill.

README.md
```diff
@@ -152,4 +152,17 @@ Starting a web ui at `localhost:8080` to visualize/analyze the profiling data:
...
+- Make sure that you have a local, up-to-date copy of each release-branch that might be in the changelog.
+
+#### Use
+It requires 2 command line arguments, `AFTER` and `UNTIL` dates.
+Example: `go run cmd/changelog/main.go 2021-01-10 2021-01-20`
```
Sergey1011010, line 168: Maybe we could add some update mode which would detect the latest change from the file and add only the new things

0sewa0: That would be cool but it would require some regexp magic, which they say is the source of all evil. ;)

I will look into it, this is not an urgent thing and I don't really feel like to have a BZ for this.

tremes:
@0sewa0 would it be possible to revert the order? Newer commits at the top.

README.md
tremes: I quite like the categories.

cmd/changelog/main.go
```diff
@@ -0,0 +1,351 @@
...
+			_, _ = file.WriteString("### Misc\n")
+			_, _ = file.WriteString(fmt.Sprintf("%s\n", misc))
+		}
+	}
+
+}
+
+func pruneChanges(changes []Change) []Change {
+	// TODO: Somehow determine which change is changelog worthy
+	return changes
```
tremes, line 213: I guess if we keep the categories then there's no need to prune anything

tremes:
When will it add 4.8? We already have this branch...maybe it would be better to include it right from the beginning.

cmd/changelog/main.go
```diff
@@ -0,0 +1,351 @@
...
+	bugfix_sectionRegExp      = regexp.MustCompile(`### Bug fixes\n((.+\n)+)`)
+	other_sectionRegExp       = regexp.MustCompile(`### Others\n((.+\n)+)`)
+	misc_sectionRegExp        = regexp.MustCompile(`### Misc\n((.+\n)+)`)
+
+	// PR categories
+	categories = map[string]*regexp.Regexp{
+		"BugFix":      regexp.MustCompile(`- \[[xX]\] Bugfix`),
+		"Enhancement": regexp.MustCompile(`- \[[xX]\] Enhancement`),
+		"Other":       regexp.MustCompile(`- \[[xX]\] Others`),
+		"Backporting": regexp.MustCompile(`- \[[xX]\] Backporting`),
```
tremes, line 36: It looks like these strings repeat more times in various forms. What about to unify it (I guess in the PR template as well) and make them as constants here?

0sewa0:
> When will it add 4.8? We already have this branch...maybe it would be better to include it right from the beginning.

You have to have that branch locally available as 'release-4.8' for the script to work

0sewa0:
> would it be possible to revert the order? Newer commits at the top.

Not really, because the `Change` structs are created after it gets the categories info from GitHub and it does that in an async way so it will be semi random.

tremes:
Nice I like it :) Good job. Thanks
/lgtm

tremes:
Would it be worth of backporting to 4.6 (together with the PR template)?

PR #319, opened 2021-01-25
Bug 1921013: Gather PersistentVolume definition (if any) used in Image registry st…

tremes:
…orage config

This gathers PV (if there's any) used in image registry config as storage. 

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

I didn't add the corresponding PV to the sample archive, because it could look like we gather all the persistent volumes by default. 

## Documentation

- Doc updated in `docs/gathered-data.md`

## Unit Tests

- No update. No way to use "real" PV in the unit tests AFAIK. 

## Privacy

## Changelog

Not yet

## References

https://issues.redhat.com/browse/INSIGHTOCP-54

### DIFF
--- a/pkg/gather/clusterconfig/image_registries.go
+++ b/pkg/gather/clusterconfig/image_registries.go
@@ -5,9 +5,12 @@ import (
 	"fmt"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/v2"
 
 	registryv1 "github.com/openshift/api/imageregistry/v1"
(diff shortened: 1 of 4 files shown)


pkg/gather/clusterconfig/image_registries.go
```diff
@@ -19,28 +22,51 @@ import (
...
+	// if there is some PVC then try to gather used persistent volume
+	if config.Spec.Storage.PVC != nil {
+
+		pvcName := config.Spec.Storage.PVC.Claim
+		pv, err := findPVByPVCName(ctx, coreClient, pvcName)
+		if err != nil {
+			klog.Errorf("unable to find persistent volume: %s", err)
+		} else {
+			pvRecord := record.Record{
+				Name: fmt.Sprintf("config/persistentvolumes/%s", pv.Name),
```
0sewa0, line 64: Wouldn't it make sense to have a subdir specifying where the PVs 'come from' ?
Something like: `Name: fmt.Sprintf("config/persistentvolumes/image-registry/%s", pv.Name),`

tremes: Well I am not sure but I tend to say no. PV is not a namespaced resource AFAIK (guess it's cluster wide) and it's bound to specific PVC (this one is namespaced) that is then used in the IR config. AFAIK not the whole PV must be used when claimed by PVC, then it could be a bit misleading....

rluders: Yep. PV are cluster-wide, not namespaced. Fine to have it as it is in the code.

0sewa0:
Tested it on a cluster-bot cluster:
- Created a PVC and a PV
- Update `configs.imageregistry.operator.openshift.io`  (`oc edit configs.imageregistry.operator.openshift.io cluster`) `storage` section with: 
```
pvc:
      claim: <NAME OF THE PVC>
``` 

/lgtm

PR #320, opened 2021-01-25
WIP: Monitors how many gatherings failed in a row, and applies degraded status accordingly

0sewa0:
Tries to solve the issue described in https://bugzilla.redhat.com/show_bug.cgi?id=1919778

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- not necessary

## Documentation

- not necessary
## Unit Tests

- none

## Privacy

## Changelog

not necessary

## References

https://bugzilla.redhat.com/show_bug.cgi?id=1919778


pkg/controller/periodic/periodic.go
```diff
@@ -81,16 +82,23 @@ func (c *Controller) Run(stopCh <-chan struct{}, initialDelay time.Duration) {
...
 	for name := range c.gatherers {
-		start := time.Now()
-		err := c.runGatherer(name)
-		if err == nil {
-			klog.V(4).Infof("Periodic gather %s completed in %s", name, time.Since(start).Truncate(time.Millisecond))
-			c.statuses[name].UpdateStatus(controllerstatus.Summary{Healthy: true})
-			return
+		retry_count := 0
+		for retry_count < threshold {
+			time.Sleep(interval / time.Duration(threshold) * time.Duration(retry_count))
```
tremes, line 90: So in case of an error we will wait 24 min right (second round 48 min) ? It seems too long....

tremes: Maybe we could try to use https://godoc.org/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoff

0sewa0: Yeah, I think the cleanest would be to try `X` times in between the intervals, this is trying to do that but math is hard and I overshot it a bit, while it should probably do the `X` retries in` interval-5m` or something.

The easiest implementation of this would be to just call `time.Sleep(interval / time.Duration(threshold))` so it will try after 24 minutes, does this 5 times.

Will try it with that https://godoc.org/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoff too, just for the fun of it :)

tremes:
I tested it this locally. When there is an error in an imporant gatherer then the retry logic is correctly applied (gathering is running again after some sime and IO is set as degraded for this time period). When there's no gathering error then everything runs as before. We can mark more gatherers as `failable` (maybe we should come up with better word :) in the future.

PR #325, opened 2021-01-29
Bug 1923888: Fixes error metadata gathering

0sewa0:
Errors produced by the gather-functions are supposed to be stored in the metadata, but because it was left up to go on how to parse/store the errors, so they either showed up correctly or just an empty object, which is not really helpful.

This PR fixes that by just storing the error string (ie.: `error.Error()`).

## Categories

- [x] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- not necessary

## Documentation

- not necessary

## Unit Tests

- not necessary

## Privacy

## Changelog

not necessary

## References

-

### DIFF
--- a/pkg/gather/clusterconfig/0_gatherer.go
+++ b/pkg/gather/clusterconfig/0_gatherer.go
@@ -19,7 +19,7 @@ type gatherStatusReport struct {
 	Name    string        `json:"name"`
 	Elapsed time.Duration `json:"elapsed"`
 	Report  int           `json:"report"`
-	Errors  []error       `json:"errors"`
+	Errors  []string      `json:"errors"`
 }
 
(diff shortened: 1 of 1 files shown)


tremes:
One minor suggestion. It would be nice to change https://github.com/openshift/insights-operator/pull/325/files#diff-95486e6db8a4eaef301b750a6b2feebdcb2aad64f8f3fc601f0d7ca58d6ec197R102 to `var gatherReport []gatheStatusReport`

tremes:
I verified locally that the errors are correctly serialized in the `insights-operator/gathers.json` file in the archive. The naming used in this JSON structure now seems clearer. Thanks
/lgtm

PR #336, opened 2021-02-11
Disable instead of Degrade in case of gather fails

0sewa0:
Sets the status of the operator to disabled on repeated gather failures instead of degraded.  

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- not needed

## Documentation

- not needed

## Unit Tests

- not needed

## Privacy

## Changelog

not needed

## References

https://bugzilla.redhat.com/show_bug.cgi?id=1925659

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -158,11 +158,13 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 				disabledMessage = summary.Message
 			}
 		} else if summary.Operation == controllerstatus.GatheringReport {
+			degradingFailure = false
 			if summary.Count < GatherFailuresCountThreshold {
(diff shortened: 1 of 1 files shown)


pkg/controller/status/status.go
```diff
@@ -163,6 +163,9 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 				degradingFailure = false
 			} else {
 				klog.V(3).Infof("Number of last gather failures %d exceeded the threshold %d. Marking as degraded.", summary.Count, GatherFailuresCountThreshold)
+				degradingFailure = false
+				disabledReason = summary.Reason
+				disabledMessage = summary.Message
```
tremes, line 168: Well we need to update the log messages as well right? The condition can be simplified IMO. We always can set `degradingFailure = false`.

0sewa0: it was so simple that I overlooked these, thx :D

PR #337, opened 2021-02-11
Recorder refactoring that improves maintainability

rluders:
This PR implement improves the recorder maintainability, giving it the possibility to have multiple record drivers and by isolating code responsibility.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

No.

## Documentation

No.

## Unit Tests

No.

## Privacy

## Changelog

No.

## References

https://issues.redhat.com/browse/CCXDEV-2108


tremes:
@rluders no need to create a new bugzilla to merge to master

rluders:
@tremes all right, that is true. :man_facepalming: I always forget about it.

0sewa0:
tested it locally everything is working fine, changes make sense

/lgtm

pkg/recorder/recorder_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+func TestRecord(t *testing.T) {
+	rec := newRecorder()
+	err := rec.Record(record.Record{
+		Name: "config/mock1",
+		Item: tests.RawReport{Data: "mock1"},
+	})
+	assert.Nil(t, err)
```
tremes, line 50: This doesn't test much IMO. I think we can assert that the recorder actually has the new record.

pkg/recorder/recorder_test.go
```diff
@@ -0,0 +1,93 @@
...
+		Name:        "config/mock1",
+		Item:        tests.RawReport{Data: "mock1"},
+		Fingerprint: "abc",
+	})
+	err := rec.Record(record.Record{
+		Name:        "config/mock1",
+		Item:        tests.RawReport{Data: "mock1"},
+		Fingerprint: "abc",
+	})
+	assert.Nil(t, err)
```
tremes, line 65: Same here. We can assert that there's only one (or two?) record available.

pkg/recorder/recorder_test.go
```diff
@@ -0,0 +1,93 @@
...
+func TestFlush(t *testing.T) {
+	rec := newRecorder()
+	for i := range []int{1, 2, 3} {
+		_ = rec.Record(record.Record{
+			Name: fmt.Sprintf("config/mock%d", i),
+			Item: tests.RawReport{Data: "mockdata"},
+		})
+	}
+	err := rec.Flush(context.TODO())
+	assert.Nil(t, err)
```
tremes, line 86: Same here. I guess the records should be cleared from recorder after flush call.

pkg/recorder/recorder_test.go
rluders: So, actually, there isn't a way to get this info from the recorder. Unless if I change something to expose this information to the Recorder structure, or return the recorder item.

pkg/recorder/recorder_test.go
rluders: The same thing above.

pkg/recorder/recorder_test.go
rluders: The same comments above. And the recorder gets clear by [this line](https://github.com/openshift/insights-operator/pull/337/files/70480deff57f2d0eaf7b0483719b405b09411b92#diff-93cbf5b12a5398c622a4f899c81da8bdca98a02a5cf178e90f1b6b43e6b7005cR98).

pkg/recorder/recorder_test.go
tremes: Well you can do something like `assert.Equal(t, 1, len(rec.records))` right

rluders: Oh... You are right. I totally forgot about the fact that 'cause the tests are in the same package we can access non-public struct props.

pkg/recorder/recorder_test.go
rluders: Yep. Bug confirmed. Let me fix it.

tremes:
Please suqash the commits :)

tremes:
Let me experiment:
/tide/merge-method-merge

PR #342, opened 2021-02-12
sap license management logs gatherer

Sergey1011010:
```
The SAP installation may fail in license management pod, so We need error logs to detect the failure.

Get the filtered error logs as followings.

# oc logs deploy/license-management-l4rvh
Found 2 pods, using pod/license-management-l4rvh-74595f8c9b-flgz9
+ iptables -D PREROUTING -t nat -j VSYSTEM-AGENT-PREROUTING
+ true
+ iptables -F VSYSTEM-AGENT-PREROUTING -t nat
+ true
+ iptables -X VSYSTEM-AGENT-PREROUTING -t nat
+ true
+ iptables -N VSYSTEM-AGENT-PREROUTING -t nat
iptables v1.6.2: can't initialize iptables table `nat': Permission denied
Perhaps iptables or your kernel needs to be upgraded.
```
Tested on SAP cluster with different substrings.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

No changes

## Documentation

- `docs/gathered-data.md`

## Unit Tests

No changes

## Privacy

## References

https://issues.redhat.com/browse/CCXDEV-3484
https://bugzilla.redhat.com/show_bug.cgi?id=1928512
https://access.redhat.com/articles/5100521#license-manager-cannot-be-initialized


tremes:
@Sergey1011010 No need for bugzilla at the moment (we are not in feature freeze)

pkg/gather/clusterconfig/sap_license_management_logs.go
```diff
@@ -0,0 +1,46 @@
...
+		c <- gatherResult{nil, []error{err}}
+		return
+	}
+
+	coreClient := gatherKubeClient.CoreV1()
+
+	records, err := gatherLogsFromPodsInNamespace(
+		g.ctx,
+		coreClient,
+		"sdi",
```
tremes, line 31: Are you sure the namespace will always have this name?

Sergey1011010: The SAP installation instruction uses this one, so it should always be the same.

tremes: Well OK and what about if there's no SAP at all and someone has some `sdi` namespace? Then we will be scanning some pod logs in that as well right? I think I don't like this fact.

Makefile
```diff
@@ -1,6 +1,9 @@
 build:
 	go build -mod=vendor -ldflags "-X k8s.io/client-go/pkg/version.gitCommit=$$(git rev-parse HEAD) -X k8s.io/client-go/pkg/version.gitVersion=v1.0.0+$$(git rev-parse --short=7 HEAD)" -o bin/insights-operator ./cmd/insights-operator
 
+build-debug:
```
natiiix, line 4: This stuff should probably go into a separate pull request, or else someone could complain about this later.

debug.Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN make build-debug
+
+FROM registry.ci.openshift.org/ocp/4.8:base
+COPY --from=builder /go/src/github.com/openshift/insights-operator/bin/insights-operator /usr/bin/
+COPY --from=builder /usr/bin/dlv /usr/bin/
+COPY config/pod.yaml /etc/insights-operator/server.yaml
+COPY manifests /manifests
+LABEL io.openshift.release.operator=true
+EXPOSE 40000 40000
+ENTRYPOINT ["dlv", "--listen=:40000", "--headless=true", "--api-version=2", "--accept-multiclient", "exec", "/usr/bin/insights-operator", "--"]
```
natiiix, line 14: Same as above, not really directly related to the enhancement.

pkg/gather/clusterconfig/sap_license_management_logs.go
natiiix: @Sergey1011010 Have you talked to Michal or someone else from the SAP team about this? Because when I asked about the namespace name, he told me that it was arbitrary and that there could easily be multiple installations of SDI on the same cluster, each in its own namespace, of course. As far as I can tell, the namespace name is arbitrary and can be set to whatever the administrator wants. Therefore I don't believe we should many any assumptions about it.

natiiix (changes requested):

Makefile
Sergey1011010: Yeah, it makes sense, thanks.

pkg/gather/clusterconfig/sap_license_management_logs.go
Sergey1011010: Check the new version please

Sergey1011010: btw, we could also store container name in the path of the resulting file, but it would break back compatibility

pkg/gather/clusterconfig/sap_license_management_logs.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func gatherSAPLicenseManagementLogs(
+	ctx context.Context,
+	dynamicClient dynamic.Interface,
+	coreClient corev1client.CoreV1Interface,
+	messagesToSearch []string,
+) ([]record.Record, []error) {
+	datahubsResource := schema.GroupVersionResource{
+		Group: "installers.datahub.sap.com", Version: "v1alpha1", Resource: "datahubs",
```
tremes, line 55: One idea. I know we are doing it just like here, but when there will be a new SAP version available then it will be a bit pain to find all the occurences. What about to extract the GVR to some more widely shared constant?

docs/gathered-data.md
```diff
@@ -399,6 +399,17 @@ Relevant OpenShift API docs:
 Location in archive: config/securitycontentconstraint/, config/clusterrolebinding/
 
 
+## SAPLicenseManagementLogs
+
+collects logs from license management pods with the following substrings:
```
tremes, line 404: Please mention SAP here as well...I think it should be as explicit as possible in case of these SAP enhancements.

docs/gathered-data.md
```diff
@@ -399,6 +399,17 @@ Relevant OpenShift API docs:
...
+  - "can't initialize iptables table",
+
+The Kubernetes API https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/pod_expansion.go#L48
+Response see https://docs.openshift.com/container-platform/4.6/rest_api/workloads_apis/pod-core-v1.html#apiv1namespacesnamespacepodsnamelog
+
+Location in archive: config/pod/sdi/logs/{pod-name}/errors.log
```
tremes, line 410: Maybe `{namespace}` instead of `sdi` ?

pkg/gather/clusterconfig/gather_logs.go
```diff
@@ -16,26 +16,50 @@ import (
...
-	regexSearch bool,
 ) ([]record.Record, error) {
+	var (
+		namespace                = containersFilter.namespace
+		labelSelector            = containersFilter.labelSelector
+		containerNameRegexFilter = containersFilter.containerNameRegexFilter
+		messagesToSearch         = logMessagesFilter.messagesToSearch
+		regexSearch              = logMessagesFilter.regexSearch
+		sinceSeconds             = logMessagesFilter.sinceSeconds
+		limitBytes               = logMessagesFilter.limitBytes
```
tremes, line 60: Is this blocked really needed here? Maybe I would use some shorter name for the local variables and then use them directly.

natiiix (commented):
I'm not giving it an approval right now because I don't want to accidentally trigger the auto-merge. The code is fine now. I have no more serious issues with it.

docs/gathered-data.md
Sergey1011010: fixed

docs/gathered-data.md
Sergey1011010: fixed

pkg/gather/clusterconfig/gather_logs.go
Sergey1011010: fixed

pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -84,6 +84,7 @@ var gatherFunctions = map[string]gathering{
 	"openshift_authentication_logs":     failable(GatherOpenshiftAuthenticationLogs),
 	"sap_config":                        failable(GatherSAPConfig),
 	"olm_operators":                     failable(GatherOLMOperators),
+	"sap_license_management_logs":       failable(GatherSAPLicenseManagementLogs),
```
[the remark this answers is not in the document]
Sergey1011010: why not, put them closer 
image

pkg/gather/clusterconfig/sap_license_management_logs.go
Sergey1011010: Yeah, the function makes sense. Do you think we should put it in this PR or maybe in some refactoring later?

tremes:
It would be nice to have something in the sample archive. Is there some problem with reproducing the error? Maybe Michal can help us? Maybe he has some example log?

pkg/gather/clusterconfig/constants.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	datahubGroupVersionResource = schema.GroupVersionResource{
+		Group: "installers.datahub.sap.com", Version: "v1alpha1", Resource: "datahubs",
+	}
+)
```
tremes, line 11: I think we already have `const.go` after last Marcell commit. Please put it there.

tremes:
One last comment from me, but otherwise LGTM. Thanks!!!!

pkg/gather/clusterconfig/sap_license_management_logs.go
natiiix: @Sergey1011010 Either way is fine with me, but I'd say it may be easier to just refactor it later for backporting consistency.

Sergey1011010:
@tremes I wasn't able to reproduce the error, but I guess the string would be "iptables v1.6.2: can't initialize iptables table `nat': Permission denied"

tremes:
I verified only basic run on cluster without SAP, but the changes looks reasonable. 
/lgtm

PR #345, opened 2021-02-16
Small clean up and utils reorg

0sewa0:

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- not necessary

## Documentation

- not necessary

## Unit Tests

- not necessary

## Privacy


pkg/gather/clusterconfig/0_gatherer_test.go
```diff
@@ -17,14 +18,14 @@ func (e *testError) Error() string {
 func mockGatherFunction1(g *Gatherer, c chan<- gatherResult) {
 	c <- gatherResult{[]record.Record{{
 		Name: "config/mock1",
-		Item: Raw{"mock1"},
+		Item: marshal.Raw{Str: "mock1"},
```
rluders, line 21: Isn't this one duplicated with [this](https://github.com/openshift/insights-operator/blob/dcca69ec2d24e9efa045586253b5354049e96ee3/tests/reports.go)? Is this marshal only used for testing? If the answer is YES, I'll recommend moving it to the `/tests/` packages. Same for everything else that is only used as a testing toolkit.

rluders (changes requested):

pkg/gather/clusterconfig/0_gatherer_test.go
0sewa0: its not only used for testing,
 its used in `pkg/gather/clusterconfig/gather_logs.go`, `pkg/gather/clusterconfig/operators.go` and pkg/gather/clusterconfig/version.go

rluders: Gotcha.

rluders:
Tested locally.
/lgtm

PR #346, opened 2021-02-16
Reduce binary file size

rluders:
This PR reduces the binary file by ~28% stripping the DWARF tables needed for debuggers. Stack traces and panics still readable and valid. It is NOT a compression, so, should not cause any overhead.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

No.

## Documentation

No.

## Unit Tests

No.

## Privacy

Not applicable.

## Changelog

No.

## References

Not applicable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 GOCMD ?= go
 GORUN ?= $(GOCMD) run 
 GOBUILD ?= $(GOCMD) build 
-GOBUILDFLAGS ?= -mod=vendor -ldflags "-X k8s.io/client-go/pkg/version.gitCommit=$$(git rev-parse HEAD) -X k8s.io/client-go/pkg/version.gitVersion=v1.0.0+$$(git rev-parse --short=7 HEAD)"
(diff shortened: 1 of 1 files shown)


tremes:
Why do we need a smaller binary? I can't see any advantage TBH and when looking quickly at the flags https://stackoverflow.com/questions/22267189/what-does-the-w-flag-mean-when-passed-in-via-the-ldflags-option-to-the-go-comman it seems it will disable pprof.

rluders:
@tremes are we profiling on production as well?

tremes:
@rluders We don't as I mentioned today. My point was that now it's pretty easy to turn on the profiling (and I guess it was added on purpose). After this change it means one needs to build a new image to be able to turn it on AFAIK. I think I would prefer the original behavior when comparing it to just smaller binary.

smarterclayton:
We have actually asked that people stop stripping binaries now, because go versions have made the cost cheaper.

PR #351, opened 2021-02-19
Reduce Gatherer's code complexity

rluders:
This PR reduces the `0_gatherer.go` code complexity.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

No.

## Documentation

No.

## Unit Tests

No.

## Privacy

Not applicable.

## Changelog

No.

## References

Not applicable.

### DIFF
--- a/pkg/gather/clusterconfig/0_gatherer.go
+++ b/pkg/gather/clusterconfig/0_gatherer.go
@@ -96,6 +96,27 @@ func New(gatherKubeConfig *rest.Config, gatherProtoKubeConfig *rest.Config, metr
 	}
 }
 
+// GatherInfo from reflection
+type GatherInfo struct {
+	name     string
+	result   gatherResult
+	function gatherFunction
+	canFail  bool
+	rvString string
+}
+
+// NewGatherInfo that holds reflection information
+func NewGatherInfo(gather string, rv reflect.Value) *GatherInfo {
+	gatherFunc := gatherFunctions[gather].function
+	return &GatherInfo{
+		name:     runtime.FuncForPC(reflect.ValueOf(gatherFunc).Pointer()).Name(),
+		result:   rv.Interface().(gatherResult),
+		function: gatherFunc,
+		canFail:  gatherFunctions[gather].canFail,
+		rvString: rv.String(),
+	}
+}
+
 // Gather is hosting and calling all the recording functions
 func (g *Gatherer) Gather(ctx context.Context, gatherList []string, recorder recorder.Interface) error {
 	g.ctx = ctx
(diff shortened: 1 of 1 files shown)


pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -160,6 +142,38 @@ func (g *Gatherer) Gather(ctx context.Context, gatherList []string, recorder rec
...
+
+	if gatherCanFail {
+		for _, err := range gatherResults.errors {
+			klog.V(5).Infof("Couldn't gather %s' received following error: %s\n", gatherName, err.Error())
+		}
+	} else {
+		errorsReport = extractErrors(gatherResults.errors)
+	}
+
+	for _, record := range gatherResults.records {
```
0sewa0, line 166: wouldn't it makes sense to have the "recording" part in a separate function or just rename the function.
because the name `gatherStatusReport` is a bit misleading, this functions can be better described as `handleResult` or something. Also it doesn't uses the `g` receiver anywhere so I just wouldn't make it a method.

rluders: I think that splitting in an even smaller function could be too much, but I agree with the rename. How about renaming it to `handleStatusReport`?

0sewa0: It creates the `statusReport` and stores the `gatherResult` neither of which is described by  `handleStatusReport` :sweat_smile: 
I would go with `handleGatherResult` and have the `reflect.Value` be already be converted to `gatherResult` before calling it.
Something like: 
```
func handleGatherResult(gather string, gr gatherResult, recorder recorder.Interface, starts time.Time) (gatherStatusReport, []string) {

```

rluders: I mean, fine with the name change, and it does make sense, but as you told, it creates the report. Probably would be better to call it `createGatherReport` instead. 

If you check the function response is that what it is returning. I'm not sure about passing the `gatherResult`, otherwise, you will have to put some other parts back to that other function not actually reducing the code complexity.

For example, the `gatherCanFail` variable is something that you will have to pass as a parameter as well, then you will have 5 parameters in the function call, which is way too much (I mean, IMHO 4 is already more than it should have. Code clean says that <= 3 is fine.).

But I can see where are you going. Maybe there is some space to improve it, let me try something.

rluders: @0sewa0 hopefully, I may have improved it.

0sewa0: much improved, thx :smile:

PR #352, opened 2021-02-25
Do not use context in the recorder

tremes:
Previously when the context exceeded then we end up with an empty archive. 

I believe we don't need the context when saving records to the disk. IOW I think we want to store the records everytime regardless of the context. 
## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

No new data

## Documentation

No update

## Unit Tests

No new test.

## Privacy

## References

https://issues.redhat.com/browse/CCXDEV-4118


rluders:
Yep, we don't need it. Nice simplification.
/lgtm

PR #354, opened 2021-02-27
Obfuscate IPv4 addresses and hide cluster base domain

Sergey1011010:
This PR implements anonymization of cluster base domain and IPv4 addresses

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

- `insights-operator/gathering.json` contains option `is_global_obfuscation_enabled` which is set to `true` when  obfuscation is enabled
- cluster base domains and IPs are changed to obfuscated values

## Unit Tests

- `pkg/anonymization/anonymizer_test.go`

## References

https://issues.redhat.com/browse/CCXDEV-3655


pkg/config/config.go
```diff
@@ -101,6 +106,12 @@ type Controller struct {
 	ReportPullingTimeout time.Duration
 	Impersonate          string
 	Gather               []string
+	// DisabledGlobalAnonymizations specifies which of global anonymizations to disable.
+	// By default, we anonymize everything.
+	// To see the detailed info about what we anonymize, go to the docs of package anonymization.
+	DisabledGlobalAnonymizations struct {
+		DisableClusterBaseDomainAnonymization bool
```
tremes, line 113: I guess this should be the opposite. We don't anonymize/obfusscate the IPs and hostnames by default. So I would call it e.g `EnableGlobalObfuscation` or something like that. It would be good to somehow emphasize that this is "only" for IPs and hostnames.

pkg/config/config.go
```diff
@@ -17,20 +17,24 @@ type Serialized struct {
 		Timeout      string `json:"timeout"`
 		MinRetryTime string `json:"min_retry"`
 	} `json:"pull_report"`
-	Impersonate string   `json:"impersonate"`
-	Gather      []string `json:"gather"`
+	Impersonate                  string   `json:"impersonate"`
+	Gather                       []string `json:"gather"`
+	DisabledGlobalAnonymizations []string `json:"disabledGlobalAnonymizations"`
```
tremes, line 22: If I understand it correctly, this slice defines what types of anonymization we want. I don't think it's necessary. I would simply introduce one flag saying obfuscation enabled/disabled (covering all - IPs and hostnames) but let's double check.

pkg/anonymization/anonymizer.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	conf := anonymizer.configObserver.Config().DisabledGlobalAnonymizations
+
+	if !conf.DisableClusterBaseDomainAnonymization && len(anonymizer.clusterBaseDomain) != 0 {
+		memoryRecord.Data = bytes.ReplaceAll(
+			memoryRecord.Data,
+			[]byte(anonymizer.clusterBaseDomain),
+			[]byte(config.ClusterBaseDomainPlaceholder),
+		)
```
tremes, line 47: I really like the idea of doing this in one place. However I can see one limitation right now. The result of this is e.g:
`https://api-int.<CLUSTER_BASE_DOMAIN>:6443/api/v1/namespaces/openshift-kube-scheduler/configmaps/kube-scheduler?timeout=10s` 
I guess we should obfuscate the whole string, but again let's discuss and double check.

Sergey1011010: Well, I don't see any sensitive information in the namespace and configmap name.

Sergey1011010: Btw, it's not a limitation of this approach. If we do need to anonymize more that just a base domain(the whole domain or the whole URL) we can use a regular expression to find a potential match and then check if it contains cluster base domain and anonymize the whole thing.

pkg/anonymization/anonymizer.go
```diff
@@ -0,0 +1,162 @@
...
+			// the problem is that some strings can be incorrectly identified as ip v6
+			// we can try looking only for those which are wrapped by quotes or something
+			return originalIPBytes
+		}
+
+		for _, network := range anonymizer.networks {
+			if network.Contains(originalIP) {
+				// return the first matched network
+				return []byte(network.IP.String())
+			}
```
tremes, line 150: I guess you know it, but this won't match all the IPs in the archive  - e.g `hostIP` in `HostSubnet` definition and more...

Sergey1011010: Well, it will become 0.0.0.0

docs/insights-archive-sample/insights-operator/gathering-metadata.json
```diff
@@ -0,0 +1 @@
...on_in_ms":3336,"records_count":0,"errors":null},{"name":"clusterconfig.GatherNodes","duration_in_ms":3339,"records_count":6,"errors":null},{"name":"clusterconfig.GatherCRD","duration_in_ms":3342,"records_count":2,"errors":null},{"name":"clusterconfig.GatherConfigMaps","duration_in_ms":3347,"records_count":10,"errors":["configmaps \"cluster-monitoring-config\" not found"]},{"name":"clusterconfig.GatherOpenshiftAuthenticationLogs","duration_in_ms":3727,"records_count":0,"errors":null},{"name":"clusterconfig.GatherContainerImages","duration_in_ms":3877,"records_count":9,"errors":null},{"name":"clusterconfig.GatherOpenshiftSDNControllerLogs","duration_in_ms":3899,"records_count":0,"errors":null},{"name":"clusterconfig.GatherOpenshiftSDNLogs","duration_in_ms":5437,"records_count":0,"errors":null},{"name":"clusterconfig.GatherClusterOperators","duration_in_ms":7763,"records_count":31,"errors":null},{"name":"clusterconfig.GatherInstallPlans","duration_in_ms":13395,"records_count":1,"errors":null},{"name":"clusterconfig.GatherServiceAccounts","duration_in_ms":13805,"records_count":1,"errors":null}],"memory_alloc_bytes":11979440,"uptime_seconds":59.876,"is_global_obfuscation_enabled":false}
```
tremes, line 1: Please format

pkg/anonymization/anonymizer.go
```diff
@@ -0,0 +1,306 @@
...
+	return NewAnonymizer(configObserver, baseDomain, networks)
+}
+
+// IsObfuscationEnabled returns true if obfuscation(hiding IP and domain names) is enabled and false otherwise
+func (anonymizer *Anonymizer) IsObfuscationEnabled() bool {
+	if anonymizer.configObserver == nil {
+		return false
+	}
+
+	return anonymizer.configObserver.Config().EnableGlobalObfuscation
```
tremes, line 168: The config will be read only once during operator start right? I think it's OK for now, but we should think about the situation when user enables it and IO is running. Right now it will require IO restart to take some effect.

pkg/controller/operator.go
```diff
@@ -105,15 +104,23 @@ func (s *Operator) Run(ctx context.Context, controller *controllercmd.Controller
 	// the last sync time, if any was set
 	statusReporter := status.NewController(configClient, gatherKubeClient.CoreV1(), configObserver, os.Getenv("POD_NAMESPACE"))
 
+	// anonymizer is responsible for anonymizing sensitive data, it can be configured to disable specific anonymization
+	anonymizer, err := anonymization.NewAnonymizerFromConfigClient(ctx, configObserver, kubeClient, configClient)
+	if err != nil {
+		return err
```
tremes, line 110: Hmm this can lead to failing IO (e.g it won't be able to read the `cluster-config-v1` CM -> fails horribly) even when the obfuscation is disabled right? I think we want to avoid such a situation. I guess we can check the new flag right here and create the anonymizer only when it's enabled. WDYT?

pkg/config/config.go
```diff
@@ -58,11 +62,13 @@ func ToController(s *Serialized, cfg *Controller) (*Controller, error) {
 	if cfg == nil {
 		cfg = &Controller{}
 	}
+
 	cfg.Report = s.Report
 	cfg.StoragePath = s.StoragePath
 	cfg.Endpoint = s.Endpoint
 	cfg.Impersonate = s.Impersonate
 	cfg.Gather = s.Gather
+	cfg.EnableGlobalObfuscation = cfg.EnableGlobalObfuscation || s.EnableGlobalObfuscation
```
tremes, line 71: I guess `cfg.EnableGlobalObfuscation` will be always false so probably not necessary to add it here.

tremes:
It would also be good to check cluster proxy. I guess the data won't be obfuscated when there's some cluster proxy configured. The question is if it should be obfuscated.... I would say yes.

natiiix (commented):
Considering the size, I didn't really investigate it in detail to look for any micro-optimizations, but as a general idea I think it looks alright and I believe it could work fine as is. I didn't see anything exactly wrong with any of the changes as I went through them.
I agree with many of the comments from Tomáš. He made some good points there. I don't think it's a problem to only obfuscate the hostname and keep the rest of the URL in most cases.

pkg/controller/operator.go
0sewa0: +1

pkg/anonymization/anonymizer.go
Sergey1011010: I thought config observer would observe the changes, but it works somehow differently. Considering the comment below we can keep this behaviour and not create the anonymizer at all if the feature is disabled.

pkg/config/config.go
Sergey1011010: Does it work now the way that it would be enabled if enableGlobalObfuscation is true in either the config or the secret?

tremes: The config file is a part of the image (and we don't expect users building their custom IO images) so this is useful only for dev purposes IMO.

pkg/anonymization/anonymizer.go
tremes: You need to "listen" on channel from `anonymizer.configObserver.ConfigChanged()` AFAIK.

pkg/config/config.go
Sergey1011010: Yeah, it doesn't make sense here, we only need it in merge config to be able to enable obfuscation by secret or by config.

tremes:
@Sergey1011010 I guess `retest` won't help you. Please check https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_insights-operator/354/pull-ci-openshift-insights-operator-master-insights-operator-e2e-tests/1379402907039830016. Looks like we will need to update the failing test.

Sergey1011010:
@tremes, ah, yeah, there are also integration tests. So we have two options: change the test to look for `gathering-metadata.json` instead of `gathers.json`, or to keep `gathers.json`. Maybe the second option is better if something processing the archive depends on `gathers.json` name. WDYT?

tremes:
@Sergey1011010 I like the new name more TBH. I hope that it's not so critical from the processing perspective (though it would be good to check). I am thinking about older versions. I mean it would be good if the file name was the same everywhere. That means we would need to backport the renaming change.

pkg/controller/operator.go
```diff
@@ -105,15 +104,26 @@ func (s *Operator) Run(ctx context.Context, controller *controllercmd.Controller
 	// the last sync time, if any was set
 	statusReporter := status.NewController(configClient, gatherKubeClient.CoreV1(), configObserver, os.Getenv("POD_NAMESPACE"))
 
+	var anonymizer *anonymization.Anonymizer
+	if anonymization.IsObfuscationEnabled(configObserver) {
+		// anonymizer is responsible for anonymizing sensitive data, it can be configured to disable specific anonymization
+		anonymizer, err = anonymization.NewAnonymizerFromConfigClient(ctx, kubeClient, configClient)
+		if err != nil {
+			return err
```
tremes, line 112: I am still not sure if we should fail or just log an error. This still can theoretically lead to crash loop.

Sergey1011010: Well, if obfuscation is disabled `anonymization.IsObfuscationEnabled` will return false and we won't reach here, but we can also just log an error in case it's enabled.

tremes: Yeah sure, but when it's enabled and some of the required resources (e.g DNS) is not available -> crash & problem :)

Sergey1011010: changed to logging an error

pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -236,14 +260,12 @@ func (g *Gatherer) startGathering(gatherList []string, errors *[]string) ([]refl
...
 	return cases, starts, nil
 }
 
-func recordGatherReport(recorder recorder.Interface, report []gatherStatusReport) error {
-	var m runtime.MemStats
-	runtime.ReadMemStats(&m)
-	metadata := gatherMetadata{report, m.HeapAlloc, time.Since(startTime).Truncate(time.Millisecond).Seconds()}
-	r := record.Record{Name: "insights-operator/gathers", Item: record.JSONMarshaller{Object: metadata}}
+func recordGatherReport(recorder recorder.Interface, metadata gatherMetadata) error {
+	r := record.Record{Name: "insights-operator/gathering-metadata", Item: record.JSONMarshaller{Object: metadata}}
```
tremes, line 268: I think you forgot to keep the original name of the file.

pkg/utils/strings_test.go
```diff
@@ -0,0 +1,35 @@
+package utils
+
+import (
+	"reflect"
+	"testing"
+)
+
+func Test_Gather_UniqueStrings(t *testing.T) {
```
tremes, line 8: I guess `Test_UniqueStrings` is sufficient.

pkg/anonymization/anonymizer_test.go
```diff
@@ -0,0 +1,201 @@
...
+func TestAnonymizer(t *testing.T) {
+	anonymizer := getAnonymizer(t)
+
+	type testCase struct {
+		before string
+		after  string
+	}
+
+	nameTestCases := []testCase{
+		{"node1.example.com", "node1.<CLUSTER_BASE_DOMAIN>"}, // TODO:
```
tremes, line 135: TODO ? :) More cases?

pkg/anonymization/anonymizer_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/openshift/insights-operator/pkg/config"
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+func TestGetNextIP(t *testing.T) {
```
tremes, line 14: I am not sure, but @rluders suggested `Test_GetNextIP` name format or am I wrong?

rluders: Correct.

pkg/anonymization/anonymizer_test.go
Sergey1011010: Hm, don't remember what I meant, probably I just forgot to remove some reminder for myself :)

pkg/anonymization/anonymizer.go
```diff
@@ -0,0 +1,308 @@
...
+	// so it's fine to just iterate over them
+	for i := range anonymizer.networks {
+		networkInfo := &anonymizer.networks[i]
+		network := &networkInfo.network
+
+		if network.IP.Equal(originalIP) {
+			return originalIP.String()
+		}
+
+		if network.Contains(originalIP) {
```
tremes, line 224: If I understand it correctly then we silently skip the IP if this is not true. Is that OK? I think I am missing the context.

Sergey1011010: Nope, it means we haven't found a network for this IP and we will return 0.0.0.0 after for loop

pkg/anonymization/anonymizer_test.go
tremes: OK np

pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -95,15 +104,19 @@ var gatherFunctions = map[string]gathering{
 	"olm_operators":                     failable(GatherOLMOperators),
 }
 
-var startTime time.Time
-
 // New creates new Gatherer
-func New(gatherKubeConfig *rest.Config, gatherProtoKubeConfig *rest.Config, metricsGatherKubeConfig *rest.Config) *Gatherer {
-	startTime = time.Now()
+func New(
```
rluders, line 108: It is just a suggestion, but you can simplify this one by using:

```
func New(gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config, anonymizer *anonymization.Anonymizer)
```

pkg/anonymization/anonymizer.go
tremes: Got it. First I didn't realize it can be found in other network. Thanks

pkg/anonymization/anonymizer.go
```diff
@@ -0,0 +1,308 @@
...
+
+		network1 := strings.Split(networks[i], "/")
+		network2 := strings.Split(networks[j], "/")
+
+		// subnet length
+		if network1[1] > network2[1] {
+			return true
+		} else if network1[1] < network2[1] {
+			return false
+		} else {
```
rluders, line 155: This else statement is not necessary.

rluders (changes requested):

pkg/anonymization/anonymizer.go
Sergey1011010: Why? It is. If the length is equal, we use subnet

pkg/gather/clusterconfig/0_gatherer.go
Sergey1011010: done

pkg/anonymization/anonymizer.go
rluders: If it doesn´t match if previous if and if-else it will always gonna be watched by the else... right?

So, you can simplify it by removing the else, and keeping the return. Something like:

```
// subnet length
if network1[1] > network2[1] {
    return true
} else if network1[1] < network2[1] {
    return false
}
return network1[0] > network2[0]
```

rluders (changes requested):

pkg/anonymization/anonymizer.go
rluders: If it doesn´t match if previous if and if-else it will always gonna be watched by the else... right?

So, you can simplify it by removing the else, and keeping the return. Something like:

```go
// subnet length
if network1[1] > network2[1] {
    return true
} else if network1[1] < network2[1] {
    return false
}
return network1[0] > network2[0]
```

tremes: Hmm and what about:
```
if network1[1] != network2[1] {
   return network1[1] > network2[1]
}
return network1[0] > network2[0]
```

rluders: @tremes even better. I think so.

Sergey1011010: done

PR #355, opened 2021-03-01
Bug 1924663: Gather related pod logs when a cluster operator is degraded

rluders:
This PR changes the ClusterOperator gather in order to allow it to collect also logs from pods that are related to unhealthy operators.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

No.

## Documentation

- `docs/gathered-data.md` _(updated Operators gather description)_

## Unit Tests

- `pkg/gather/clusterconfig/operators_test.go` _(adds extra testing to it)_

## Privacy

## Changelog

No.

## References

https://issues.redhat.com/browse/CCXDEV-4033
https://bugzilla.redhat.com/show_bug.cgi?id=1924663

### DIFF
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -130,8 +130,12 @@ Id in config: oauths
 
 ## ClusterOperators
 
-collects all ClusterOperators and their resources.
-It finds unhealthy Pods for unhealthy operators
+collects all the ClusterOperators definitions and their resources.
+It finds degraded Pods for degraded cluster operators or that lives at the Cluster Operator's namespace, to collect:
+
+- Pod definitions
+- Previous and current Pod Container logs (when available)
+- Namespace Events
 
 The Kubernetes api https://github.com/openshift/client-go/blob/master/config/clientset/versioned/typed/config/v1/clusteroperator.go#L62
 Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#clusteroperatorlist-v1config-openshift-io
--- a/pkg/gather/clusterconfig/operators.go
+++ b/pkg/gather/clusterconfig/operators.go
@@ -28,9 +28,9 @@ import (
 	"github.com/openshift/insights-operator/pkg/record"
 	"github.com/openshift/insights-operator/pkg/recorder"
 	"github.com/openshift/insights-operator/pkg/utils"
+	"github.com/openshift/insights-operator/pkg/utils/anonymize"
 	"github.com/openshift/insights-operator/pkg/utils/check"
 	"github.com/openshift/insights-operator/pkg/utils/marshal"
-	"github.com/openshift/insights-operator/pkg/utils/anonymize"
 )
 
 const (
@@ -60,8 +60,12 @@ type CompactedEventList struct {
 	Items []CompactedEvent `json:"items"`
 }
 
-// GatherClusterOperators collects all ClusterOperators and their resources.
-// It finds unhealthy Pods for unhealthy operators
+// GatherClusterOperators collects all the ClusterOperators definitions and their resources.
+// It finds degraded Pods for degraded cluster operators or that lives at the Cluster Operator's namespace, to collect:
+//
+// - Pod definitions
+// - Previous and current Pod Container logs (when available)
+// - Namespace Events
 //
 // The Kubernetes api https://github.com/openshift/client-go/blob/master/config/clientset/versioned/typed/config/v1/clusteroperator.go#L62
 // Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#clusteroperatorlist-v1config-openshift-io
@@ -92,28 +96,124 @@ func GatherClusterOperators(g *Gatherer, c chan<- gatherResult) {
 		c <- gatherResult{nil, []error{err}}
 		return
 	}
-	records, errors := gatherClusterOperators(g.ctx, gatherConfigClient, gatherKubeClient.CoreV1(), discoveryClient, dynamicClient)
-	c <- gatherResult{records, errors}
+	records, errs := gatherClusterOperators(g.ctx, gatherConfigClient, gatherKubeClient.CoreV1(), discoveryClient, dynamicClient)
+	if errs != nil {
+		c <- gatherResult{records, []error{errs}}
+		return
+	}
+	c <- gatherResult{records, []error{}}
 }
 
-func gatherClusterOperators(ctx context.Context, configClient configv1client.ConfigV1Interface, coreClient corev1client.CoreV1Interface, discoveryClient discovery.DiscoveryInterface, dynamicClient dynamic.Interface) ([]record.Record, []error) {
+// gatherClusterOperators collects cluster operators
+func gatherClusterOperators(ctx context.Context, configClient configv1client.ConfigV1Interface, coreClient corev1client.CoreV1Interface, discoveryClient discovery.DiscoveryInterface, dynamicClient dynamic.Interface) ([]record.Record, error) {
 	config, err := configClient.ClusterOperators().List(ctx, metav1.ListOptions{})
 	if errors.IsNotFound(err) {
 		return nil, nil
 	}
 	if err != nil {
-		return nil, []error{err}
+		return nil, err
 	}
+
+	// collect the cluster operators reports
+	records := clusterOperatorsRecords(ctx, config.Items, dynamicClient, discoveryClient)
+
+	// gather pods from unhealthy operators
+	pods, recs, totalContainers := gatherUnhealthyClusterOperator(ctx, config.Items, coreClient)
+	records = append(records, recs...)
+	// Exit early if no pods found
+	klog.V(2).Infof("Found %d pods with %d containers", len(pods), totalContainers)
+	if len(pods) == 0 || totalContainers <= 0 {
+		return records, nil
+	}
+
+	// gather pods containers logs
+	bufferSize := int64(recorder.MaxLogSize * logCompressionRatio / totalContainers / 2)
+	clogs, err := gatherPodContainersLogs(ctx, coreClient, pods, bufferSize)
+	if err != nil {
+		klog.V(2).Infof("Unable to gather pod containers logs: %v", err)
+		return records, nil
+	}
+	if len(clogs) > 0 {
+		records = append(records, clogs...)
+	}
+
+	return records, nil
+}
+
+// gatherPodContainersLogs collect the pod current and previous containers logs
+func gatherPodContainersLogs(ctx context.Context, client corev1client.CoreV1Interface, pods []*corev1.Pod, bufferSize int64) ([]record.Record, error) {
+	if bufferSize <= 0 {
+		return nil, fmt.Errorf("invalid buffer size %d", bufferSize)
+	}
+
+	// Fetch a list of containers in pods and calculate a log size quota
+	// Total log size must not exceed maxLogsSize multiplied by logCompressionRatio
+	klog.V(2).Infof("Maximum buffer size: %v bytes", bufferSize)
+	buf := bytes.NewBuffer(make([]byte, 0, bufferSize))
+
+	// Fetch previous and current container logs
+	var records []record.Record
+	for _, isPrevious := range []bool{true, false} {
+		for _, pod := range pods {
+			clog := getContainerLogs(ctx, client, pod, isPrevious, buf, bufferSize)
+			if len(clog) > 0 {
+				records = append(records, clog...)
+			}
+		}
+	}
+
+	return records, nil
+}
+
+// getContainerLogs get previous and current log reports for pod containers using the k8s API response
+func getContainerLogs(ctx context.Context, client corev1client.CoreV1Interface, pod *corev1.Pod, isPrevious bool, buf *bytes.Buffer, bufferSize int64) []record.Record {
+	var records []record.Record
+
+	allContainers := append(pod.Spec.InitContainers, pod.Spec.Containers...)
+	for _, c := range allContainers {
+		// only grab previous log if the pod is restarted
+		if isPrevious && !isPodRestarted(pod) {
+			continue
+		}
+
+		logName := fmt.Sprintf("%s_current.log", c.Name)
+		if isPrevious {
+			logName = fmt.Sprintf("%s_previous.log", c.Name)
+		}
+
+		buf.Reset()
+		klog.V(2).Infof("Fetching logs for %s container %s pod in namespace %s (previous: %v).", c.Name, pod.Name, pod.Namespace, isPrevious)
+		// Fetch container logs and continue on error
+		err := fetchPodContainerLog(ctx, client, pod, buf, c.Name, isPrevious, &bufferSize)
+		if err != nil {
+			klog.V(2).Infof("Error: %q", err)
+			continue
+		}
+
+		// Do not record empty records
+		if buf.Len() == 0 {
+			continue
+		}
+		records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/logs/%s/%s", pod.Namespace, pod.Name, logName), Item: marshal.Raw{Str: buf.String()}})
+	}
+
+	return records
+}
+
+// clusterOperatorsRecords generates the cluster operator records
+func clusterOperatorsRecords(ctx context.Context, items []configv1.ClusterOperator, dynamicClient dynamic.Interface, discoveryClient discovery.DiscoveryInterface) []record.Record {
 	resVer, _ := getOperatorResourcesVersions(discoveryClient)
-	records := make([]record.Record, 0, len(config.Items))
-	for idx, co := range config.Items {
+	records := make([]record.Record, 0, len(items))
+
+	for _, co := range items {
 		records = append(records, record.Record{
-			Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[idx].Name),
-			Item: record.JSONMarshaller{Object: &config.Items[idx]},
+			Name: fmt.Sprintf("config/clusteroperator/%s", co.Name),
+			Item: record.JSONMarshaller{Object: &co},
 		})
 		if resVer == nil {
 			continue
 		}
+
 		relRes := collectClusterOperatorResources(ctx, dynamicClient, co, resVer)
 		for _, rr := range relRes {
 			// imageregistry resources (config, pruner) are gathered in image_registries.go, image_pruners.go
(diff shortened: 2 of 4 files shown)


pkg/gather/clusterconfig/operators.go
```diff
@@ -130,80 +182,66 @@ func gatherClusterOperators(ctx context.Context, configClient configv1client.Con
...
-	// Total log size must not exceed maxLogsSize multiplied by logCompressionRatio
-	klog.V(2).Infof("Found %d unhealthy pods", len(unhealthyPods))
-	totalUnhealthyContainers := 0
-	for _, pod := range unhealthyPods {
-		totalUnhealthyContainers += len(pod.Spec.InitContainers) + len(pod.Spec.Containers)
-	}
-	bufferSize := int64(recorder.MaxLogSize * logCompressionRatio / totalUnhealthyContainers / 2)
-	klog.V(2).Infof("Maximum buffer size: %v bytes", bufferSize)
-	buf := bytes.NewBuffer(make([]byte, 0, bufferSize))
+func checkUnhealtyPods(items []corev1.Pod, when time.Time) ([]*corev1.Pod, []record.Record, int) {
```
tremes, line 232: The name doesn't look accurate. The function is not checking anything AFAIK.

pkg/gather/clusterconfig/operators.go
```diff
@@ -130,80 +182,66 @@ func gatherClusterOperators(ctx context.Context, configClient configv1client.Con
 			})
 		}
 	}
+
+	return records
+}
+
+func checkClusterOperator(ctx context.Context, items []configv1.ClusterOperator, coreClient corev1client.CoreV1Interface) ([]*corev1.Pod, []record.Record, int) {
```
tremes, line 189: It would be good to add some comment to the function. Same below.

pkg/gather/clusterconfig/operators.go
tremes: maybe `pods` instead of `items` ?

tremes: and the second argument is not used

tremes:
This will also need an update of the documentation for the `GatherClusterOperators` function to describe it in more detail.

pkg/gather/clusterconfig/operators.go
```diff
@@ -104,16 +103,69 @@ func gatherClusterOperators(ctx context.Context, configClient configv1client.Con
...
+		for _, pod := range unhealthyPods {
+			allContainers := pod.Spec.InitContainers
+			allContainers = append(allContainers, pod.Spec.Containers...)
+			for _, c := range allContainers {
+				logName := fmt.Sprintf("%s_current.log", c.Name)
+				if isPrevious {
+					logName = fmt.Sprintf("%s_previous.log", c.Name)
+				}
+				buf.Reset()
+				klog.V(2).Infof("Fetching logs for %s container %s pod in namespace %s (previous: %v): %v", c.Name, pod.Name, pod.Namespace, isPrevious, err)
```
tremes, line 136: Passing the `err` doesn't seem to make sense.

pkg/gather/clusterconfig/operators.go
```diff
@@ -314,6 +352,10 @@ func collectContainerLogs(ctx context.Context, coreClient corev1client.CoreV1Int
 	readCloser, err := req.Stream(ctx)
 	if err != nil {
 		klog.V(2).Infof("Failed to fetch log for %s pod in namespace %s for failing operator %s (previous: %v): %q", pod.Name, pod.Namespace, containerName, isPrevious, err)
+		// it is fine to not have a previous log
+		if isPrevious {
+			return nil
+		}
```
tremes, line 358: I wouldn't do that. It can fail regardless `isPrevious` flag and it should always return the error in such case.

pkg/gather/clusterconfig/operators.go
```diff
@@ -104,16 +103,69 @@ func gatherClusterOperators(ctx context.Context, configClient configv1client.Con
...
+	// Fetch previous and current container logs
+	for _, isPrevious := range []bool{true, false} {
+		for _, pod := range unhealthyPods {
+			allContainers := pod.Spec.InitContainers
+			allContainers = append(allContainers, pod.Spec.Containers...)
+			for _, c := range allContainers {
+				logName := fmt.Sprintf("%s_current.log", c.Name)
+				if isPrevious {
+					logName = fmt.Sprintf("%s_previous.log", c.Name)
+				}
```
tremes, line 134: I think we are interested in the previous log only when the pod/container was unhealthy (restarted). So I think we should keep the original check. I think the logic should be following:
- if the CO is degraded then try to collect all the related current logs
- if the CO is degraded and you found some restarted pods then try to collect the previous logs

WDYT @vrutkovs  pls?

tremes: @rluders It looks better now, however the logic is still not correct IMO. Now it tries to read previous log for every degraded operator pod (if I am not mistaken). I don't think this is necessary. I think clusteroperator can be degraded even without any pod restart (is this assumption correct @vrutkovs pls?).

I still think we should check if the given pod is unhealthy and try to collect the previous log only in this case. And what about some test?

rluders: @tremes check the last changes. It should be fine, I guess. If this is the expected behavior let me know, then I'll cover it with tests. BTW, I already separate the code in some way that we can implement some gorotines later one.

pkg/gather/clusterconfig/operators.go
```diff
@@ -61,7 +61,7 @@ type CompactedEventList struct {
 }
 
 // GatherClusterOperators collects all ClusterOperators and their resources.
-// It finds unhealthy Pods for unhealthy operators
+// It finds Pods for unhealthy operators
```
tremes, line 64: This doesn't tell much IMO. It would be nice to describe it more.

pkg/gather/clusterconfig/operators.go
```diff
@@ -104,16 +105,114 @@ func gatherClusterOperators(ctx context.Context, configClient configv1client.Con
...
+			continue
+		}
+		records = append(records, record.Record{Name: fmt.Sprintf("config/pod/%s/logs/%s/%s", pod.Namespace, pod.Name, logName), Item: marshal.Raw{Str: buf.String()}})
+	}
+
+	return records
+}
+
+// clusterOperatorsReports generates the cluster operator reports
+func clusterOperatorsReports(ctx context.Context, items []configv1.ClusterOperator, dynamicClient dynamic.Interface, discoveryClient discovery.DiscoveryInterface) []record.Record {
```
tremes, line 203: `clusterOperatorsReports` ? Didn't you want to call it `clusterOperatorsRecords`?

pkg/gather/clusterconfig/operators.go
rluders: Check it now.

docs/gathered-data.md
```diff
@@ -130,8 +130,12 @@ Id in config: oauths
 
 ## ClusterOperators
 
-collects all ClusterOperators and their resources.
-It finds unhealthy Pods for unhealthy operators
+collects all the ClusterOperators definitions and their resources.
+It finds degraded Pods for degraded cluster operators or that lives at the Cluster Operator's namespace, to collect:
```
0sewa0, line 134: Some grammar question/suggestion:
It finds degraded Pods for degraded cluster operators or those that live at the Cluster Operator's namespace, to collect:

tremes: We just agreed with Ricardo that using `unhealthy` is probably better. Because the CO can be just unavailable...and yeah pod is not degraded :)

pkg/gather/clusterconfig/operators.go
```diff
@@ -104,16 +109,116 @@ func gatherClusterOperators(ctx context.Context, configClient configv1client.Con
...
 	resVer, _ := getOperatorResourcesVersions(discoveryClient)
-	records := make([]record.Record, 0, len(config.Items))
-	for idx, co := range config.Items {
+	records := make([]record.Record, 0, len(items))
+
+	for idx, co := range items {
 		records = append(records, record.Record{
-			Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[idx].Name),
-			Item: record.JSONMarshaller{Object: &config.Items[idx]},
+			Name: fmt.Sprintf("config/clusteroperator/%s", items[idx].Name),
```
natiiix, line 215: You could just use `co` instead of `items[idx]` here, right? But I get it for the sake of consistency with the line below.

pkg/gather/clusterconfig/operators.go
```diff
@@ -130,82 +235,74 @@ func gatherClusterOperators(ctx context.Context, configClient configv1client.Con
...
+	totalContainers := 0
+
+	for _, item := range items {
 		if isHealthyOperator(&item) {
 			continue
 		}
+
 		for _, namespace := range namespacesForOperator(&item) {
-			pods, err := coreClient.Pods(namespace).List(ctx, metav1.ListOptions{})
+			pdl, err := coreClient.Pods(namespace).List(ctx, metav1.ListOptions{})
```
natiiix, line 256: There was a discussion some time ago about the use of C-style abbreviations in Go and why they're discouraged. I'm not really a supporter of the 10 words long Java-style naming conventions, but I believe that name should be clearly understandable even by someone who isn't familiar with the API. Names like `pdl`, `pds`, `tc` don't seem very descriptive. I'm mentioning this because you seem interested in best Go practices and these kinds of names are not generally considered a great practice in Go.

natiiix (commented):
Other than the few nitpicks I mentioned in the comments, I don't see anything wrong, but I haven't tested it locally.

pkg/gather/clusterconfig/operators.go
rluders: True.

pkg/gather/clusterconfig/operators.go
rluders: I guess that it just depends on the size of the function body, and its context. 

If you have a function that has a huge body, it would be really hard to track down the variables without having a clean name. Also, in this case, it would be probably a good idea to split the function into multiple functions. But if you are using the variable in some small context, giving huge names is just extra verbosity to the code.

But, I agree, the names quoted above are not the most descriptive ones, and I'll change them.

pkg/gather/clusterconfig/operators.go
```diff
@@ -92,28 +96,120 @@ func GatherClusterOperators(g *Gatherer, c chan<- gatherResult) {
...
+	// Exit early if no pods found
+	if len(pods) == 0 {
+		return records, nil
+	}
+
+	// gather pods containers logs
+	clogs, err := gatherPodContainersLogs(ctx, coreClient, pods, totalContainers)
+	if err != nil {
+		return records, nil
+	}
```
tremes, line 128: This doesn't seem right.

pkg/gather/clusterconfig/operators.go
```diff
@@ -92,28 +96,120 @@ func GatherClusterOperators(g *Gatherer, c chan<- gatherResult) {
...
+	return records, nil
+}
+
+// gatherPodContainersLogs collect the pod current and previous containers logs
+func gatherPodContainersLogs(ctx context.Context, client corev1client.CoreV1Interface, pods []*corev1.Pod, total int) ([]record.Record, error) {
+	var records []record.Record
+
+	if total == 0 {
+		return nil, fmt.Errorf("Total can't be zero.")
+	}
```
tremes, line 142: Does it make sense to return an error in this case? I guess if `total` == 0 then everything is OK. In other words all the operators are healthy or not?

pkg/gather/clusterconfig/operators.go
```diff
@@ -92,28 +96,120 @@ func GatherClusterOperators(g *Gatherer, c chan<- gatherResult) {
...
+				records = append(records, clog...)
+			}
+		}
+	}
+
+	return records, nil
+}
+
+// collectPodContainerLogs generates the previous and current log reports for pod containers using the k8s API response
+func collectPodContainerLogs(ctx context.Context, client corev1client.CoreV1Interface, pod *corev1.Pod, isPrevious bool, buf *bytes.Buffer, bufferSize int64) []record.Record {
```
tremes, line 165: It seems to me little bit confusing having `collectPodContainerLogs` and `gatherPodContainersLogs` functions. The names are almost the same.

pkg/gather/clusterconfig/operators_test.go
```diff
@@ -41,5 +59,577 @@ func Test_Operators_Gather(t *testing.T) {
...
+		},
+		{
+			name: "container with previous log",
+			args: args{
+				ctx:           context.TODO(),
+				coreClient:    kubefake.NewSimpleClientset().CoreV1(),
+				pod:           &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "testPod"}},
+				buf:           bytes.NewBuffer(make([]byte, 0, bufferSize)),
+				containerName: "testContainer",
+				isPrevious:    false,
```
tremes, line 403: Shouldn't it be `true` ?

pkg/gather/clusterconfig/operators.go
rluders: If you have a `total` of zero, and you pass it to the `gatherPodContainersLogs` you gonna have an error. 'Cause at line 147 we gonna have a division of zero.

pkg/gather/clusterconfig/operators.go
```diff
@@ -327,6 +417,45 @@ func collectContainerLogs(ctx context.Context, coreClient corev1client.CoreV1Int
 	return nil
 }
 
+// isHealthyOperator checks if operator ins't degraded or unavailable
+func isHealthyOperator(operator *configv1.ClusterOperator) bool {
+	for _, condition := range operator.Status.Conditions {
+		switch {
```
0sewa0, line 423: wouldn't a simple `if` be sufficient here ?

pkg/gather/clusterconfig/operators_test.go
```diff
@@ -41,5 +59,577 @@ func Test_Operators_Gather(t *testing.T) {
...
+			if (err != nil) != tt.wantErr {
+				t.Errorf("getOperatorResourcesVersions() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("getOperatorResourcesVersions() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
```
tremes, line 363: I don't consider this type of tests (basically passing empty args and checking no error returned) very useful TBH, but OK... :)

pkg/gather/clusterconfig/operators.go
tremes: I know. The question is does it make sense to have errors in case of everything is fine in the gatherer? The error is ignored above (which shouldn't be) and in the opposite case it should be propagated and visible in our metadata, which doesn't make sense to me....but maybe I am wrong.

pkg/gather/clusterconfig/operators.go
0sewa0: it should return the `err` as well IMO

pkg/gather/clusterconfig/operators.go
0sewa0: +1 for ^ but also
I don't see the point of counting the failing containers because we only use it here, and if a pod is failing it probably has a failing container so we will get passed here furthermore we will get all the pod logs even the none failing ones if I understand correctly which is not desirable IMO

rluders: You both are just assuming things, that we never ever will have a `total` as zero. I prefer to make sure that at any point we will have a division by zero, it is just an extra check that has no impact on performance or code readability, in fact, it even helps you to understand that `total` should not be zero.

I totally disagree that this is pointless or unnecessary.

pkg/gather/clusterconfig/operators.go
rluders: Yeah, I think that it should return the `err` and not return the `records` as well... Right? But, in this case, we first need to finish the discussion down below.

pkg/gather/clusterconfig/operators.go
rluders: @tremes any suggestion?

pkg/gather/clusterconfig/operators.go
rluders: All right, that is true. Old code didn't check it, but... yeah, if.

pkg/gather/clusterconfig/operators_test.go
rluders: Yeh, me neither. I'll address the test improvements in another task, so we can get this one merged to unblock the other one.

pkg/gather/clusterconfig/operators.go
```diff
@@ -92,28 +96,120 @@ func GatherClusterOperators(g *Gatherer, c chan<- gatherResult) {
 		c <- gatherResult{nil, []error{err}}
 		return
 	}
-	records, errors := gatherClusterOperators(g.ctx, gatherConfigClient, gatherKubeClient.CoreV1(), discoveryClient, dynamicClient)
-	c <- gatherResult{records, errors}
+	records, errs := gatherClusterOperators(g.ctx, gatherConfigClient, gatherKubeClient.CoreV1(), discoveryClient, dynamicClient)
+	c <- gatherResult{records, []error{errs}}
```
tremes, line 100: This is wrong. We shouldn't send the `error` array when the `errs` is `nil`.

rluders: Nice catch. Thanks.

pkg/gather/clusterconfig/operators.go
tremes: I was wrong a bit. This code is never executed if everything is fine. It's obvious from https://github.com/openshift/insights-operator/pull/355/files#diff-d7d4c2ca2ff27679a6db98198f692ae0bf7f3c50355e525da4f327e7e25d401bR256 that `total` will be greater than 0 only when there are some Pods found so I think it's not necessary for `gatherPodContainersLogs` to return error as second return value. 

@0sewa0 number of failing pods != number of failing containers AFAIK. See https://github.com/openshift/insights-operator/pull/355/files#diff-d7d4c2ca2ff27679a6db98198f692ae0bf7f3c50355e525da4f327e7e25d401bR290. So the `total` probably makes sense when calculating the buffer size.

pkg/gather/clusterconfig/operators.go
rluders: BTW, now checking the previous/other returns on this exact same method. We always return the `records`... My thoughts would be... and if we continue returning the `records` but added a `log` event that logs the error when it occurred?

pkg/gather/clusterconfig/operators.go
tremes: Not sure. Either merge these two functions or distinguish the naming.

pkg/gather/clusterconfig/operators.go
rluders: I see your point, but... Basically, you are saying that everyone that uses the `gatherPodContainersLogs` should be aware that the `total` should never be zero. My tendency is to not trust people's awareness.

I don't see any harm to keep this control, it doesn't complicate anything and assure that, even if we change something, we never gonna have to implicitly know that `total` should never be zero. I prefer to keep the validation and make sure that it is explicit in the code.

And yes, the total is required 'cause it is used to calculate the buffer size, `len(pods)` and `total` are totally different things and values.

pkg/gather/clusterconfig/operators.go
rluders: If we merge those functions we gonna increase its code complexity. They are already splitter to reduce it. Let me find a better name so...

pkg/gather/clusterconfig/operators.go
tremes: OK I see your point too. I am thinking then it would be probably better to pass the `bufferSize` as an argument to the `gatherPodContainersLogs` instead of `total`. I think it would be more clear for the potential caller then.

rluders: I see your idea, and I liked it. :)

pkg/gather/clusterconfig/operators.go
```diff
@@ -92,28 +96,124 @@ func GatherClusterOperators(g *Gatherer, c chan<- gatherResult) {
...
 		return
 	}
-	records, errors := gatherClusterOperators(g.ctx, gatherConfigClient, gatherKubeClient.CoreV1(), discoveryClient, dynamicClient)
-	c <- gatherResult{records, errors}
+	records, errs := gatherClusterOperators(g.ctx, gatherConfigClient, gatherKubeClient.CoreV1(), discoveryClient, dynamicClient)
+	if errs != nil {
+		c <- gatherResult{records, []error{errs}}
+		return
+	}
+	c <- gatherResult{records, []error{}}
```
tremes, line 104: Can we send to the channel `gatherResult{records, nil}` pls?

pkg/gather/clusterconfig/operators.go
```diff
@@ -92,28 +96,124 @@ func GatherClusterOperators(g *Gatherer, c chan<- gatherResult) {
...
-	records := make([]record.Record, 0, len(config.Items))
-	for idx, co := range config.Items {
+	records := make([]record.Record, 0, len(items))
+
+	for _, co := range items {
 		records = append(records, record.Record{
-			Name: fmt.Sprintf("config/clusteroperator/%s", config.Items[idx].Name),
-			Item: record.JSONMarshaller{Object: &config.Items[idx]},
+			Name: fmt.Sprintf("config/clusteroperator/%s", co.Name),
+			Item: record.JSONMarshaller{Object: &co},
```
natiiix, line 211: Unless I'm mistaken, which is entirely possible, this will not work as expected. In this context, `co` is a copy of the structure, not a reference to it, so storing a pointer to the copy will result in records being a list of N instances of the last item in `items`. I can totally see how this mistake is easily made because I keep on making it myself and then wondering why and how stuff that shouldn't be gathered ends up in the archive. Only the line above should have been changed. For the object reference, it is necessary to use the old approach with an index.

natiiix (changes requested):

pkg/gather/clusterconfig/operators.go
tremes: Ivo is right and yes I made this mistake too. Please fix.

rluders: I wasn't aware of this. Nice catch, I'll change it right now. Thanks, @natiiix

natiiix (approved):
/lgtm
Tested locally against a testing cluster. IO ran without any issues for multiple gathering iterations. The directories used by this gatherer contained relevant files (pod definitions, logs, events, etc.) in their respective paths.
It would be great to add a list of files/directories that contain newly gathered files to make it easier to track down what was added without having to go through the code.

natiiix:
/lgtm
I messed up the command the first time. Yikes.

PR #356, opened 2021-03-02
Bug 1934112: Adds memory usage to the metadata

0sewa0:
This PR adds 2 new fields to the metadata (`gathers.json`).
`uptime`: the duration since the start of the first gather
`memory_alloc`: the memory allocated at the time of creating the report. (in bytes)

Some disclaimers:
- I intended to have memory usage stats for each gather-function, but it wasn't really viable. The only way I could measure it was by marshaling the produced records, which is not only inefficient but we can just as easily measure the size of the produced archive and get the same idea.
- The `memory_alloc` stat wont show the peak memory allocation. Even if we are only talking about the gather cycle, knowing when to measure to get the max usage is impossible so we would need to monitor the entire process, which would more complex without much benefit IMO. The purpose of this field is to be able to notice if over a long period there might be a memory leak.    

## Categories

- [ ] Bugfix
- [x] Enhancement
- [ ] Backporting

## Sample archive

- `docs/insights-archive-sample/insights-operator/gathers.json`

## Documentation

- none

## Unit Tests

- none

## Privacy

## References

none

### DIFF
--- a/docs/insights-archive-sample/insights-operator/gathers.json
+++ b/docs/insights-archive-sample/insights-operator/gathers.json
@@ -1,0 +1,214 @@
+{
+    "status_reports": [
+        {
+            "name": "github.com/openshift/insights-operator/pkg/gather/clusterconfig.GatherMostRecentMetrics",
+            "duration_in_ms": 23,
+            "records_count": 0,
+            "errors": [
+                "Get \"https://prometheus-k8s.openshift-monitoring.svc:9091/federate?match%5B%5D=etcd_object_counts\u0026match%5B%5D=cluster_installer\u0026match%5B%5D=namespace%3Acontainer_cpu_usage_seconds_total%3Asum_rate\u0026match%5B%5D=namespace%3Acontainer_memory_usage_bytes%3Asum\": dial tcp: lookup prometheus-k8s.openshift-monitoring.svc: no such host"
+            ]
+        },
+        {
+            "name": "github.com/openshift/insights-operator/pkg/gather/clusterconfig.GatherHostSubnet",
+            "duration_in_ms": 127,
+            "records_count": 6,
+            "errors": null
+        },
+        {
+            "name": "github.com/openshift/insights-operator/pkg/gather/clusterconfig.GatherClusterInfrastructure",
+            "duration_in_ms": 128,
+            "records_count": 1,
+            "errors": null
+        },
+        {
+            "name": "github.com/openshift/insights-operator/pkg/gather/clusterconfig.GatherSAPVsystemIptablesLogs",
+            "duration_in_ms": 240,
+            "records_count": 0,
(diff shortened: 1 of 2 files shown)


pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -16,6 +16,12 @@ import (
 	"github.com/openshift/insights-operator/pkg/recorder"
 )
 
+type gatherMetadata struct {
+	StatusReports []gatherStatusReport `json:"status_reports"`
+	MemoryAlloc   uint64               `json:"memory_alloc"`
+	Uptime     	  string        	   `json:"uptime"`
```
tremes, line 22: Formatting

docs/insights-archive-sample/insights-operator/gathers.json
```diff
@@ -0,0 +1,214 @@
...
+        },
+        {
+            "name": "github.com/openshift/insights-operator/pkg/gather/clusterconfig.GatherServiceAccounts",
+            "duration_in_ms": 11065,
+            "records_count": 1,
+            "errors": null
+        }
+    ],
+    "memory_alloc": 10924832,
+    "uptime": "54.024s"
```
tremes, line 213: Why string? :)

docs/insights-archive-sample/insights-operator/gathers.json
```diff
@@ -0,0 +1,214 @@
...
+            "errors": null
+        },
+        {
+            "name": "github.com/openshift/insights-operator/pkg/gather/clusterconfig.GatherServiceAccounts",
+            "duration_in_ms": 11065,
+            "records_count": 1,
+            "errors": null
+        }
+    ],
+    "memory_alloc": 10924832,
```
tremes, line 212: Maybe we should mention the unit again...but as you want.

docs/insights-archive-sample/insights-operator/gathers.json
```diff
@@ -0,0 +1,214 @@
...
+            "errors": null
+        },
+        {
+            "name": "github.com/openshift/insights-operator/pkg/gather/clusterconfig.GatherInstallPlans",
+            "duration_in_ms": 10445,
+            "records_count": 1,
+            "errors": null
+        },
+        {
+            "name": "github.com/openshift/insights-operator/pkg/gather/clusterconfig.GatherServiceAccounts",
```
tremes, line 206: I know this is totally unrelated to this PR but I really think the name is just  unnecessarily long (It proved again when working with the data in Jupyter notebooks). I find the prefix `github.com/openshift/insights-operator` useless.

tremes:
I tried it locally and the code change looks good to me as well. Maybe we could do the `uptime` in the minutes, but it's very minor detail. 
/lgtm

PR #361, opened 2021-03-04
Disconnected start command

0sewa0:
Duo to the fact that IO will not gather anything if its disabled(caused by repeated upload failures for disconnected customers), we would still like to give the users the option to run the IO in a mode when it doesn't send anything back, but just generates the archives, so the user can manually upload it if necessary.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- nope

## Documentation

- TODO

## Unit Tests

- nope

## Privacy

## References

https://issues.redhat.com/browse/CCXDEV-4133
(https://issues.redhat.com/browse/CCXDEV-4075)


0sewa0:
Currently its just a bunch of Copy-Paste from around the codebase, I want to minimize the code duplication.

pkg/controller/disconnected_operator.go
```diff
@@ -0,0 +1,137 @@
...
+		return fmt.Errorf("unable to load config: %v", err)
+	}
+
+	controller, err := cfg.ToDisconnectedController(&d.Controller)
+	if err != nil {
+		return err
+	}
+	d.Controller = *controller
+
+	data, _ := json.Marshal(cfg)
```
Sergey1011010, line 49: Are you sure the error will never happen?

0sewa0: good question, for now its mainly a copy paste from the `operator.go` in the most part. Will look into it

0sewa0: It shouldn't reach this point if something was wrong with `cfg` and also its only for printing it to the log. :)

pkg/controller/disconnected_operator.go
```diff
@@ -0,0 +1,119 @@
...
+		"clusterconfig": clusterConfigGatherer,
+	})
+	statusReporter.AddSources(periodic.Sources()...)
+
+	// check we can read IO container status and we are not in crash loop
+	err = wait.PollImmediate(20*time.Second, wait.Jitter(d.Controller.Interval/24, 0.1), isRunning(ctx, gatherKubeConfig))
+	if err != nil {
+		initialDelay = wait.Jitter(d.Controller.Interval/12, 0.5)
+		klog.Infof("Unable to check insights-operator pod status. Setting initial delay to %s", initialDelay)
+	}
```
tremes, line 112: I don't think we need this in case of "manual/disconnected" run. The initial dely should be always 0 in this case.

pkg/controller/disconnected_operator.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// the status controller initializes the cluster operator object and retrieves
+	// the last sync time, if any was set
+	statusReporter := status.NewController(configClient, gatherKubeClient.CoreV1(), configObserver, os.Getenv("POD_NAMESPACE"))
+
+	// the recorder periodically flushes any recorded data to disk as tar.gz files
+	// in s.StoragePath, and also prunes files above a certain age
+	recdriver := diskrecorder.New(d.StoragePath)
+	recorder := recorder.New(recdriver, d.Interval)
+	go recorder.PeriodicallyPrune(ctx, statusReporter)
```
tremes, line 97: Do we need this when triggering manually?

pkg/controller/disconnected_operator.go
```diff
@@ -0,0 +1,119 @@
...
+	// ensure the insight snapshot directory exists
+	if _, err := os.Stat(d.StoragePath); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(d.StoragePath, 0777); err != nil {
+			return fmt.Errorf("can't create --path: %v", err)
+		}
+	}
+
+	// configobserver synthesizes all config into the status reporter controller
+	configObserver := configobserver.New(d.Controller, kubeClient)
+	go configObserver.Start(ctx)
```
tremes, line 87: Do we need this when triggering manually? I think we can read the config once and don't need to periodically check. It's probably good time to start thinking about the way how user can use it and trigger it. If he/she creates a new deployment then we will need this.....but I think it would be easier to recommend it as one time job, but then there's the question where the archive would be stored

0sewa0:
closing this PR because there was a bit of misunderstanding, the gathering should only run once and store the results and not start a whole operator that does the gathering periodically, but not uploading it

PR #371, opened 2021-03-10
Introduce quick gather command

0sewa0:
Introduces a new `gather` command that will run a single gather without any uploading or anything.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- none

## Documentation

- Updated/created docs strings, the actual user documentation is a separate task

## Unit Tests

-none

## Privacy

### DIFF
--- a/pkg/cmd/start/start.go
+++ b/pkg/cmd/start/start.go
@@ -9,8 +9,11 @@ import (
 
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/serviceability"
+
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/pkg/version"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/insights-operator/pkg/config"
@@ -19,8 +22,12 @@ import (
 
 const serviceCACertPath = "/var/run/configmaps/service-ca-bundle/service-ca.crt"
 
+type operatorController interface {
+	Run(ctx context.Context, controller *controllercmd.ControllerContext) error
+}
+
 func NewOperator() *cobra.Command {
-	operator := &controller.Support{
+	operator := &controller.Operator{
 		Controller: config.Controller{
 			StoragePath:          "/var/lib/insights-operator",
 			Interval:             10 * time.Minute,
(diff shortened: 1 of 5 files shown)


pkg/cmd/start/start.go
```diff
@@ -19,8 +22,12 @@ import (
 
 const serviceCACertPath = "/var/run/configmaps/service-ca-bundle/service-ca.crt"
 
+type operatorController interface {
+	Run(ctx context.Context, controller *controllercmd.ControllerContext) error
+}
```
tremes, line 27: I think we don't need this interface. It can be `func runOperator(operator controller.Operator, cfg *controllercmd.ControllerCommandConfig)` or am I missing anything?

pkg/cmd/start/start.go
```diff
@@ -35,59 +42,126 @@ func NewOperator() *cobra.Command {
...
-				WithRestartOnChange(exitOnChangeReactorCh, startingFileContent, observedFiles...)
-
-			if err := builder.Run(ctx2, unstructured); err != nil {
-				klog.Fatal(err)
-			}
+func NewGather() *cobra.Command {
+	operator := &controller.GatherJob{
+		Controller: config.Controller{
+			StoragePath: "/var/lib/insights-operator",
+			Interval:    10 * time.Minute,
```
tremes, line 56: I would give it more time :)

pkg/cmd/start/start.go
```diff
@@ -35,59 +42,126 @@ func NewOperator() *cobra.Command {
...
+			Endpoint:    "local",
```
tremes, line 57: Can we just omit this?

pkg/cmd/start/start.go
```diff
@@ -35,59 +42,126 @@ func NewOperator() *cobra.Command {
...
+		}
+		clientConfig, err := kubeConfig.ClientConfig()
+		if err != nil {
+			klog.Fatal(err)
+		}
+		protoConfig := rest.CopyConfig(clientConfig)
+		protoConfig.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
+		protoConfig.ContentType = "application/vnd.kubernetes.protobuf"
+
+		ctx, cancel := context.WithTimeout(context.Background(), operator.Interval/2)
```
tremes, line 106: Maybe we can use the whole `operator.Interval` as the timeout or WDYT?

pkg/cmd/start/start.go
0sewa0: Just a leftover from a failed attempt :)

pkg/cmd/start/start.go
0sewa0: It used to be necessary because if it wasn't set then we would be 'Disabled', but no longer necessary

pkg/cmd/start/start.go
0sewa0: Its really up to what we set in config, so if we are going to have a separate config for this command it would definitely be more straight forward :+1:

Sergey1011010:
Just tested, it creates the archive. /lgtm

PR #374, opened 2021-03-12
Bug 1949907: Gather summary of PodNetworkConnectivityChecks

natiiix:
This PR implements a new data enhancement for gathering a summary of `PodNetworkConnectivityChecks`. Time of the most recently failed check with each reason and message is recorded in the IO archive. This enhancement is proactive and will not be backported.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

- `docs/insights-archive-sample/config/podnetworkconnectivitychecks.json`

## Documentation

- `docs/gathered-data.md` - "PNCC" section

## Unit Tests

- `pkg/gather/clusterconfig/pod_network_connectivity_checks_test.go` - `Test_PNCC`

## Privacy

:warning: Gathered data may contain external IP addresses and ports of services (i.e., when an important service is unreachable). :warning: 

## Changelog

No.

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-4192
Bugzilla Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1949907


pkg/gather/clusterconfig/pod_network_connectivity_checks.go
```diff
@@ -0,0 +1,93 @@
...
+			unsuccessful = append(unsuccessful, getUnsuccessfulChecks(outage.StartLogs)...)
+			unsuccessful = append(unsuccessful, getUnsuccessfulChecks(outage.EndLogs)...)
+		}
+	}
+
+	reasons := map[string]map[string]int{}
+	for _, entry := range unsuccessful {
+		if _, reasonExist := reasons[entry.Reason]; reasonExist {
+			msgMap := reasons[entry.Reason]
+			if _, msgExist := msgMap[entry.Message]; msgExist {
```
rluders, line 80: How about...

```go
if _, msgExist := msgMap[entry.Message]; msgExist {
    msgMap[entry.Message]++
    continue
}
msgMap[entry.Message] = 1
```

pkg/gather/clusterconfig/pod_network_connectivity_checks.go
```diff
@@ -0,0 +1,93 @@
...
+				msgMap[entry.Message]++
+			} else {
+				msgMap[entry.Message] = 1
+			}
+		} else {
```
rluders, line 85: I guess that you can use a similar approach here as well.

rluders (changes requested):

pkg/gather/clusterconfig/pod_network_connectivity_checks.go
natiiix: Isn't `continue` considered a bad practice when it isn't strictly necessary? I would say that the readability is also worse than with a regular if-else because this isn't really a "happy path" situation with an error check or something like that. Both of the branches are a part of normal functionality.

rluders: I don´t think so... when you put the `continue`, you mostly are saying: "there is nothing after this that you should worry about if it matches this condition"... IMO, it kind of reduces the cyclomatic complexity. My tendency is always to remove unnecessary `else` statements for that specific reason. If the `else` breaks the flow, better to use something that indicates it immediately.

natiiix: Statements like `continue` or `break` are practically equivalent to using `goto`, so I don't think they're ever the preferable way of solving a problem where two equally probably options exist. Go only suggests the omission of an `else` statement after a condition that either returns or terminates the runtime. To me this kind of abuse of flow control indicates some kind of a special case (we have a lot of code like this, where there is a `break`/`continue`/`return` in a condition), but it doesn't seem natural here. Using if-else clearly indicates that there are two options rather than one special condition (typically an error of some sort) and one normal/healthy path.

tremes: I don't see any issue with the original Ivo's definition TBH.

natiiix: Nevermind, I've figured out that in Go, this entire syntax is unnecessary because you can safely increment non-existing keys in a Go map. Apparently that's the idiomatic Go solution to this problem. [source](https://stackoverflow.com/a/41851720/3043260)

natiiix:
/retest
No idea why they're failing. The 503 status is suspicious and doesn't seem like an IO-related problem.

pkg/gather/clusterconfig/pod_network_connectivity_checks.go
```diff
@@ -0,0 +1,85 @@
...
+
+// GatherPNCC collects a summary of failed PodNetworkConnectivityChecks.
+// The checks are requested via a dynamic client and
+// then unmarshaled into the appropriate structure.
+//
+// Resource API: podnetworkconnectivitychecks.controlplane.operator.openshift.io/v1alpha1
+// Docs for relevant types: https://pkg.go.dev/github.com/openshift/api/operatorcontrolplane/v1alpha1
+//
+// * Location in archive: config/podnetworkconnectivitychecks.json
+// * Id in config: pod_network_connectivity_checks
```
tremes, line 24: You will need a bugzilla, but I am wondering if we can already add the version `4.8+`

docs/insights-archive-sample/config/podnetworkconnectivitychecks.json
```diff
@@ -0,0 +1,8 @@
+{
+  "TCPConnectError": {
+    "kubernetes-apiserver-endpoint-master-0: failed to establish a TCP connection to 10.74.181.101:6443: dial tcp 10.74.181.101:6443: connect: connection refused": 6,
+    "kubernetes-apiserver-service-cluster: failed to establish a TCP connection to 172.30.101.107:443: dial tcp 172.30.101.107:443: connect: connection refused": 6,
+    "openshift-apiserver-endpoint-master-0: failed to establish a TCP connection to 10.128.0.31:8443: dial tcp 10.128.0.31:8443: connect: connection refused": 3,
+    "openshift-apiserver-service-cluster: failed to establish a TCP connection to 172.30.137.129:443: dial tcp 172.30.137.129:443: i/o timeout": 3
```
tremes, line 6: Wouldn't be more interesting to provide the last time the given problem occurred (than to count the number of occurrence)? Just an idea...

pkg/gather/clusterconfig/pod_network_connectivity_checks.go
natiiix: I guess it's okay to add it there if we can make it in time for 4.8.0.

docs/insights-archive-sample/config/podnetworkconnectivitychecks.json
natiiix: At least according to the list of PNCC I got from the QuickLab cluster, it seems (and this is a very wild guess, but that's what the gathered JSON suggests) that these checks are performed in a quick succession and the results of the last ~10 checks are stored. In the file I have, there are 10 successes in each batch of checks and they are always 1 minute apart from each other. But it's true that the failures are much older than just a couple of minutes, so maybe it would be a good idea to add that information.
Perhaps the information from `failures` and `outages` should be separated as well. They seem to be related, but adding them together may cause some duplication because it seems like each message appears 3 times (failure message, outage start message, outage end message).

pkg/gather/clusterconfig/pod_network_connectivity_checks_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	recItem, ok := rec.Item.(record.JSONMarshaller)
+	if !ok {
+		t.Fatalf("unexpected type of record item in the first run: %q", rec.Name)
+	}
+	if !reflect.DeepEqual(recItem.Object, map[string]map[string]time.Time{}) {
+		t.Fatalf("unexpected value of record item in the first run: %#v", recItem)
+	}
+
+	// Create the DataHubs resource and now the SCCs and CRBs should be gathered.
```
tremes, line 66: Some copy/paste comment?

pkg/gather/clusterconfig/pod_network_connectivity_checks.go
```diff
@@ -0,0 +1,90 @@
...
+	unsuccesseful := []controlplanev1.LogEntry{}
+	for _, entry := range entries {
+		if !entry.Success {
+			unsuccesseful = append(unsuccesseful, entry)
+		}
+	}
+	return unsuccesseful
+}
+
+func gatherPNCC(ctx context.Context, dynamicClient dynamic.Interface, coreClient corev1client.CoreV1Interface) ([]record.Record, []error) {
```
tremes, line 54: Linter told me that `coreClient` was not used :)

pkg/gather/clusterconfig/pod_network_connectivity_checks_test.go
natiiix: Oops, nice catch, I fixed the comments in the implementation, but didn't look closely enough at the tests.

pkg/gather/clusterconfig/pod_network_connectivity_checks.go
natiiix: VSC betrayed me and didn't notice that. I'm surprised it even compiled with an unreferenced variable.

docs/gathered-data.md
```diff
@@ -421,6 +421,19 @@ Response see https://docs.openshift.com/container-platform/4.6/rest_api/workload
   * 4.7+
 
 
+## PNCC
+
+collects a summary of failed PodNetworkConnectivityChecks.
+The checks are requested via a dynamic client and
+then unmarshaled into the appropriate structure.
```
tremes, line 428: It would be good to mention that only failed checks or `Failures` are collected.

docs/insights-archive-sample/config/podnetworkconnectivitychecks.json
natiiix: I'm on it! I always forget the sample archive exists when making these changes.

docs/gathered-data.md
natiiix: @tremes Well, doesn't it already say this?
> collects a summary of **failed** PodNetworkConnectivityChecks.

tremes: Ahh OK I am blind. Ignore me :)

tremes:
I went through the code and ran it locally. Thanks!
/lgtm

natiiix:
:tada: That took an awful lot of attempts to pass the CI. :upside_down_face:

PR #375, opened 2021-03-16
Bug 1988351: OCM controller - periodically pull the data and update corresponding

tremes:
secret

This adds a new OCM controller which periodically (each 8 hours) tries to pull the SCA (Simple content access) data from the OCM API. The SCA data (basically a x509 certificate) is exposed in the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace. 

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

No new data

## Documentation

No update here

## Unit Tests

Basic test covered in `pkg/ocm/ocm_test.go ` 

## Privacy

## References

https://issues.redhat.com/browse/CCXDEV-4211

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,6 +24,10 @@ type Serialized struct {
 	Impersonate             string   `json:"impersonate"`
 	Gather                  []string `json:"gather"`
 	EnableGlobalObfuscation bool     `json:"enableGlobalObfuscation"`
+	Ocm         struct {
+		Endpoint string `json:"endpoint"`
+		Interval string `json:"interval"`
+	}
 }
 
 // Controller defines the standard config for this operator.
@@ -50,6 +54,8 @@ type Controller struct {
 	// EnableGlobalObfuscation enables obfuscation of domain names and IP addresses
 	// To see the detailed info about how anonymization works, go to the docs of package anonymization.
 	EnableGlobalObfuscation bool
+	OcmEndpoint          string
+	OcmInterval          time.Duration
 
 	Username string
 	Password string
@@ -159,6 +165,17 @@ func ToController(s *Serialized, cfg *Controller) (*Controller, error) { // noli
 		return nil, fmt.Errorf("storagePath must point to a directory where snapshots can be stored")
 	}
 
+	if len(s.Ocm.Endpoint) > 0 {
+		cfg.OcmEndpoint = s.Ocm.Endpoint
+	}
+
+	if len(s.Ocm.Interval) > 0 {
+		i, err := time.ParseDuration(s.Ocm.Interval)
+		if err != nil {
+			return nil, fmt.Errorf("ocm interval must be a valid duration: %v", err)
+		}
+		cfg.OcmInterval = i
+	}
 	return cfg, nil
 }
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,9 +24,10 @@ type Serialized struct {
 	Impersonate             string   `json:"impersonate"`
 	Gather                  []string `json:"gather"`
 	EnableGlobalObfuscation bool     `json:"enableGlobalObfuscation"`
-	Ocm         struct {
+	Ocm                     struct {
 		Endpoint string `json:"endpoint"`
 		Interval string `json:"interval"`
+		Disabled bool   `json:"disabled"`
 	}
 }
 
@@ -54,14 +55,13 @@ type Controller struct {
 	// EnableGlobalObfuscation enables obfuscation of domain names and IP addresses
 	// To see the detailed info about how anonymization works, go to the docs of package anonymization.
 	EnableGlobalObfuscation bool
-	OcmEndpoint          string
-	OcmInterval          time.Duration
 
 	Username string
 	Password string
 	Token    string
 
 	HTTPConfig HTTPConfig
+	OCMConfig  OCMConfig
 }
 
 // HTTPConfig configures http proxy and exception settings if they come from config
@@ -69,6 +69,13 @@ type HTTPConfig struct {
 	HTTPProxy  string
 	HTTPSProxy string
 	NoProxy    string
+}
+
+// OCMConfig configures the interval and endpoint for retrieving the data from OCM API
+type OCMConfig struct {
+	Interval time.Duration
+	Endpoint string
+	Disabled bool
 }
 
 type Converter func(s *Serialized, cfg *Controller) (*Controller, error)
@@ -166,15 +173,16 @@ func ToController(s *Serialized, cfg *Controller) (*Controller, error) { // noli
 	}
 
 	if len(s.Ocm.Endpoint) > 0 {
-		cfg.OcmEndpoint = s.Ocm.Endpoint
+		cfg.OCMConfig.Endpoint = s.Ocm.Endpoint
 	}
+	cfg.OCMConfig.Disabled = s.Ocm.Disabled
 
 	if len(s.Ocm.Interval) > 0 {
 		i, err := time.ParseDuration(s.Ocm.Interval)
 		if err != nil {
 			return nil, fmt.Errorf("ocm interval must be a valid duration: %v", err)
 		}
-		cfg.OcmInterval = i
+		cfg.OCMConfig.Interval = i
 	}
 	return cfg, nil
 }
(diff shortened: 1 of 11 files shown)


pkg/ocm/ocm.go
```diff
@@ -0,0 +1,161 @@
...
+			interval = cfg.OcmInterval
+			endpoint = cfg.OcmEndpoint
+		}
+	}
+}
+
+func (c *Controller) requestDataAndCheckSecret(endpoint string) {
+	data, err := requestData(endpoint)
+	if err != nil {
+		klog.Errorf("errror requesting data from %s: %v", endpoint, err)
```
Sergey1011010, line 68: Don't we need a return here?

tremes: Yeah probably yeah. Good catch..the error states will need an update anyway. I think we will introduce new IO state (something like `ScaDataDegraded`) and then probably a new alert based on this state.

pkg/ocm/ocm.go
```diff
@@ -0,0 +1,161 @@
...
+	return s, nil
+}
+
+// TODO
+// - no need to create new HTTP client every time
+// - add authorization
+// - add response status check
+// - move this to insightsclient?
+func requestData(endpoint string) ([]byte, error) {
+	req, err := http.NewRequest(http.MethodGet, endpoint, nil)
```
tremes, line 147: I guess we will need to know the cluster id here.

pkg/ocm/ocm.go
```diff
@@ -0,0 +1,161 @@
...
+	c.requestDataAndCheckSecret(endpoint)
+	for {
+		select {
+		case <-time.After(interval):
+			c.requestDataAndCheckSecret(endpoint)
+		case <-configCh:
+			cfg := c.configurator.Config()
+			interval = cfg.OcmInterval
+			endpoint = cfg.OcmEndpoint
+		}
```
natiiix, line 61: Glad to see some of Go's features finally making its way into IO! :-)

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -265,16 +265,67 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
...
 	if cv == nil {
 		return nil, ErrWaitingForVersion
 	}
-
-	endpoint = fmt.Sprintf(endpoint, cv.Spec.ClusterID)
-	klog.Infof("Retrieving report for cluster: %s", cv.Spec.ClusterID)
+	// TODO the if should be likely removed
+	if addClusterID {
+		endpoint = fmt.Sprintf(endpoint, cv.Spec.ClusterID)
+	}
```
tremes, line 327: The `if` is temporary hack and will be removed

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -284,16 +284,67 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
...
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == http.StatusOK {
+		return &resp.Body, nil
+	}
+
+	klog.Warningf("Report response status code: %d", resp.StatusCode)
+	return nil, fmt.Errorf("Report response status code: %d", resp.StatusCode)
```
natiiix, line 306: Nitpick: Couldn't we perhaps create the error, log it and then return it instead of two largely duplicate lines?

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -284,16 +284,67 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
...
+	if resp.StatusCode == http.StatusOK {
+		data, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+		return data, nil
+	}
+
+	klog.Warningf("OCM response status code: %d", resp.StatusCode)
+	return nil, fmt.Errorf("OCM response status code: %d", resp.StatusCode)
```
natiiix, line 332: Same nitpick as above. I just feel like this doesn't look very nice and it could result in some pretty confusing behavior if someone only changed one of the lines.

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -304,34 +355,33 @@ func (c Client) RecvReport(ctx context.Context, endpoint string) (*io.ReadCloser
...
-		klog.Errorf("Unable to retrieve latest report for %s: %v", cv.Spec.ClusterID, err)
-		counterRequestRecvReport.WithLabelValues(c.metricsName, "0").Inc()
 		return nil, err
 	}
+	return resp, nil
+}
 
-	counterRequestRecvReport.WithLabelValues(c.metricsName, strconv.Itoa(resp.StatusCode)).Inc()
+func validateResponse(resp *http.Response) error {
 	requestID := resp.Header.Get("x-rh-insights-request-id")
```
natiiix, line 367: Maybe put this into a constant? Doesn't really matter. I don't see any conceptual issues, so I'm really just looking at these silly small things, sorry.

natiiix (commented):
I don't see any serious issue here, so if this is the behaviour we want, then I think it's alright.

pkg/insights/insightsclient/insightsclient.go
tremes: Good catch. Thank you.

pkg/insights/insightsclient/insightsclient.go
tremes: No problem at all (some feedback is better than none :)). Thanks. I think our code base could generally use more constants ...

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -276,7 +280,7 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
 	}
 
 	if len(requestID) > 0 {
-		klog.V(2).Infof("Successfully reported id=%s x-rh-insights-request-id=%s, wrote=%d", source.ID, requestID, bytesRead)
+		klog.V(2).Infof("Successfully reported id=%s %s=%s, wrote=%d", insightsReqId, source.ID, requestID, bytesRead)
```
tremes, line 283: The order is wrong here. I need to update it here and above

tremes:
Ohh I need to fix permissions for the secret in the `openshift-config-managed` namespace.

pkg/config/config.go
```diff
@@ -24,6 +24,11 @@ type Serialized struct {
 	Impersonate             string   `json:"impersonate"`
 	Gather                  []string `json:"gather"`
 	EnableGlobalObfuscation bool     `json:"enableGlobalObfuscation"`
+	Ocm                     struct {
```
mfojtik, line 27: why is this defined as inline struct if you have this as full script on line 75? :-)

pkg/ocm/ocm.go
```diff
@@ -0,0 +1,158 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog/v2"
+)
+
+const (
+	targetNamespaceName = "openshift-config-managed"
+	secretName          = "etc-pki-entitlement" //nolint: gosec
```
mfojtik, line 19: is this operator owning this secret? what operator manage this?

tremes: Yes ...I am not sure how the "ownership" is defined, but this secret is created/updated by the Insights Operator.

pkg/config/config.go
tremes: Ohh good catch! Thank you.

tremes: Well hmmm...this is slightly different type than the `OCMConfig` (which is embedded in `Controller` type). I can delete the `OCMConfig` type and define the attributes directly in the `Controller` type....I guess I don't have any better idea right now.

natiiix (commented):
Obligatory "I have not tested it locally", but I see nothing objectively wrong with the code and the tests look good to me. I mostly looked at the changes since my last review and it all makes sense to me, as far as I'm aware of the relevant task.

Sergey1011010 (approved):
looks great to me

natiiix (approved):
I have reviewed this before and it should now be ready to be merged, so I'm giving it the real approval now. I have not tested this locally, or at least not recently, but I have checked the code again today and I see nothing worth arguing about.
/lgtm

PR #376, opened 2021-03-17
Bug 1940432: Gahter datahubs.installers.datahub.sap.com resources from SAP clusters

natiiix:
This PR introduces another SAP enhancement. Much like the others, this one will also have to be backported to 4.6.z for us to see any useful results in a reasonable amount of time.

The presence of the Datahubs resource(s) is used for the identification of SAP (SDI) clusters. As such, it has been decided that it should be collected from all of them in order for the team processing the IO archives to be able to tell SAP clusters apart from the remaining clusters. It is possible that all of the data gathered exclusively from SAP clusters could be missing, in which case it would be impossible to determine if a cluster has an SDI installation or not. This enhancement fixes that issue by gathering a resource that should, universally, be present on all SAP clusters. By gathering this resource, we also safely know which namespace was SDI (attempted to be) installed into.

We came up with a hopefully reasonably clear and unambiguous directory structure for custom resources (see "Sample archive" section below), so please express raise your complaints now before it gets backported because we would like to use this for other custom resources (especially those that are not directly associated with OpenShift) in the future.

There is one one-line change in an unrelated gatherer's file. It's a little linter warning that I forgot to fix, but it's already fixed in all of the backports, so I wanted to fix it here as well. It will not appear in any of the backports of this enhancement.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

- `docs/insights-archive-sample/customresources/installers.datahub.sap.com/datahubs/sdi/default.json`

## Documentation

- `docs/gathered-data.md` - "SAPDatahubs" section

## Unit Tests

- `pkg/gather/clusterconfig/sap_datahubs_test.go` - `TestSAPDatahubs`

## Privacy

## Changelog

No.

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-4228
Bugzilla Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1940432


docs/insights-archive-sample/config/installers.datahub.sap.com/datahubs/sdi/default.json
```diff
@@ -0,0 +1 @@
...conditions":[{"message":"Operation in progress","operationId":"3a248d","package":"3.1.13","reason":"preInstall","status":"False","timestamp":"2021-01-25T16:47:32Z","type":"inProgress"},{"message":"preInstall hooks successfully executed","operationId":"3a248d","package":"3.1.13","status":"True","timestamp":"2021-01-25T16:47:36Z","type":"preInstall"},{"message":"postInstall hooks successfully executed","operationId":"3a248d","package":"3.1.13","status":"True","timestamp":"2021-01-25T16:47:39Z","type":"postInstall"},{"message":"Operation in progress","operationId":"c03db8","package":"3.1.13","previousPackage":"3.1.13","reason":"preRunLevelStarted","status":"False","timestamp":"2021-01-25T16:47:43Z","type":"inProgress"},{"message":"preRunLevelStarted hooks successfully executed","operationId":"c03db8","package":"3.1.13","status":"True","timestamp":"2021-01-25T16:47:44Z","type":"preRunLevelStarted"},{"message":"postRunLevelStarted hooks successfully executed","operationId":"c03db8","package":"3.1.13","status":"True","timestamp":"2021-01-25T16:47:48Z","type":"postRunLevelStarted"}],"expiresAt":"2021-01-25T17:47:51Z","reason":"Expired","startedAt":"2021-01-25T16:47:51Z","state":"Failed"}}
```
tremes, line 1: Pretty format please.

docs/gathered-data.md
```diff
@@ -382,6 +382,13 @@ Relevant OpenShift API docs:
 Location in archive: config/securitycontentconstraint/, config/clusterrolebinding/
 
 
+## SAPDatahubs
+
+collects `datahubs.installers.datahub.sap.com` resources from SAP/SDI clusters.
+
+Location in archive: config/installers.datahub.sap.com/datahubs/<namespace>/<name>.json
```
tremes, line 389: Hmm we don't have any "third party" custom resources in the archive yet right? Just thinking if it's reasonable path...
Maybe `config/customresources/installers.datahub.sap.com/....` if we have more in the future.

docs/insights-archive-sample/config/installers.datahub.sap.com/datahubs/sdi/default.json
natiiix: Oh, I didn't notice we did that in the sample archive. I thought we kept things authentic.

docs/gathered-data.md
natiiix: I didn't want the path to be longer than what was strictly necessary because it is already a real pain to type. When we discussed the possibilities, this option seemed sufficiently specific that no one would confuse it. In fact, I would say that it would be best if all resources were stored in this format. I'd like to believe that in the future we are going to collect a lot more of these non-Kubernetes, non-OpenShift resources and I feel like it's a bit odd to put them into one directory even though they are not related. I can change it if you want, but it feels like the path is way too long then.

tremes: I don't have a strong opinion TBH. I just wanted to discuss it. I think the length of the path doesn't play any role. I know that the structure of the IO archive is not ideal but I would like to keep it as clear as possible. 

My feeling is that we're blindly putting everything under `config`. What about putting this CR to the root dir then? Just an idea...

natiiix: I was thinking the same, but we've been putting stuff into `config/` for so long I was not brave enough to suggest putting anything elsewhere. I would say that resource that are gathered verbatim, without any preprocessing (except for potential anonymization where applicable) should be stored in a more formal structure. Either directly in the root of the archive or in a directory like `resources/` or something. Then I'd suggest to use the most sensible directory structure based on the API group and resource name (API version should not be necessary, hopefully), followed by an optional namespace for non-cluster-wide resources and the name of the resource itself. This kind of a standardization, at least for non-core resources (like pods or nodes), should help us with organization of the archive in a more reasonable way than just throwing everything in `config/`. I'll bring this up on Monday's stand-up if I don't forget about it.

tremes:
I went through the code and did basic check on a cluster without SAP.
/lgtm

PR #377, opened 2021-03-18
Fixing code style

rluders:
This PR fixes some code styling issues, based on the linting tool report.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

No.

## Documentation

No.

## Unit Tests

It doesn't include any tests, but most lint issues were related to it.

## Privacy

## Changelog

No.

## References

https://issues.redhat.com/browse/CCXDEV-4079

### DIFF
--- a/pkg/config/configobserver/configobserver_test.go
+++ b/pkg/config/configobserver/configobserver_test.go
@@ -21,7 +21,7 @@ import (
 	clienttesting "k8s.io/client-go/testing"
 )
 
-func TestChangeSupportConfig(t *testing.T) {
+func Test_ConfigObserver_ChangeSupportConfig(t *testing.T) {
 	var cases = []struct {
 		name      string
 		config    map[string]*corev1.Secret
--- a/pkg/config/configobserver/configobserver_test.go
+++ b/pkg/config/configobserver/configobserver_test.go
@@ -2,10 +2,8 @@ package configobserver
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"reflect"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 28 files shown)


pkg/config/configobserver/configobserver_test.go
```diff
@@ -100,100 +98,11 @@ func TestChangeSupportConfig(t *testing.T) {
...
-			}
-		}
-	}()
-	<-done
-
-	if !reflect.DeepEqual(setIntervals, actualIntervals) {
-		t.Fatalf("the expected intervals didn't match actual intervals. \nExpected %v \nActual %v", setIntervals, actualIntervals)
-	}
-}
-
```
tremes, line 180: Why are you removing all this? I know it's ignored, but I don't see any harm in leaving it there.

pkg/gather/clusterconfig/config_maps_test.go
```diff
@@ -58,6 +58,10 @@ func TestConfigMapAnonymizer(t *testing.T) {
 			mustNotFail(t, err, "error creating test data %+v")
 			cm := findMap(cml, tt.configMapName)
 			mustNotFail(t, cm != nil, "haven't found a ConfigMap %+v")
+			// just to make lint happy
+			if cm != nil {
+				return
+			}
```
tremes, line 64: I don't think this is correct. I think we want to continue in the test when `cm != nil`.

pkg/config/configobserver/configobserver_test.go
rluders: Unused code should be always removed, we are using version control to handle it to us. So, no need to keep it in the repository, a few enumerated reasons related to code quality:

- "Out of sight" is also "Out of mind";
- Reduce the code base size;
- Version control handles it and keeps it forever;
- Lint will always complain about unused functions, variables, etc;

There are also few articles about it:

https://www.infoq.com/news/2017/02/dead-code/
https://blog.ndepend.com/no-excuse-dead-code/
https://www.oreilly.com/library/view/becoming-a-better/9781491905562/ch04.html

pkg/gather/clusterconfig/config_maps_test.go
rluders: Oh, wait! Wrong operator. :facepalm:

pkg/config/configobserver/configobserver_test.go
tremes: OK then it would be good to fix the test or at least create some tracking issue, because "Out of sight" is also "Out of mind" as you said :)

rluders: I'll create an issue to review this test.

0sewa0:
+1 from me, love to see the removal of dead code. Also the consistency the linting will bring will be a breath of fresh air. :) 

/lgtm

PR #379, opened 2021-03-18
Update resource request/limit in the deployment

0sewa0:
Updates the resource limits and request in the deployment according to research done previous sprints.

### Raises the resource requests to `40m` cpu and `80Mi` memory: 

> During testing it turned that if we used the original amounts as limits the operator couldn't even start up, and the new amounts are close to the minimum for the operator to start up and run at least one gather. (I was a bit more generous with the CPU)

### The limit is set to a `100m` cpu and a veery generous `1Gi` memory:

> After analyzing the metadata in 4.7 clusters we found that there are very few clusters that have very excessive memory usage.(~500 Mi)
> We identified a potential cause and are making steps to solve it also we plant to get further metadata so we can set a more reasonable limit in the future.
> This limit is just a final safe-guard so the operator wouldn't consume the sun in an unforeseen error.
> In normal/average scenarios 200Mi memory is more then enough for the operator to work without any issue. 

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- None

## Documentation

- Updated the `arch.md`

## Unit Tests

- None

## Privacy

### DIFF
--- a/docs/arch.md
+++ b/docs/arch.md
@@ -256,6 +256,27 @@ Api, we can still use non type safe custom build types called [dynamic client](k
 Such a client is used in [GatherMachineSet](pkg/gather/clusterconfig/clusterconfig.go).
 
 
+
+## Resource usage
+In the IO deployment [manifest](manifests/06-deployment.yaml) there are resource:
+- `requests` set to inform the cluster how much resources it needs at minimum.
+- `limits` set to make sure we don't overstay our welcome on a cluster.
+
+> Side note: Resource limits/requests are set for Containers, but because in our case the IO operator is the only Container in the Pod therefore these limits/requests are the same for the Pod as well.
+
+Effects of:
+- `requests`: These are checked when the pod is scheduled so to make sure that the place where it's going to be deployed has enough capacity to run it.
+- CPU `limit`: Slows down the work that the operator does, but as long as we are fast enough to not cause timeouts in network communication then we are fine.
+- Memory `limit`: If the memory usage of the pod exceeds the limit, the pod is killed without mercy and restarted. There is no easy way to notice from the operator's side that this happened, so we should tread carefully here.
+
+Both `requests` and `limits` were based on tests where we tried to starve the operator and some statistics based on the metadata.
+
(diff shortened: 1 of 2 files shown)


docs/arch.md
```diff
@@ -256,6 +256,27 @@ Api, we can still use non type safe custom build types called [dynamic client](k
...
+
+Effects of:
+- `requests`: These are checked when the pod is scheduled so to make sure that the place where it's going to be deployed has enough capacity to run it.
+- CPU `limit`: Slows down the work that the operator does, but as long as we are fast enough to not cause timeouts in network communication then we are fine.
+- Memory `limit`: If the memory usage of the pod exceeds the limit, the pod is killed without mercy and restarted. There is no easy way to notice this from the operator's side that this happened, so we should tread carefully here.
+
+Both `requests` and `limits` were based on tests where we tried to starve the operator and some statistics based on the [metadata](https://github.com/openshift/insights-operator/tree/master/docs/insights-archive-sample/insights-operator/gathers.json).
+
+>Currently the memory usage `limit` is very generous because we found that there are few clusters that have very excessive memory usage.(~500 Mi)
+We identified a potential cause and are making steps to solve it also we plan to get further metadata so we can set a more reasonable limit in the future.
```
tremes, line 275: > taking steps

docs/arch.md
```diff
@@ -256,6 +256,27 @@ Api, we can still use non type safe custom build types called [dynamic client](k
 Such a client is used in [GatherMachineSet](pkg/gather/clusterconfig/clusterconfig.go).
 
 
+
+## Resource usage
+In the IO deployment [manifest](manifests/06-deployment.yaml) there are resource:
+- `requests` set to inform the cluster how much resources it needs at minimum.
+- `limits` set to make sure we don't overstay our welcome on a cluster.
```
tremes, line 263: Well nice :) but imagine someone who reads it and knows nothing about it

0sewa0:
https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#resources-and-limits

PR #382, opened 2021-03-24
[release 4.6] Bug 1942457: Adds metadata to the IO archive

0sewa0:
This PR is a backport of https://github.com/openshift/insights-operator/pull/364
It introduces the metadata gathering to 4.6 and the recently added 2 new fields. (gathers.json).

#### Info about each gather-function:
- name
- duration in milliseconds
- the count of records collected
- errors (if any)

#### Info about the whole gather process:
- uptime_seconds: the duration since the start of the first `record.Collect`
- memory_alloc_bytes: the memory allocated at the time of creating the report. (in bytes)

Some disclaimers:

I intended to have memory usage stats for each gather-function, but it wasn't really viable. The only way I could measure it was by marshaling the produced records, which is not only inefficient but we can just as easily measure the size of the produced archive and get the same idea.
The memory_alloc stat wont show the peak memory allocation. Even if we are only talking about the gather cycle, knowing when to measure to get the max usage is impossible so we would need to monitor the entire process, which would more complex without much benefit IMO. The purpose of this field is to be able to notice if over a long period there might be a memory leak.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [X] Backporting

## Sample archive

- see docs/insights-archive-sample/insights-operator/gathers.json

## Documentation

- no

## Unit Tests

- no

## Privacy

## References

### DIFF
--- a/docs/insights-archive-sample/insights-operator/gathers.json
+++ b/docs/insights-archive-sample/insights-operator/gathers.json
@@ -1,0 +1,180 @@
+{
+    "status_reports": [
+        {
+            "name": "clusterconfig.GatherPodDisruptionBudgets",
+            "duration_in_ms": 123,
+            "records_count": 2,
+            "errors": null
+        },
+        {
+            "name": "clusterconfig.GatherMostRecentMetrics",
+            "duration_in_ms": 0,
+            "records_count": 0,
+            "errors": null
+        },
+        {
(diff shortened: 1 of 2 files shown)


tremes:
Executed locally and works as expected. 
/lgtm

PR #383, opened 2021-03-25
[release-4.7] Bug 1942068: Gahter datahubs.installers.datahub.sap.com resources from SAP clusters

natiiix:
This PR is a backport of #376, which is another SAP enhancement. Much like the others, this one will also have to be backported to 4.6.z for us to see any useful results in a reasonable amount of time.

The presence of the Datahubs resource(s) is used for the identification of SAP (SDI) clusters. As such, it has been decided that it should be collected from all of them in order for the team processing the IO archives to be able to tell SAP clusters apart from the remaining clusters. It is possible that all of the data gathered exclusively from SAP clusters could be missing, in which case it would be impossible to determine if a cluster has an SDI installation or not. This enhancement fixes that issue by gathering a resource that should, universally, be present on all SAP clusters. By gathering this resource, we also safely know which namespace was SDI (attempted to be) installed into.

We came up with a hopefully reasonably clear and unambiguous directory structure for custom resources (see "Sample archive" section below).

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [x] Backporting

## Sample archive

- `docs/insights-archive-sample/customresources/installers.datahub.sap.com/datahubs/sdi/default.json`

## Documentation

- `docs/gathered-data.md` - "SAPDatahubs" section

## Unit Tests

- `pkg/gather/clusterconfig/sap_datahubs_test.go` - `TestSAPDatahubs`

## Privacy

## Changelog

No.

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-4317
Bugzilla Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1942068
Base PR: #376

### DIFF
--- a/pkg/gather/clusterconfig/sap_datahubs.go
+++ b/pkg/gather/clusterconfig/sap_datahubs.go
@@ -1,0 +1,54 @@
+package clusterconfig
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// GatherSAPDatahubs collects `datahubs.installers.datahub.sap.com` resources from SAP/SDI clusters.
+//
+// Location in archive: customresources/installers.datahub.sap.com/datahubs/<namespace>/<name>.json
+func GatherSAPDatahubs(g *Gatherer, c chan<- gatherResult) {
+	gatherDynamicClient, err := dynamic.NewForConfig(g.gatherKubeConfig)
+	if err != nil {
+		c <- gatherResult{errors: []error{err}}
+		return
+	}
+
+	records, errors := gatherSAPDatahubs(g.ctx, gatherDynamicClient)
+	c <- gatherResult{records: records, errors: errors}
+}
+
+func gatherSAPDatahubs(ctx context.Context, dynamicClient dynamic.Interface) ([]record.Record, []error) {
+	datahubsResource := schema.GroupVersionResource{Group: "installers.datahub.sap.com", Version: "v1alpha1", Resource: "datahubs"}
+	datahubsList, err := dynamicClient.Resource(datahubsResource).List(ctx, metav1.ListOptions{})
+	if errors.IsNotFound(err) {
+		return nil, nil
+	}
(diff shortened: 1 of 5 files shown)


pkg/gather/clusterconfig/sap_datahubs.go
```diff
@@ -0,0 +1,54 @@
...
+		c <- gatherResult{errors: []error{err}}
+		return
+	}
+
+	records, errors := gatherSAPDatahubs(g.ctx, gatherDynamicClient)
+	c <- gatherResult{records: records, errors: errors}
+}
+
+func gatherSAPDatahubs(ctx context.Context, dynamicClient dynamic.Interface) ([]record.Record, []error) {
+	datahubsResource := schema.GroupVersionResource{Group: "installers.datahub.sap.com", Version: "v1alpha1", Resource: "datahubs"}
```
tremes, line 30: We already have a variable for this GVR in 4.7. Can you please update it?

pkg/gather/clusterconfig/sap_datahubs_test.go
```diff
@@ -0,0 +1,81 @@
...
+    namespace: example-namespace1
+`
+	datahubYAML2 := `apiVersion: installers.datahub.sap.com/v1alpha1
+kind: DataHub
+metadata:
+    name: example-datahub
+    namespace: example-namespace2
+`
+
+	datahubsResource := schema.GroupVersionResource{Group: "installers.datahub.sap.com", Version: "v1alpha1", Resource: "datahubs"}
```
tremes, line 30: Same here pls.

pkg/gather/clusterconfig/sap_datahubs.go
natiiix: I guess it was just now added with the changes that I rebased on top of because it wasn't there when I created the PR.

pkg/gather/clusterconfig/sap_datahubs_test.go
natiiix: This is exactly the same code as in `master`. The idea here was that if someone introduces a type into the constant, then the tests would happily pass if they also use the constant. However, if each of them has their own hardcoded GVR variable, then they will start failing.

pkg/gather/clusterconfig/sap_datahubs.go
tremes: Your guess is correct AFAIK :)

tremes:
Pretty straightforward & clear. I ran it locally as usual, but no SAP on my cluster. Thanks
/lgtm

PR #389, opened 2021-04-07
Bug 1948966: One-off gather

0sewa0:
Updates the gather_job.go so that it doesn't always require a `--kubeconfig` param, also adds the `docs/gather-job.yaml` which is a one off job to run a single gather.
```
# Create Job
$ kubectl apply -n openshift-insights -f docs/gather-job.yaml
...


# Get Job's pod's name
$ kubectl describe -n openshift-insights job/insights-operator-job
...
Events:
  Type    Reason            Age    From            Message
  ----    ------            ----   ----            -------
  Normal  SuccessfulCreate  7m18s  job-controller  Created pod: insights-operator-job-zhmrd <-- THIS


# Check if the gather finished (optional)
$ kubectl logs -n openshift-insights insights-operator-job-zhmrd insights-operator
...
I0407 11:55:38.192084       1 diskrecorder.go:34] Wrote 108 records to disk in 33ms <-- THIS


# Get the archives
$ kubectl cp openshift-insights/insights-operator-job-zhmrd:/var/lib/insights-operator ./insights-data
...


# Cleanup
$ kubectl delete -n openshift-insights job insights-operator-job
...
```

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- nope

## Documentation

- TODO

## Unit Tests

- nope

## Privacy

### DIFF
--- a/docs/gather-job.yaml
+++ b/docs/gather-job.yaml
@@ -29,18 +29,9 @@ spec:
       volumes:
       - name: snapshots
         emptyDir: {}
-          #sizeLimit: 1Gi # bug https://bugzilla.redhat.com/show_bug.cgi?id=1713207
-      - name: trusted-ca-bundle
-        configMap:
-          name: trusted-ca-bundle
-          optional: true
       - name: service-ca-bundle
         configMap:
           name: service-ca-bundle
-          optional: true
-      - name: serving-cert
-        secret:
-          secretName: openshift-insights-serving-cert
           optional: true
       initContainers:
       - name: insights-operator
@@ -49,14 +40,9 @@ spec:
         volumeMounts:
         - name: snapshots
           mountPath: /var/lib/insights-operator
-        - mountPath: /var/run/configmaps/trusted-ca-bundle
-          name: trusted-ca-bundle
+        - name: service-ca-bundle
+          mountPath: /var/run/configmaps/service-ca-bundle
           readOnly: true
-        - mountPath: /var/run/configmaps/service-ca-bundle
-          name: service-ca-bundle
-          readOnly: true
-        - mountPath: /var/run/secrets/serving-cert
-          name: serving-cert
         ports:
         - containerPort: 8443
           name: https
@@ -64,17 +50,6 @@ spec:
           requests:
             cpu: 10m
             memory: 70Mi
-        env:
-        - name: POD_NAME
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.name
-        - name: POD_NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
-        - name: RELEASE_VERSION
-          value: "0.0.1-snapshot"
         args:
         - gather
         - -v=4
--- a/docs/gather-job.yaml
+++ b/docs/gather-job.yaml
@@ -42,7 +42,7 @@ spec:
         secret:
           secretName: openshift-insights-serving-cert
           optional: true
-      containers:
+      initContainers:
       - name: insights-operator
         image: quay.io/openshift/origin-insights-operator:latest
         terminationMessagePolicy: FallbackToLogsOnError
@@ -79,3 +79,11 @@ spec:
         - gather
         - -v=4
         - --config=/etc/insights-operator/server.yaml
+      containers:
+        - name: sleepy
+          image: busybox
+          args:
+            - /bin/sh
+            - -c
+            - sleep 10m
+          volumeMounts: [{name: snapshots, mountPath: /var/lib/insights-operator}]
--- a/docs/gather-job.yaml
+++ b/docs/gather-job.yaml
@@ -1,0 +1,81 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: insights-operator-job
+  annotations:
+    config.openshift.io/inject-proxy: insights-operator
+spec:
+  backoffLimit: 3
+  ttlSecondsAfterFinished: 600
+  template:
+    spec:
+      restartPolicy: OnFailure
+      serviceAccountName: operator
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+        node-role.kubernetes.io/master: ""
+      tolerations:
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/master
+        operator: Exists
+      - effect: NoExecute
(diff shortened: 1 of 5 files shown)


docs/gather-job.yaml
```diff
@@ -0,0 +1,89 @@
...
+        configMap:
+          name: service-ca-bundle
+          optional: true
+      - name: serving-cert
+        secret:
+          secretName: openshift-insights-serving-cert
+          optional: true
+      initContainers:
+      - name: insights-operator
+        image: quay.io/openshift/origin-insights-operator:latest
```
0sewa0, line 47: To test it you should replace this with an image that has the gather_job.go changes in it. (you can use mine: `quay.io/0sewa0/oneoff` or build your own :) )

docs/gather-job.yaml
```diff
@@ -0,0 +1,89 @@
...
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: RELEASE_VERSION
+          value: "0.0.1-snapshot"
```
tremes, line 77: Do we need this env vars?

docs/gather-job.yaml
```diff
@@ -0,0 +1,89 @@
...
+        key: node-role.kubernetes.io/master
+        operator: Exists
+      - effect: NoExecute
+        key: node.kubernetes.io/unreachable
+        operator: Exists
+        tolerationSeconds: 900
+      - effect: NoExecute
+        key: node.kubernetes.io/not-ready
+        operator: Exists
+        tolerationSeconds: 900
```
tremes, line 28: Do we need these tolerations?

docs/gather-job.yaml
```diff
@@ -0,0 +1,89 @@
...
+          name: trusted-ca-bundle
+          optional: true
+      - name: service-ca-bundle
+        configMap:
+          name: service-ca-bundle
+          optional: true
+      - name: serving-cert
+        secret:
+          secretName: openshift-insights-serving-cert
+          optional: true
```
tremes, line 44: These volumes (and their mounts) are required? I guess so but want to double check

docs/gather-job.yaml
tremes: For testing it's OK but I think we should recommend users to use the same image they have in their original IO deployment (Just a note for documentation... :)) .

pkg/controller/gather_job.go
```diff
@@ -79,6 +78,7 @@ func (d *GatherJob) Gather(ctx context.Context, kubeConfig *rest.Config, protoKu
 		return err
 	}
 
-	klog.Warning("stopped")
```
tremes, line 82: OT - I am still wondering what's the purpose of this in our normal execution..... :)

docs/gather-job.yaml
0sewa0: nope, deleted them

docs/gather-job.yaml
0sewa0: I see no reason for removing them, it allows the IO pod/job to be deployed on nodes with certain Taints.

docs/gather-job.yaml
0sewa0: I left in the `service-ca-bundle` for the metrics config, and the `snapshots` for the actual storage, the rest was not necessary

pkg/controller/gather_job.go
0sewa0: No idea either, its nice to know that the `operator.go` finished ie. started every necessary goroutine for IO to work, but why is it this simple warning is a bit weird.

Maybe in a bygone age, the `operator.go` shouldn't have stopped :)

pkg/cmd/start/start.go
```diff
@@ -81,28 +81,32 @@ func runGather(operator controller.GatherJob, cfg *controllercmd.ControllerComma
...
-		}
-		kubeConfig, err := clientcmd.NewClientConfigFromBytes(kubeConfigBytes)
-		if err != nil {
-			klog.Fatal(err)
-		}
-		clientConfig, err := kubeConfig.ClientConfig()
-		if err != nil {
-			klog.Fatal(err)
+		var kubeConfig *rest.Config
+		if kubeConfigPath := cmd.Flags().Lookup("kubeconfig").Value.String(); len(kubeConfigPath) > 0 {
```
rluders, line 85: If the `kubeConfigPath` is a string would be better to check it by: `kubeConfigPath != ""`

rluders (changes requested):

pkg/controller/gather_job.go
```diff
@@ -82,17 +83,17 @@ func (d *GatherJob) Gather(ctx context.Context, kubeConfig *rest.Config, protoKu
...
 	// config to the recorder
 	clusterConfigGatherer := clusterconfig.New(gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig, anonymizer)
-	err = clusterConfigGatherer.Gather(ctx, configObserver.Config().Gather, recorder)
+	err = clusterConfigGatherer.Gather(ctx, configObserver.Config().Gather, rec)
 	if err != nil {
 		return err
 	}
 
-	klog.Warning("stopped")
+	rec.Flush()
```
tremes, line 96: Why not with `defer`? :)

0sewa0: leftover from the previous commit, where I needed a sleep at the very end so I couldn't use defer :)

docs/gather-job.yaml
```diff
@@ -0,0 +1,64 @@
...
+          requests:
+            cpu: 10m
+            memory: 70Mi
+        args:
+        - gather
+        - -v=4
+        - --config=/etc/insights-operator/server.yaml
+      containers:
+        - name: sleepy
+          image: busybox
```
tremes, line 59: Is this image available in the cluster registry or does it need to be downloaded? I guess the latter one....and that would be a bit problematic in an air-gapped cluster.

0sewa0: it might be worth mentioning it in the docs for air-gaped users, but I don't know any simpler way to run anything on openshift :/

rluders (approved):
/lgtm

PR #396, opened 2021-04-13
Bug 1948943: Workloads - Add limit for the number of pods gathered

tremes:
This PR adds limit 3000 for number of gathered pods in the new workload fingerprint gatherer

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

No new data

## Documentation

No new documentation

## Unit Tests

- `path/to/file_test.go`

## Privacy

### DIFF
--- a/pkg/gather/clusterconfig/workloads.go
+++ b/pkg/gather/clusterconfig/workloads.go
@@ -34,6 +34,7 @@ import (
 const (
 	// workloadGatherPageSize is 500 (the default for Kube).
 	workloadGatherPageSize = 500
+	podsLimit              = 3000
 )
 
 // workloadPods is the top level description of the workloads on the cluster, primarily
@@ -157,6 +158,7 @@ func gatherWorkloadInfo(ctx context.Context, coreClient corev1client.CoreV1Inter
 
 	// load pods in order
 	start := time.Now()
+	limitReached := false
 
 	var info workloadPods
 	var namespace string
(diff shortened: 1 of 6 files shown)


pkg/gather/clusterconfig/workloads.go
```diff
@@ -34,6 +34,7 @@ import (
 const (
 	// workloadGatherPageSize is 500 (the default for Kube).
 	workloadGatherPageSize = 500
+	podsLimit              = 3000
```
0sewa0, line 37: Explaining the "magic-number" might come in handy later

tremes: Good point. Let me add some "explanation"

smarterclayton:
Why was the limit set to 3000?  Is that number based on anything?  That's a bit too low to capture the key workloads we'd like.  When I ran the numbers in the background I was thinking more 15k.

tremes:
@smarterclayton Yes it's based on my estimation and from what I've seen it should cover majority of the clusters. My math was following. Pod with a new image is around 500 B. 3000*500 that makes 1,5 MB in the worst case and before compression. I know there won't likely be 3000 new images (the best case is pod for ~150B, that makes 450 KB). 

We can always increase, but with a limit of 7k pods, I am afraid we will most likely have archives above 1 MB (after compression) and I am not sure how this will affect our pipelines TBH.

PR #397, opened 2021-04-14
Bug 1951034: Split up the GatherClusterOperators into smaller parts

rluders:
This PR splits the existing `GatherClusterOperators` into two pieces:

- `GatherClusterOperators`
- `GatherClusterOperatorsPodAndEvents`

It would be possible that in the future we decide to split it even more, but for now, it is clear enough. 

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

N/A

## Documentation

- `docs/gathered-data.md` (updated)

## Unit Tests

- `pkg/gather/clusterconfig/operators_pods_and_events_test.go` (new)
- `pkg/gather/clusterconfig/operators_test.go` (updated)

## Privacy

## Changelog

No.

## References

https://issues.redhat.com/browse/CCXDEV-4222
https://bugzilla.redhat.com/show_bug.cgi?id=1951034

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,10 @@ unit-verbose:
 ## Linting
 ## --------------------------------------
 
+.PHONY: precommit
+precommit: ## Executes the pre-commit hook (check the stashed changes)
+	./.githooks/pre-commit
+
 .PHONY: lint
 lint: $(GOLANGCI_LINT) ## Executes the linting tool (vet, sec, and others)
 	$(GOLANGCI_LINT) run
--- a/pkg/gather/clusterconfig/0_gatherer.go
+++ b/pkg/gather/clusterconfig/0_gatherer.go
@@ -71,7 +71,7 @@ var gatherFunctions = map[string]gathering{
 	"pdbs":                              important(GatherPodDisruptionBudgets),
 	"metrics":                           failable(GatherMostRecentMetrics),
 	"operators":                         important(GatherClusterOperators),
-	"operators_unhealthy":               important(GatherUnhealthyClusterOperator),
+	"operators_pods_and_events":         important(GatherClusterOperatorPodsAndEvents),
 	"container_images":                  important(GatherContainerImages),
 	"workload_info":                     failable(GatherWorkloadInfo),
 	"nodes":                             important(GatherNodes),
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -60,6 +60,10 @@ unit-verbose:
 ## Linting
 ## --------------------------------------
 
+.PHONY: precommit
```
rluders, line 63: I added it to this PR, ´cause I wanted to check the pre-commit before actually commit, and... well... seems useful.

pkg/gather/clusterconfig/operators_unhealty.go
```diff
@@ -0,0 +1,315 @@
...
+	Reason        string    `json:"reason"`
+	Message       string    `json:"message"`
+}
+
+// CompactedEventList is collection of events
+type CompactedEventList struct {
+	Items []CompactedEvent `json:"items"`
+}
+
+func GatherUnhealthyClusterOperator(g *Gatherer, c chan<- gatherResult) {
```
tremes, line 40: I would change the name (also of the file) to emphasize the difference between the original `GatherClusterOperators` and this one. This one is about ClusterOperator pods and events (I think it's should be obvious from the name) whereas the original is solely about `ClusterOperator` resource.

Otherwise looks good at the first glance.

rluders: I'm trying to split some pieces from this one yet. But, thanks. I'll keep in mind the change it it doesn't get splited.

rluders:
We probably could split it more, but it was causing a lot of code duplication or even a huge amount of "utils" functions. For sake of clarity and simplicity, I believe that this split is good enough. I didn´t saw any other benefits by splitting it even more.

pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -71,6 +71,7 @@ var gatherFunctions = map[string]gathering{
 	"pdbs":                              important(GatherPodDisruptionBudgets),
 	"metrics":                           failable(GatherMostRecentMetrics),
 	"operators":                         important(GatherClusterOperators),
+	"operators_pods_and_events":         important(GatherClusterOperatorPodsAndEvents),
```
tremes, line 74: I think the new one could be "failable" or WDYT? I think it is likely that it will have some errors in case of broken cluster/a lot of degraded clusteroperators.

tremes: Maybe this assumption is not correct. We should double check.

rluders: I don't think so... I mean, it wasn't "failable" before when it was together with `operators`. I mean, I agree that once we found a reasonable way to split the logs, definitely the logs should be "failable"... but right now, we are probably fine with the way that it is.

0sewa0:
tested it via a cluster-bot cluster, everything worked just fine. 

/lgtm

PR #398, opened 2021-04-14
Bug 1951456: Docs and lint fixes

0sewa0:
Goes through the codebase and adds/updates the docs for public functions and fixes the linting issues

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- none

## Documentation

- none

## Unit Tests

- none

## Privacy

### DIFF
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -19,7 +19,6 @@
 //       - 172.30.0.5 -> 172.30.0.1  // new subnet, so we use a new set of fake IPs
 //       - 127.0.0.1 -> 127.0.0.1  // it was the first IP, so the new IP matched the original in this case
 //       - 10.0.134.130 -> 0.0.0.0  // ip doesn't match any subnet, we replace such IPs with 0.0.0.0
-
 package anonymization
 
 import (
@@ -113,18 +112,10 @@ func NewAnonymizerFromConfigClient(
 	for _, network := range networksConfig.Spec.ClusterNetwork {
 		networks = append(networks, network.CIDR)
 	}
-	for _, network := range networksConfig.Spec.ServiceNetwork {
-		networks = append(networks, network)
-	}
-	for _, network := range networksConfig.Spec.ExternalIP.AutoAssignCIDRs {
-		networks = append(networks, network)
-	}
-	for _, network := range networksConfig.Spec.ExternalIP.Policy.AllowedCIDRs {
-		networks = append(networks, network)
-	}
-	for _, network := range networksConfig.Spec.ExternalIP.Policy.RejectedCIDRs {
-		networks = append(networks, network)
-	}
+	networks = append(networks, networksConfig.Spec.ServiceNetwork...)
+	networks = append(networks, networksConfig.Spec.ExternalIP.AutoAssignCIDRs...)
+	networks = append(networks, networksConfig.Spec.ExternalIP.Policy.AllowedCIDRs...)
+	networks = append(networks, networksConfig.Spec.ExternalIP.Policy.RejectedCIDRs...)
 
 	clusterConfigV1, err := kubeClient.CoreV1().ConfigMaps("kube-system").Get(ctx, "cluster-config-v1", metav1.GetOptions{})
 	if err != nil {
@@ -133,9 +124,7 @@ func NewAnonymizerFromConfigClient(
 
 	if installConfig, exists := clusterConfigV1.Data["install-config"]; exists {
 		networkRegex := regexp.MustCompile(Ipv4NetworkRegex)
-		for _, network := range networkRegex.FindAllString(installConfig, -1) {
-			networks = append(networks, network)
-		}
+		networks = append(networks, networkRegex.FindAllString(installConfig, -1)...)
 	}
 
 	// we're sorting by subnet lengths, if they are the same, we use subnet itself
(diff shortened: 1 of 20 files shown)


pkg/anonymization/anonymizer.go
```diff
@@ -113,18 +112,10 @@ func NewAnonymizerFromConfigClient(
...
-	for _, network := range networksConfig.Spec.ExternalIP.AutoAssignCIDRs {
-		networks = append(networks, network)
-	}
-	for _, network := range networksConfig.Spec.ExternalIP.Policy.AllowedCIDRs {
-		networks = append(networks, network)
-	}
-	for _, network := range networksConfig.Spec.ExternalIP.Policy.RejectedCIDRs {
-		networks = append(networks, network)
-	}
+	networks = append(networks, networksConfig.Spec.ServiceNetwork...)
```
rluders, line 115: :open_mouth: beaultiful.

pkg/config/configobserver/configobserver.go
```diff
@@ -294,12 +319,28 @@ func (c *Controller) setConfigLocked(config *config.Controller) {
 			}
 		}
 	} else {
-		klog.V(2).Infof(
-			"Configuration set: enabled=%t endpoint=%s interval=%s username=%t token=%t reportEndpoint=%s initialPollingDelay=%s minRetryTime=%s pollingTimeout=%s",
-			config.Report, config.Endpoint, config.Interval, len(config.Username) > 0, len(config.Token) > 0, config.ReportEndpoint,
-			config.ReportPullingDelay, config.ReportMinRetryTime, config.ReportPullingTimeout)
+		klog.V(2).Infof( //nolint: gomnd
+			"Configuration updated: "+
```
Sergey1011010, line 323: `set` became `updated`, maybe it's better to refactor it somehow. What about creating a function `toStr` on config?

pkg/controller/status/status.go
```diff
@@ -39,6 +39,8 @@ type Configurator interface {
 	Config() *config.Controller
 }
 
+// Responsible for managing the status of the operator according the status of the sources.
```
Sergey1011010, line 42: according to

pkg/controller/status/status.go
```diff
@@ -39,6 +39,8 @@ type Configurator interface {
...
+// Sources come from different major parts of the codebase, representing the status of said part of the codebase.
```
Sergey1011010, line 43: `of said part`? What does it mean? :)

pkg/controller/status/status.go
```diff
@@ -53,7 +55,10 @@ type Controller struct {
 	start    time.Time
 }
 
-func NewController(client configv1client.ConfigV1Interface, coreClient corev1client.CoreV1Interface, configurator Configurator, namespace string) *Controller {
+// Creates a status conteroller, responsible for monitoring the operators status and updating the it's cluster status accordingly.
```
Sergey1011010, line 58: `conteroller` typo

pkg/controller/status/status.go
0sewa0: Its trying to reference the `major` parts in the previous part of the sentence :) Will try to make it clearer

pkg/config/config.go
```diff
@@ -58,7 +58,31 @@ type HTTPConfig struct {
...
 type Converter func(s *Serialized, cfg *Controller) (*Controller, error)
 
-func ToController(s *Serialized, cfg *Controller) (*Controller, error) {
+// Returns the important fields of the config in a string form
+func (c *Controller) ToString() string {
+	return fmt.Sprintf("enabled=%t "+
+		"endpoint=%s "+
+		"interval=%s "+
+		"username=%t "+
+		"token=%t "+
```
tremes, line 67: I know we had it before, but should we log this? :)

pkg/config/config.go
```diff
@@ -129,6 +153,8 @@ func ToController(s *Serialized, cfg *Controller) (*Controller, error) {
 	return cfg, nil
 }
 
+// Creates/updates a config Controller according to the Serialized config.
+// Makes sure that the config is correct, but only checks fields necessary for disconnected operatotion.
```
tremes, line 157: Typo `operatotion`

pkg/config/config.go
0sewa0: nope, we probably shouldn't :D

0sewa0: but we only log if we have a token or not. (`len(c.Token) > 0`)
so its fine

pkg/controller/operator.go
```diff
@@ -45,7 +45,7 @@ type Operator struct {
 // 3. Initiates the recorder and starts the periodic record pruneing
 // 4. Starts the periodic gathering
 // 5. Creates the insights-client and starts uploader and reporter
-func (s *Operator) Run(ctx context.Context, controller *controllercmd.ControllerContext) error {
+func (s *Operator) Run(ctx context.Context, controller *controllercmd.ControllerContext) error { //nolint: funlen
```
rluders, line 48: Why did you ignore the `funlen` here, but at the `status.NewController` you put it into separated lines?

0sewa0: here `funlen`  is about how many statements there are in the `Run` method so the `Run` would need to be chopped into smaller functions, while `status.NewController` had a really long function signature which was easy to fix :)

rluders (approved):
/lgtm

PR #399, opened 2021-04-14
[WIP] Support of gatherers with different periods

Sergey1011010:
This PR implements mainly implements support of gatherers with different periods, but also refactores gathering logic making it more clear and easier to use. 

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

Metadata is now split to 2 files:
- insights-operator/archive-metadata.json contains metadata related to the whole archive
- insights-operator/gatherers/*-metadata.json contains metadata related to a specific gatherer

## Documentation
// TODO:

- `path/to/documentation.md`

## Unit Tests
// TODO:

- `path/to/file_test.go`

## Privacy
// TODO:

## References
// TODO:


pkg/controller/periodic/periodic.go
```diff
@@ -98,10 +108,23 @@ func (c *Controller) Gather() {
...
-	for name := range c.gatherers {
+
+	// flush when all necessary gatherers were processed
+	defer func() {
+		if err := c.recorder.Flush(); err != nil {
+			klog.Errorf("Unable to flush the recorder: %v", err)
+		}
+	}()
+
+	for periodMultiplier, gatherer := range c.gatherers {
```
tremes, line 119: I don't think I am in favor of this `periodMultiplier`. I think it's not so clear and maybe can bring some unexpected consequences. We just want to run the workloads gatherer say once a day and I think the period should be ideally independent. I think running something like:

`go wait.Until(func() { c.periodicTrigger(stopCh) }, time.Second, stopCh)` 

with different period (defined inside the `periodicTrigger`) would be more appropriate.

pkg/gather/interface.go
```diff
@@ -1,11 +1,16 @@
...
+	"github.com/openshift/insights-operator/pkg/gather/common"
 )
 
+// Interface is an interface for gathering types
 type Interface interface {
-	Gather(context.Context, []string, recorder.Interface) error
+	// GetName returns the name of the gatherer
+	GetName() string
+	// ListGatheringFunctions lists names of the functions implemented by current gatherer
+	ListGatheringFunctions() []string
```
tremes, line 12: Wouldn't a slice of `common.GatheringFunction` make more sense as the return value?

pkg/controller/periodic/periodic.go
Sergey1011010: But then we are back to having different archives with and without clusterconfig gatherer

pkg/gather/interface.go
Sergey1011010: Hm returning the list of `common.GatheringClosure` should work

tremes:
I think it's a lot of changes in this PR. I mean it includes good ideas, but I am not sure it's the right timing to do such a big change (Maybe I would rather wait when there's no feature freeze = no bugzillas needed). I think it might also happen that we won't need this extra period gatherer (that's also the reason why I wouldn't invest such an effort right now).

I know it's still WIP, but I am missing the original workloads gatherer. I was also playing little bit with this idea and maybe something like [this](https://github.com/tremes/insights-operator/commit/8e9169d64c08dd8127e61fcb6f5a39849cdec21d) could be sufficient, but I am not saying that's the way we should do that. 
I am also not sure I like the split of the metadata file into more files, but maybe it's OK :)

pkg/controller/periodic/periodic.go
tremes: I don't think so. I think it's just about when you call recorder `Flush` and you already share it (few lines above) for all the gatherers right.

Sergey1011010: If we have multiple goroutines doing gathering we'd also need to make sure there's no syncronization problems when they run at the same time.

Sergey1011010:
@tremes what's the problem with having bugzilla, btw? 

> I am also not sure I like the split of the metadata file into more files, but maybe it's OK :)

I'd prefer the split, but we can actually keep one file by changing logic little bit.

tremes:
@Sergey1011010 No problem with having bugzillas. My point is that we now need a new bugzilla for every PR and if we will need some additional fixes then we need new bugzilla for each (it's always a risk with some bigger changes). I know that we didn't care much about the feature freeze in the past, but it probably makes sense not to make some bigger changes when there is this freeze.

Sergey1011010:
@tremes so do we want to postpone this?

tremes:
@Sergey1011010 I am not sure. Maybe I am too cautious. Let's see what other team members think.

pkg/controller/periodic/periodic.go
tremes: Yes probably yes

pkg/gather/gather.go
```diff
@@ -0,0 +1,254 @@
...
+	if err := rec.Record(archiveMetadata); err != nil {
+		errs = append(errs, fmt.Errorf("unable to record archive metadata because of error: %v", err))
+	}
+
+	return sumErrors(errs)
+}
+
+// StartGatheringConcurrently starts gathering of enabled functions of the provided gatherer and returns a channel
+// with results which will be closed when processing is done
+func StartGatheringConcurrently(
```
tremes, line 179: Does this function have to be exported (starting with upper case letter)?

pkg/gather/gather.go
```diff
@@ -0,0 +1,254 @@
...
+// GetListOfEnabledFunctionForGatherer parses a list of gathering functions to enable,
+// which has the following structure: []string {
+//   "clusterconfig/container_images",
+//   "clusterconfig/nodes",
+//   "clusterconfig/authentication",
+//   "othergatherer/some_function",
+// } where each item consists of a gatherer name and a function name split by a slash.
+// If there's a string "ALL", we enable everything and return the first parameter as true,
+// otherwise it will be false and the second parameter will contain function names
+func GetListOfEnabledFunctionForGatherer(gathererName string, allFunctionsList []string) (bool, []string) {
```
tremes, line 223: Same here. Does this function have to be exported (starting with upper case letter)?

pkg/gather/tasks_processing.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/openshift/insights-operator/pkg/gather/common"
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// TODO: docs
+
+type Task struct {
+	Name    string
+	F       common.GatheringClosure
```
tremes, line 16: I don't know what the Go convention is, but calling it `F` looks to me quite cryptic. What about call it `Run` ?

pkg/gather/tasks_processing.go
```diff
@@ -0,0 +1,73 @@
...
+
+		for _, task := range tasks {
+			wg.Add(1)
+			go func(task Task) {
+				startTime := time.Now()
+				var panicked interface{}
+				var records []record.Record
+				var errs []error
+
+				func() {
```
tremes, line 46: Do we need this wrapping function?

pkg/gather/gather.go
```diff
@@ -0,0 +1,254 @@
...
+		if result.Panic != nil {
+			errStr := fmt.Sprintf(
+				"gatherer %v's function %v panicked with error: %v",
+				gathererName, result.FunctionName, result.Panic,
+			)
+
+			if result.IgnoreErrors {
+				klog.Error(errStr)
+			} else {
+				panic(errStr)
```
tremes, line 108: Checking panic of particular gatherers is a good idea IMO, but do we want to do this? It looks like that nothing will be gathered for the given gatherer and there isn't any related info in the metadata.

pkg/gather/gather.go
Sergey1011010: Actually no, it can be private, it will be used only in this package and in unit tests

pkg/gather/gather.go
Sergey1011010: Probably not

pkg/gather/tasks_processing.go
Sergey1011010: My thoughts were that it's a function pointer, so `F`, but can be `Run` to, why not

pkg/gather/tasks_processing.go
Sergey1011010: yup, to catch the panic by `defer` and continue execution

pkg/gather/gather.go
Sergey1011010: There is, I added panic field to metadata as well. 

https://github.com/openshift/insights-operator/pull/399/files#diff-bdd45c95687562d23e5fdd43dea644d8fb86c95354192cedd72835743ba9fc0fR45

The idea of this whole catching panic thing is to make the IO more stable for failable gatherers.

pkg/gather/tasks_processing.go
tremes: Ahh now I guess I see it. We want to call defer and get possible panicked result before writing to the channel.

pkg/gather/gather.go
Sergey1011010: Then `panic(errStr)` will return panic back and it will be the same as existing behaviour.

tremes: I don't think so. Right now when there's a panic -> container terminates AFAIK -> pod restart and when experimenting with this draft  I have seen the behavior mentioned above...maybe I am missing something or was doing something wrong.

Sergey1011010: @tremes it's actually previous code not working correctly. I found out that exponential backoff which we use here https://github.com/openshift/insights-operator/blob/master/pkg/controller/periodic/periodic.go#L104 panics only when `ReallyCrash` set to `true` https://github.com/openshift/insights-operator/blob/master/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go#L53 (what an unobvious way...) which is set in this file https://github.com/openshift/insights-operator/blob/master/vendor/github.com/openshift/library-go/pkg/serviceability/panic.go by `BehaviorOnPanic` which is called in some places in the libs(`grep -r BehaviorOnPanic`), but also in our code here https://github.com/openshift/insights-operator/blob/master/pkg/cmd/start/start.go#L123 as you can see from the file https://github.com/openshift/insights-operator/blob/master/vendor/github.com/openshift/library-go/pkg/serviceability/panic.go there are multiple options and if you set `OPENSHIFT_ON_PANIC` env variable to `crash` it's gonna panic, if you keep it empty, the default behaviour of just logging an error is used. Why did the previous code stop? Because the panic happened in the goroutine and couldn't be catched by exponential backoff.

tremes: Good findings. I am not saying that the original behavior is correct. I am just saying that right now (with your PR) we will not fail (container terminates), but it may also happen that there will be almost empty archive with no corresponding information in the metadata. 

In other words this is panicking in the main goroutine (if I am not mistaken) and that is not necessary IMO.

tremes:
I've got another idea about the extra period. We could introduce a new variable e.g `dailyInterval` in the `periodicTrigger` method (in `periodic.go`) and increment this variable by the "common" interval in https://github.com/openshift/insights-operator/pull/399/files#diff-dec12f0cf28a91a837b46e755b46c081c9014121cdb770b2cd8ac916af29c8deR173 and then only add condition e.g:
```
if dailyInterval >= 24*time.Hour {
    // extra gather with workloads here
}
```
I think this would be more clear.

Sergey1011010:
@tremes I think it works only as a temporarily solution, it's not very flexible

pkg/gather/gather.go
Sergey1011010: @tremes, I've changed the logic so that now panics are logged and treated as errors. In case there was a panic in a gathering function we won't crash the operator and get nice and healthy archive with all other collected data and panic information about this bad function, which in my opinion is better than crashing cuz in case of a crash we won't get any archive. WDYT?

Sergey1011010:
@tremes changed the logic of periods. Now we have gatherers which are added to each archive and optional gatherers which are processed only when it's there time, periods don't need to match with interval in config. Optional gatherers have periods(time.Duration) which are now defined in the file, but can be easily moved to the config. Tomorrow I'm gonna replace `some_gatherer` with the real one, rebase and it will be ready.

pkg/controller/operator.go
```diff
@@ -120,14 +119,11 @@ func (s *Operator) Run(ctx context.Context, controller *controllercmd.Controller
 	rec := recorder.New(recdriver, s.Interval, anonymizer)
 	go rec.PeriodicallyPrune(ctx, statusReporter)
 
-	// the gatherers periodically check the state of the cluster and report any
-	// config to the recorder
-	clusterConfigGatherer := clusterconfig.New(
+	// the gatherers are periodically called to collect the data from the cluster and write the results to the recorder
+	importantGatherers, optionalGatherers := gather.CreateAllGatherers(
```
tremes, line 123: I think the naming `importantGatherers` and `optionalGatherers` is not that clear and can be a litle bit misleading. I would call them for example `clusterConfigGatherer` and `workloadGatherer`. WDYT?

pkg/gather/gather.go
```diff
@@ -0,0 +1,247 @@
...
+	Uptime float64 `json:"uptime_seconds"`
+	// IsGlobalObfuscationEnabled shows if obfuscation(hiding IPs and cluster domain) is enabled
+	IsGlobalObfuscationEnabled bool `json:"is_global_obfuscation_enabled"`
+}
+
+// CreateAllGatherers creates all the gatherers important gatherers and optional ones with their periods
+func CreateAllGatherers(
+	gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config,
+	anonymizer *anonymization.Anonymizer,
+) ([]Interface, map[time.Duration]Interface) {
```
tremes, line 62: I find these return values a little bit confusing TBH. I think returning just `[]Interface` (maybe also change the name of the `Interface`?) would be more clear. 

Just Friday's idea - What about adding the last processed time (and maybe the `period` as well?) to each gatherer and access it via some `GetLastProcessedTime` (via the `Interface`)?

pkg/controller/operator.go
Sergey1011010: Yeah, naming could be better, but the idea is that we can add more than 2 gatherers later.

pkg/gather/gather.go
Sergey1011010: > I find these return values a little bit confusing TBH. I think returning just []Interface (maybe also change the name of the Interface?) would be more clear.

Well, they are 2 different types of gatherers, first list is processed always, every interval, second - only when it's there time.

> Just Friday's idea - What about adding the last processed time (and maybe the period as well?) to each gatherer and access it via some GetLastProcessedTime (via the Interface)?

Hm, yeah, great idea! The only thing is that we want to make sure that with the new approach some gatherers(now it's only cluster config gatherer) are processed every interval and get to every archive as it is now.

tremes: Yes I know that you wanted to distinguish the types that's why I suggested the idea. I guess I am saying let's distinguish them on the impl level of the `Interface`. 
And yes the `clusterconfig` one will be kinda special from this point of view.

pkg/gather/clusterconfig/0_gatherer.go
```diff
@@ -2,290 +2,92 @@ package clusterconfig
```
tremes, line 2: Let's rename this file please. `0_gatherer` is a bit meaningless IMHO :). We can go with `clusterconfig.go` (or `clusterconfig_gatherer.go`) or any better suggestions?

pkg/controller/operator.go
Sergey1011010: changed it

pkg/gather/gather.go
Sergey1011010: Done

pkg/gather/common/common.go
```diff
@@ -0,0 +1,35 @@
+package common
```
0sewa0, line 1: small nitpick: I feel this could be in the `gather` package, no need for a separate package

pkg/gather/gather.go
tremes: Sorry I missed this one. Yeah sounds good.

pkg/gather/common/common.go
Sergey1011010: There was a circular dependency since the types are used in gatherer, cluster config and workloads packages and this was the simplest way to avoid it.

0sewa0:
TLDR: It looks great, better then the current implementation, bit of an overkill currently.

Lots of changes, thankfully the majority are small simple changes to the gather-functions, making them a bit simpler which is a plus.
The bigger changes around how we actually run a Gather is actually great. 
The only "real problem" I might see is that it might be a bit of an overkill. Having all this extra logic just to run 1 gather-function separately feels a bit much.  Of course this might be only true right now, if in the future there will be more stuff to run in different periods then this change is a way to do it. 
We should also not forget, if we chose to go down this path then we will have 3 distinct ways of running a Gather in the last 3 versions of Openshift. (4.6: the "original" way; 4.7: 1.refactor; 4.8: this) Change is good and all, but make sure that it wont be 4 distinct ways when time comes for 4.9. :D 

(also run it against a cluste-bot cluster, it worked without any issues :+1: )

Sergey1011010:
@0sewa0, I can backport this change little later, so we'll have only one implementation

pkg/controller/periodic/periodic.go
```diff
@@ -6,42 +6,58 @@ import (
...
+// and flushes the recorder to create archives
 type Controller struct {
-	configurator Configurator
-	recorder     recorder.FlushInterface
-	gatherers    map[string]gather.Interface
-	statuses     map[string]*controllerstatus.Simple
+	configurator                 configobserver.Configurator
+	recorder                     recorder.FlushInterface
+	gatherers                    []gather.Interface
+	gatherersLastProcessingTimes map[string]time.Time
```
tremes, line 27: Do we need this map? Sorry, maybe I wasn't eplicit in my previous review. My thinking was roughly as follows:
- add the new time attribute to the `Gatherer` struct
- access the time by the "getter" method via the `Interface` (as you already updated it)
In other words it's about merging `Interface` and `LowPriorityGatherer` (btw this is not a good name IMO) into one. What do you think? 
I guess it would require also creating a setter method (to set the last processing time), but you will know better.

pkg/controller/periodic/periodic.go
```diff
@@ -98,15 +114,52 @@ func (c *Controller) Gather() {
...
 		_ = wait.ExponentialBackoff(backoff, func() (bool, error) {
+			name := gatherer.GetName()
 			start := time.Now()
-			err := c.runGatherer(name)
+
+			ctx, cancel := context.WithTimeout(context.Background(), c.configurator.Config().Interval/2)
+			defer cancel()
+
+			klog.V(4).Infof("Running %s gatherer", gatherer.GetName())
+			var err error
```
tremes, line 154: It looks like we don't need this line :)

pkg/gather/clusterconfig/clusterconfig_gatherer.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (g *Gatherer) GetName() string {
+	return "clusterconfig"
+}
+
+func (g *Gatherer) GetGatheringFunctions() map[string]common.GatheringClosure {
+	result := make(map[string]common.GatheringClosure)
+
+	for funcName, function := range gatheringFunctions {
+		function := function
```
tremes, line 82: It looks like we don't need this line :)

pkg/gather/clusterconfig/clusterconfig_gatherer.go
```diff
@@ -0,0 +1,94 @@
...
+	gatherKubeConfig        *rest.Config
+	gatherProtoKubeConfig   *rest.Config
+	metricsGatherKubeConfig *rest.Config
+	anonymizer              *anonymization.Anonymizer
+}
+
+type GathererFuncPtr = func(*Gatherer, context.Context) ([]record.Record, []error)
+
+var gatheringFunctions = map[string]common.GatheringFunction{
+	"pdbs":                              common.ImportantFunc((*Gatherer).GatherPodDisruptionBudgets),
```
tremes, line 24: `(*Gatherer).GatherPodDisruptionBudgets` hmmm....it's a bit confusing (at least to me), but I guess I don't have any better idea.

pkg/controller/periodic/periodic.go
Sergey1011010: But then we'd need to repeat this logic in each gatherer. Maybe not a big deal and we can add a few more methods to LowPriorityGatherer, but merging Interface and LowPriorityGatherer into one doesn't feel right to me, cuz then we'd have some useless code in the gatherers that aren't special. 

> btw this is not a good name IMO

Any better ideas?

pkg/controller/periodic/periodic.go
Sergey1011010: No idea why it's here :D

pkg/gather/clusterconfig/clusterconfig_gatherer.go
Sergey1011010: What do you mean?

pkg/gather/common/common.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+}
+
+// FailableFunc creates an object describing a gathering function that can fail
+func FailableFunc(function interface{}) GatheringFunction {
+	return GatheringFunction{
+		CanFail:  true,
+		Function: function,
+	}
+}
```
tremes, line 28: These three (`GatheringFunction`, `ImportantFunc`, `FailableFunc`) seem to be used in the clusterconfig gatherer. Maybe it would make sense to move them there and change the `GatheringFunction` to:
```
type GatheringFunction struct {
	CanFail  bool
	Function GathererFuncPtr
}
```

pkg/gather/clusterconfig/clusterconfig_gatherer.go
Sergey1011010: We actually do https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables otherwise all the closures will reference one function.

pkg/controller/periodic/periodic.go
tremes: Which logic exactly? I was thinking just about the accessing the new attribute. Hmm yeah my thinking was to have `GetLastProcessedTime` and comparing it to `GetDesiredGatheringPeriod`.... but it's probably true that it's useless for the `clusterconfig` gatherer.

tremes: Maybe `CustomPeriodGatherer` or something like that?

Sergey1011010: Implemented a little different thing where custom period gatherers just return true or false if they want to be processed and they are responsible for tracking time. It's little easier to unit test. https://github.com/openshift/insights-operator/pull/399/files#diff-095e9955b4ce960ffb4ea94ee251d09818bc2d6a76b8f4b3c5414ee419ed82f1R42 WDYT?

tremes: @Sergey1011010 Yes I like it. Thank you.

pkg/gather/clusterconfig/clusterconfig_gatherer.go
tremes: Well I know what is the purpose, but I don't know how to call this declaration :)  It's not type casting, is it? It looks like that I don't know Golang at all... :)

pkg/gather/common/common.go
tremes: Let me reopen this one.

pkg/gather/workloads/workloads_info.go
```diff
@@ -392,7 +296,7 @@ func gatherWorkloadImageInfo(ctx context.Context, imageClient imageclient.ImageI
 // TODO: some optimization in very large namespaces with diverse shapes may be
 //   necessary
 func workloadPodShapeIndex(shapes []workloadPodShape, shape workloadPodShape) int {
-	for i := len(shapes) - 1; i >= 0; i-- {
+	for i := len(shapes) - 1; i >= 0; i-- { // TODO: fix unreachable code
```
tremes, line 299: This was already fixed it AFAIK

pkg/gather/workloads/workloads_info.go
```diff
@@ -475,6 +374,7 @@ func idForImageReference(s string) string {
 	return ref.ID
 }
 
+// TODO: fix unused function
```
tremes, line 377: Same here.

pkg/gather/common/common.go
```diff
@@ -0,0 +1,14 @@
+package common
```
tremes, line 1: Do we need this new package (given the fact we just declare one new type here?)?

pkg/gather/common/common.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// GatheringClosure is a struct containing a closure each gatherer returns
+// it also contains CanFail field showing if we should just log the failures
+type GatheringClosure struct {
+	Run     func(context.Context) ([]record.Record, []error)
+	CanFail bool
+}
```
tremes, line 14: I am still puzzled by the fact that this is very similar to the `gatheringFunction` type. I think it would be nice to have only one type (but it would probably mean to somehow share the `Gatherer` type), but not sure if it's feasible.

tremes: Hmm it's probably not a good idea to try to reuse one `Gatherer` type.... :-/

pkg/gather/clusterconfig/clusterconfig_gatherer.go
tremes: Wouldn't be better to keep the gathering function eg ` GatherPodDisruptionBudgets(g *Gatherer, ctx context.Context)` as we have it right now?

pkg/controller/periodic/periodic.go
```diff
@@ -98,15 +106,46 @@ func (c *Controller) Gather() {
...
+
+	var allFunctionReports []gather.GathererFunctionReport
+
+	for _, gatherer := range gatherersToProcess {
 		_ = wait.ExponentialBackoff(backoff, func() (bool, error) {
+			name := gatherer.GetName()
 			start := time.Now()
-			err := c.runGatherer(name)
+
+			ctx, cancel := context.WithTimeout(context.Background(), c.configurator.Config().Interval/2)
```
tremes, line 137: Lol... so I was experimenting with setting some short timeout here (e.g 10s) and the log kinda exploded....especially due to two following gatherers:

- olm_operators
-  service_accounts

pkg/gather/common/common.go
rluders: I'm a little bit afraid that to have this `common` package inside gather can be a little bit confusing, 'cause the other folders are gatherer folders. If we need this package, I'll definitely suggest moving it to the `pkg` root, or even reorganize the gatherers folder structures by creating some `pkg/gatherers` folder.

rluders (changes requested):

pkg/gather/clusterconfig/clusterconfig_gatherer.go
Sergey1011010: It's basically a pointer to a method. Well, there's no differences from the point of how it works, but I think it makes more sense to convert the functions to methods cuz they all use the same data from the gatherer.

pkg/gather/common/common.go
Sergey1011010: Well, moving `GatheringClosure` to the `gather` package would still cause circular dependency. Maybe we can reorganize things somehow differently.

pkg/gather/common/common.go
Sergey1011010: Yeah, gatherers have different types so we have to hide it from the calling code somehow and using the same type would make gatherers not flexible.

pkg/gather/clusterconfig/clusterconfig_gatherer.go
```diff
@@ -0,0 +1,116 @@
...
+type gathererFuncPtr = func(*Gatherer, context.Context) ([]record.Record, []error)
+
+// gatheringFunction describes a gathering function
+type gatheringFunction struct {
+	CanFail  bool
+	Function gathererFuncPtr
+}
+
+// importantFunc creates an object describing a gathering function that canNOT fail
+func importantFunc(function gathererFuncPtr) gatheringFunction {
```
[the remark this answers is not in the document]
Sergey1011010: They were there, but I moved them to clusterconfig cuz right now we aren't using them anywhere else.

pkg/gather/workloads/workloads_info.go
Sergey1011010: fixed

pkg/gather/workloads/workloads_info.go
Sergey1011010: fixed

pkg/gather/clusterconfig/clusterconfig_gatherer.go
tremes: OK thanks. Yeah there's no difference. We don't use them like methods anyway. It's minor and up to you. It seemed to me more straightforward without the pointer.

pkg/gather/common/common.go
tremes: Hmm true...any suggestions how to reorganize? I am running out of ideas....or maybe just another name for the package?

pkg/controller/periodic/periodic.go
tremes: I was experimenting with this again and maybe I would remove the extra logging statement in https://github.com/openshift/insights-operator/pull/399/files#diff-bdd45c95687562d23e5fdd43dea644d8fb86c95354192cedd72835743ba9fc0fR104. WDYT? I think the error should be obvious from a particular gatherer function and we shouldn't log it again....

I noticed it's almost the same now in https://github.com/openshift/insights-operator/blob/master/pkg/gather/clusterconfig/0_gatherer.go#L216, but the verbosity level is 5.

pkg/gather/common/common.go
rluders: Let me take another look into it... Normally I do have some good ideas about folder organization. :rofl:

rluders: How about this: https://github.com/rluders/insights-operator/tree/support-of-gatherers-with-different-periods

pkg/controller/periodic/periodic.go
Sergey1011010: But where else do we log it?

pkg/gather/common/common.go
Sergey1011010: @rluders looks great

pkg/controller/periodic/periodic.go
tremes: The particular gatherer should log its own error IMO and it should also be visible in the metadata.

Sergey1011010: But if we forget to log the error in the gatherer we're only gonna get it in the metadata

tremes: Yes, what I am trying to say is that it should be gatherer responsibility (i.e. we shouldn't forget). The metadata is most important for me, because we usually don't have the IO log in our archive.

Sergey1011010: Well, from my point of view it's better to do it once in the calling code than many times in gatherers. But I agree that metadata is more important than logs for us.

pkg/gather/common/common.go
Sergey1011010: done

pkg/gather/gather.go
```diff
@@ -0,0 +1,244 @@
...
+// which has the following structure: []string {
+//   "clusterconfig/container_images",
+//   "clusterconfig/nodes",
+//   "clusterconfig/authentication",
+//   "othergatherer/some_function",
+// } where each item consists of a gatherer name and a function name split by a slash.
+// If there's a string "ALL", we enable everything and return the first parameter as true,
+// otherwise it will be false and the second parameter will contain function names
+func getListOfEnabledFunctionForGatherer(gathererName string, allFunctionsList []string) (bool, []string) {
+	if utils.StringInSlice(AllGatherersConst, allFunctionsList) {
```
rluders, line 203: Is there any way to disable the gatherer as well? Right now we only support disabling the gathers.

Also... Wouldn't it be better if we assume that we don't specify the gathers per configuration we always run all the gathers? Let me try to explain it.

Right now, we should have at the configuration file the value `ALL` like this:

```yml
gather:
  - ALL
```

And If the configuration is missing we run all the gathers? Also, don't we want to specify the gathers per gatherer when disabling it?

```yml
gather:
    clusterconfig:
      - image_registries
      - image_pruners
      - install_plans
```
In this case we will run only the `clusterconfig` with the listed gathers.

Sergey1011010: Right now we specify only enabled gathering functions, we can enable the whole gatherer, but we'd need to mention every function. I thought of making it lil more flexible, but decided to not make the PR even bigger. I'd vote for possibly improving it later.

pkg/gatherers/clusterconfig/sap_vsystem_iptables_logs.go
```diff
@@ -29,41 +30,36 @@ import (
...
-		return
+		return nil, []error{err}
 	}
 
-	datahubsList, err := dynamicClient.Resource(datahubGroupVersionResource).List(g.ctx, metav1.ListOptions{})
+	datahubsList, err := dynamicClient.Resource(datahubGroupVersionResource).List(ctx, metav1.ListOptions{})
 	if errors.IsNotFound(err) {
-		c <- gatherResult{nil, nil}
-		return
+		klog.Info("SAP resources weren't found")
```
tremes, line 46: It would be probably better to mention the `datahubGroupVersionResource` in the info log, but we can do this anytime.

pkg/gather/gather.go
rluders: Sure! We should have another PR to implement it. I'm just rising the discussion here, and maybe wasn't the best place, but it was easy to explain by adding the comment here.

pkg/gather/tasks_processing_test.go
```diff
@@ -0,0 +1,202 @@
...
+				{Name: "task_2_record_0"},
+				{Name: "task_2_record_1"},
+			},
+		},
+	})
+}
+
+// TODO: consider removing. The idea of the test is to check that execution is not sequential,
+// but the problem is that it's not guaranteed that it will take some specific time.
+// Maybe it should be an integration test
```
tremes, line 74: +1 for removing and creating an integration test

pkg/gather/tasks_processing_test.go
```diff
@@ -0,0 +1,202 @@
...
+	results[0].TimeElapsed = 0
+	assert.Equal(t, results, []GatheringFunctionResult{
+		{
+			Panic:        "test panic",
+			IgnoreErrors: false,
+		},
+	})
+}
+
+func Test_HandleTasksConcurrently_CannotFail_Panic(t *testing.T) {
```
tremes, line 123: I guess the name of this test is swapped with the test name above.

Sergey1011010: fixed

pkg/gather/tasks_processing_test.go
Sergey1011010: Replaced with a todo :)

tremes:
I played with it and reviewed several times. Thanks for your patience :)
/lgtm

rluders (approved):
Tested and it seems to work as expected.
/lgtm

rluders:
Just rebase. Adding tag again.
/lgtm

PR #407, opened 2021-04-21
Linting fixes in gather package

0sewa0:
Goes through the gather/clusterconfig package to fix the linting errors.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## References


tisnik (approved):
Seems legit

pkg/gather/clusterconfig/authentications.go
```diff
@@ -1,3 +1,4 @@
+//nolint: dupl
```
tremes, line 1: What is the reason for this pls?

0sewa0: code duplication, some parts of the code is duplicated mostly the place where the clients get created, but IMO its cleaner this way, (for backporting)

tremes: Ahh OK thanks. I needed to run the linting to understand it ... :)

pkg/gather/clusterconfig/certificate_signing_requests.go
```diff
@@ -98,8 +101,8 @@ func FromCSRs(requests *certificatesv1api.CertificateSigningRequestList) *CSRs {
 
 func (c *CSRs) Anonymize() *CSRs {
 	res := &CSRs{}
-	for _, r := range c.Requests {
-		af := anonymizeCSR(&r)
+	for _, r := range c.Requests { //nolint: gocritic
+		af := anonymizeCSR(&r) //nolint: gosec
```
tremes, line 105: Maybe we could use indexing? As SO told me https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop

pkg/gather/clusterconfig/gather_logs.go
```diff
@@ -44,6 +44,7 @@ type logMessagesFilter struct {
 //   - logFileName sets the name of the file to save the logs to.
 //
 // Location of the logs is `config/pod/{namespace}/logs/{podName}/{fileName}.log`
+//nolint: gocritic, unparam
```
tremes, line 47: Same for the `gocritic`. Indexing in the for loops seems to fix it.

tremes:
Much less `gocritic` usage. I think we should continue to work on removing in the future, but looks good for now. Thanks!
/lgtm

PR #409, opened 2021-04-23
Bug 1952906: Add a few tests to configobserver_test.go

0sewa0:
Adds a few extra tests for the configoberver

## Categories

- [ ] Bugfix
- [ ]Enhancement
- [ ] Backporting

## Sample archive

- no change

## Documentation

- no change

## Unit Tests

- see `pkg/config/configobserver/configobserver_test.go`

## Privacy

### DIFF
--- a/pkg/config/configobserver/configobserver_test.go
+++ b/pkg/config/configobserver/configobserver_test.go
@@ -19,6 +19,7 @@ import (
 	clienttesting "k8s.io/client-go/testing"
 )
 
+//nolint: lll, funlen
 func Test_ConfigObserver_ChangeSupportConfig(t *testing.T) {
 	var cases = []struct {
 		name      string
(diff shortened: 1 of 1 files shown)


rluders (changes requested):
Shouldn't const, vars, and structs be defined at the beginning of the file?

0sewa0:
the CI fails because of the CI's golang version doesn't match the golang version used for IO and the unit-test I added has a behavior that is different on the two versions. (CI: 1.14 -- go.mod: 1.16) the error has to do with the way an `error`  is converted to `string`

pkg/config/configobserver/configobserver_test.go
```diff
@@ -40,6 +41,50 @@ func Test_ConfigObserver_ChangeSupportConfig(t *testing.T) {
...
+			config: map[string]*corev1.Secret{
+				pullSecretKey: {Data: map[string][]byte{
+					".dockerconfigjson": nil,
+				}},
+				supportKey: {Data: map[string][]byte{
+					"reportPullingDelay": []byte("every second"),
+				}},
+			},
+			expConfig: &config.Controller{}, // it only produces a warning in the log
+		},
```
tremes, line 87: It looks like we have this one twice.

tremes: Should it be for `reportPullingTimeout` ?

pkg/config/configobserver/configobserver_test.go
```diff
@@ -97,6 +180,62 @@ func Test_ConfigObserver_ChangeSupportConfig(t *testing.T) {
...
+	// 2. config update
+	err = c.retrieveConfig(context.TODO())
+	if err != nil {
+		t.Fatalf("Unexpected error %s", err)
+	}
+	// Check if unsubscribe worked. ie: no new event on the channel
+	if len(configCh) != 1 {
+		t.Fatalf("The closing function failed to unsubscribe from the config change event. len(configCh)==%d", len(configCh))
+	}
+}
```
tremes, line 237: What about to do this test in similar fashion (declaring the testcases and iterating over them) as the above test? There are two cases (if I am not mistaken) and the code looks repeating.

pkg/config/configobserver/configobserver_test.go
0sewa0: yep, nice catch :)

pkg/config/configobserver/configobserver_test.go
0sewa0: I wanted to test the scenario:
1. Subscribe to config changes
2. Config change happens ==> Event received
3. Unsubscribe from config changes
4. Config change happens ==> Event NOT received

And yeah I could probably test these in 2 iterations, but I personally don't like the "declaring the testcases and iterating over them" for few test cases (<4 or so), its too messy for a small gain. 

Also I feel that in this case having these steps happen on the same config.Controller is more "realistic".

tremes: Fair points. Yeah I don't like the repeating code, but it's not that catastrophic in this case and in the end it's only test....

pkg/config/configobserver/configobserver_test.go
```diff
@@ -19,6 +19,17 @@ import (
 	clienttesting "k8s.io/client-go/testing"
 )
 
+type kubeClientResponder struct {
+	clsetfake.Clientset
+}
+
+var _ kubernetes.Interface = (*kubeClientResponder)(nil)
```
tremes, line 26: I think we can remove this line right

tremes:
Verified :), reviewed... :) 
/lgtm

PR #411, opened 2021-04-26
Bug 1953743: Correctly detect workload shapes when gathering instead of only detecting the first match

smarterclayton:
The find algorithm was returning no match unless the first shape
matched, which caused us to not collapse shapes correctly.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting


smarterclayton:
@0sewa0 PTAL

0sewa0:
@smarterclayton there is another PR with this change and its going through the last CI tests: https://github.com/openshift/insights-operator/pull/407

Issue #417, opened 2021-04-30
Pull Insights report retry mechanism is broken


joselsegura:
When Insights Operator uploads a new archive, it waits for a configured amount of time and then starts to retrieve the report generated by the Insights Results Smart Proxy service.

If something goes wrong, it should be a retry mechanism that allow IO to try again after some time, but it doesn't work properly with some errors.

For example, if the Smart Proxy service returns a 404 (which is something very usual after the first archive upload), the retry mechanism doesn't try to get it again after some time. (The error is very likely to be in this code block: https://github.com/openshift/insights-operator/blob/7c60684c556da795663c602e618ae4bf7ed6621e/pkg/insights/insightsreport/insightsreport.go#L93).

This problem is impacting users, as they won't be able to see Insights results for their clusters in the OCP WebConsole until the second archive is uploaded

tremes:
This issue can be closed, but it looks like I am not able to do that...... :-/

joselsegura:
I cannot assign it to you, but it is already reviewed, resolved and merged, so I'll close it.

TYVM!

PR #418, opened 2021-05-03
Bug 1958759: #417 insights report - add basic retry logic in case of 404

tremes:
This is adding retry logic for the insights report pulling - addressing https://github.com/openshift/insights-operator/issues/417 and also updating Policy API version to v1.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

No new data

## Documentation

No update

## Unit Tests

No new test

## Privacy

## References

### DIFF
--- a/pkg/insights/insightsreport/insightsreport.go
+++ b/pkg/insights/insightsreport/insightsreport.go
@@ -165,14 +165,13 @@ func (r *Gatherer) RetrieveReport() {
 			}
 
 			firstPullDone = true
-			if retryCounter < retryThreshold {
-				t := wait.Jitter(config.ReportMinRetryTime, 0.1)
-				klog.Infof("Reseting the delay timer to retry in %s again", t)
-				delayTimer.Reset(t)
-				retryCounter++
-			} else {
+			if retryCounter >= retryThreshold {
 				return
 			}
+			t := wait.Jitter(config.ReportMinRetryTime, 0.1)
+			klog.Infof("Reseting the delay timer to retry in %s again", t)
+			delayTimer.Reset(t)
+			retryCounter++
 		case <-timeoutTimer.C:
 			// timeout, ends
 			if !delayTimer.Stop() {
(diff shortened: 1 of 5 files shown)


pkg/insights/insightsreport/insightsreport.go
```diff
@@ -152,11 +165,14 @@ func (r *Gatherer) RetrieveReport() {
 			}
 
 			firstPullDone = true
-			if !delayTimer.Stop() {
-				<-delayTimer.C
```
tremes, line 156: It's still not clear to me whether we need to call this @joselsegura. This part of code was almost never (with waiting for clusterversion exception) called before AFAIK. 
My thinking is following. We already drained the channel in this moment and thus the `delayTimer.Stop()` returns `false` (see https://golang.org/pkg/time/#Timer.Stop), which leads to blocking call of `<-delayTimer.C`. Maybe it should be:
```
if delayTimer.Stop() {
	<-delayTimer.C
}
```
or maybe I am completely wrong....

joselsegura: TBH I wrote it a long time ago and I'm not 100% sure, but according to the example in the link to Golang `time` library, the `Stop` call will return "false" «if the timer has already expired or been stopped. Stop does not close the channel, to prevent a read from the channel succeeding incorrectly.»

So I think this logic is correct: if `Stop` returns "true", you are done with it. If it returns "false", you need to read from the channel to ensure anybody else is consuming it (if I am understanding the library reference correctly!).

It is probably a very rare corner case, so probably your solution will work much better

tremes: The documentation for `Reset` describes it even better IMO https://golang.org/pkg/time/#Timer.Reset and I think we don't need to call/check `Stop`, because it will always return `false` in this situation, because the timer has already expired (as per doc).

rluders:
It looks pretty straightforward, so just reading the code it looks good.
/lgtm

PR #419, opened 2021-05-05
Bug 1958790: Store translation table in a secret

0sewa0:
The anonymizer has a translation-table about the obfuscated IP addresses. As of right now its inaccessible, so this change would store the  translation-table in a `Secret` in the `openshift-insights` namespace.
On the Web UI you can easily find the secrets under `Workloads` -> `Secrets`, the name of the secret is `obfuscation-translation-table`.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

- no

## Documentation

- Added docstrings, user facing documentation is a separate task.

## Unit Tests

- Added new unit tests to `pkg/anonymization/anonymizer_test.go`

## Privacy

### DIFF
--- a/manifests/03-clusterrole.yaml
+++ b/manifests/03-clusterrole.yaml
@@ -179,7 +179,7 @@ rules:
   verbs:
     - get
     - list
-    - watch 
+    - watch
 - apiGroups:
     - operators.coreos.com
   resources:
@@ -213,6 +213,38 @@ subjects:
 - kind: ServiceAccount
   namespace: openshift-insights
   name: gather
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: insights-operator-obfuscation-secret
+  namespace: openshift-insights
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - delete
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: insights-operator-obfuscation-secret
+  namespace: openshift-insights
+roleRef:
+  kind: Role
+  name: insights-operator-obfuscation-secret
+subjects:
+- kind: ServiceAccount
+  name: gather
+  namespace: openshift-insights
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -263,8 +263,7 @@ func (anonymizer *Anonymizer) ObfuscateIP(ipStr string) string {
 }
 
 // Stores the translation table in a Secret in the openshift-insights namespace.
-// The actual data is stored in the StringData protion of the Secret.
-// Uses the Apply mechanism so it creates the secret if not present or updates it when present.
+// The actual data is stored in the StringData portion of the Secret.
 func (anonymizer *Anonymizer) StoreTranslationTable() *corev1.Secret {
 	if len(anonymizer.translationTable) == 0 {
 		return nil
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -58,6 +58,7 @@ var (
 	TranslationTableSecretName = "obfuscation-translation-table" //nolint: gosec
 	secretAPIVersion           = "v1"
 	secretKind                 = "Secret"
+	secretNamespace            = "openshift-insights"
 )
 
 type subnetInformation struct {
@@ -135,7 +136,7 @@ func NewAnonymizerFromConfigClient(
 		return nil, err
 	}
 
-	secretsClient := kubeClient.CoreV1().Secrets("openshift-insights")
+	secretsClient := kubeClient.CoreV1().Secrets(secretNamespace)
 
 	if installConfig, exists := clusterConfigV1.Data["install-config"]; exists {
 		networkRegex := regexp.MustCompile(Ipv4NetworkRegex)
@@ -263,6 +264,9 @@ func (anonymizer *Anonymizer) ObfuscateIP(ipStr string) string {
 	return "::"
 }
 
+// Stores the translation table in a Secret in the openshift-insights namespace.
+// The actual data is stored in the StringData protion of the Secret.
+// Uses the Apply mechanism so it creates the secret if not present or updates it when present.
 func (anonymizer *Anonymizer) StoreTranslationTable() *corev1.Secret {
 	if len(anonymizer.translationTable) == 0 {
 		return nil
@@ -290,9 +294,11 @@ func (anonymizer *Anonymizer) StoreTranslationTable() *corev1.Secret {
 		klog.Errorf("Failed to create/update the translation table secret. err: %s", err)
 		return nil
 	}
+	klog.V(3).Infof("Created/Updated %s secret in %s namespace", TranslationTableSecretName, secretNamespace)
 	return result
 }
 
+// Resets the translation table, so that the translation table of multiple gathers wont mix toghater.
 func (anonymizer *Anonymizer) ResetTranslationTable() {
 	anonymizer.translationTable = make(map[string]string)
 }
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -55,7 +55,7 @@ const (
 )
 
 var (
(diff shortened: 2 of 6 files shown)


pkg/anonymization/anonymizer.go
```diff
@@ -49,6 +53,12 @@ const (
 		"some data won't be anonymized(ipv4 and cluster base domain). The error is %v"
 )
 
+var (
+	TranslationTableSecretName = "translation"
```
Sergey1011010, line 57: I'd name it more precisely, something like obfuscation-translation-table or whatever

pkg/anonymization/anonymizer.go
```diff
@@ -231,6 +245,39 @@ func (anonymizer *Anonymizer) ObfuscateIP(ipStr string) string {
...
+	}
+
+	applyOptions := metav1.ApplyOptions{
+		Force:        true,
+		FieldManager: "insights-operator",
+	}
+
+	result, err := anonymizer.secretsClient.Apply(context.TODO(), secret, applyOptions)
+	if err != nil {
+		klog.Errorf("Failed to create/update the translation table secret. err: %s", err)
```
Sergey1011010, line 272: Is the result still valid if an error happened?

pkg/anonymization/anonymizer.go
```diff
@@ -231,6 +245,39 @@ func (anonymizer *Anonymizer) ObfuscateIP(ipStr string) string {
 	return "::"
 }
 
+func (anonymizer *Anonymizer) StoreTranslationTable() *corev1.Secret {
+	if len(anonymizer.translationTable) == 0 {
+		return nil
+	}
+	defer anonymizer.ResetTranslationTable()
```
Sergey1011010, line 252: Why do we need to reset it btw?

Sergey1011010:
Looks great, would be fine to have a simple test of the behaviour, but up to you

pkg/anonymization/anonymizer.go
0sewa0: so that only the latest archive's IPs are present in it. Maybe its not clear for me how the anonymizer uses the translation tables, as I understand it we never prune the translation table, we just check if an IP is already in it so theoretically the translation table constantly grows even if some IPs in it are not used anymore.

pkg/anonymization/anonymizer.go
0sewa0: it returns the secret you wanted to create, which is not really useful in our case, so I will just return `nil`

pkg/anonymization/anonymizer.go
```diff
@@ -49,6 +53,13 @@ const (
 		"some data won't be anonymized(ipv4 and cluster base domain). The error is %v"
 )
 
+var (
+	TranslationTableSecretName = "obfuscation-translation-table" //nolint: gosec
```
tremes, line 57: Is there need to export this variable?

pkg/anonymization/anonymizer.go
```diff
@@ -231,6 +246,45 @@ func (anonymizer *Anonymizer) ObfuscateIP(ipStr string) string {
...
+func (anonymizer *Anonymizer) StoreTranslationTable() *corev1.Secret {
+	if len(anonymizer.translationTable) == 0 {
+		return nil
+	}
+	defer anonymizer.ResetTranslationTable()
+
+	secret := &applycorev1.SecretApplyConfiguration{
+		TypeMetaApplyConfiguration: applymetav1.TypeMetaApplyConfiguration{
+			Kind:       &secretKind,
+			APIVersion: &secretAPIVersion,
```
tremes, line 261: Is this required pls? It doesn't seem right to hardcode the API version.....

0sewa0: Yes
```
Failed to create/update the translation table secret. err: Incorrect version specified in apply patch. Specified patch version: , expected: v1
```

We already hardcode API version in the `mainfests`, and IMO we should just update the API version on the fly

pkg/anonymization/anonymizer.go
```diff
@@ -231,6 +246,45 @@ func (anonymizer *Anonymizer) ObfuscateIP(ipStr string) string {
...
+	result, err := anonymizer.secretsClient.Apply(context.TODO(), secret, applyOptions)
+	if err != nil {
+		klog.Errorf("Failed to create/update the translation table secret. err: %s", err)
+		return nil
+	}
+	klog.V(3).Infof("Created/Updated %s secret in %s namespace", TranslationTableSecretName, secretNamespace)
+	return result
+}
+
+// Resets the translation table, so that the translation table of multiple gathers wont mix toghater.
```
Sergey1011010, line 283: They would actually mix now. Just tested by faking the translation table right before saving and got all the previous values in the secret and my fake ones.

0sewa0: Nice catch, I replaced the `apply` with a `delete` and `create`, it should fix this issue.

pkg/gather/gather_test.go
```diff
@@ -182,7 +182,7 @@ func Test_StartGatheringConcurrently(t *testing.T) {
 		},
 	})
 
-	resultsChan, err = startGatheringConcurrently(context.Background(), gatherer, []string{
+	resultsChan, _ = startGatheringConcurrently(context.Background(), gatherer, []string{
```
Sergey1011010, line 185: It's better to add assert.NoError instead of ignoring it

pkg/gather/gather_test.go
```diff
@@ -41,7 +41,7 @@ func Test_GetListOfEnabledFunctionForGatherer(t *testing.T) {
 	assert.False(t, all)
 	assert.Empty(t, functions)
 
-	all, functions = getListOfEnabledFunctionForGatherer("", list)
+	_, functions = getListOfEnabledFunctionForGatherer("", list)
```
Sergey1011010, line 44: I guess it should be assert.False(t, all)

Sergey1011010:
Just tested and it doesn't work anymore 
```
E0519 10:34:25.607470   13606 anonymizer.go:296] Failed to create the translation table secret. err: secrets is forbidden: User "system:serviceaccount:openshift-insights:gather" cannot create resource "secrets" in API group "" in the namespace "openshift-insights"
```
Looks like it's because of this change https://github.com/openshift/insights-operator/pull/424 which makes obfuscation initialize using gather serviceaccount instead of operator. So looks like you need to create secret client from a different kubeconfig from pkg/controller/gather_job.go and pkg/controller/operator.go, the one with operator service account.

0sewa0:
@Sergey1011010 I think giving the `gather` service account the permission to create/delete secrets in the `openshift-insights` namespace is simpler, WDYT ?

Sergey1011010:
@0sewa0 well, you can just pass another config to anonymization and use it for secrets client, and also `gather` service account probably shouldn't have it. It seems like it should only have all the read only permissions for gathering.

manifests/03-clusterrole.yaml
```diff
@@ -214,6 +214,46 @@ subjects:
...
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - ""
+  verbs:
+  - get
+  - create
+  - update
+  - delete
```
tremes, line 238: Can't we use the existing role https://github.com/openshift/insights-operator/blob/master/manifests/03-clusterrole.yaml#L274 and just extend its rules (to be able to work with secrets)?

pkg/anonymization/anonymizer.go
```diff
@@ -249,6 +262,49 @@ func (anonymizer *Anonymizer) ObfuscateIP(ipStr string) string {
 	return "::"
 }
 
+// Stores the translation table in a Secret in the openshift-insights namespace.
+// The actual data is stored in the StringData protion of the Secret.
```
tremes, line 266: `protion` typo

pkg/anonymization/anonymizer.go
```diff
@@ -249,6 +262,49 @@ func (anonymizer *Anonymizer) ObfuscateIP(ipStr string) string {
...
+// Uses the Apply mechanism so it creates the secret if not present or updates it when present.
```
tremes, line 267: This doesn't seem to be true anymore

manifests/03-clusterrole.yaml
0sewa0: that is for the operator SA, this is for the gather SA

tremes:
I experimented with this one quite a lot (building custom image and deploying it in the cluster) and I found a new bug, which I am about to open a new PR for.  
/lgtm

PR #421, opened 2021-05-07
Bug 1961067: Collect full pod log for stack traces

rluders:
This PR adds the capability to collect full container logs if the partial log contains possible stack traces.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

None.

## Documentation

None.

## Unit Tests

- `pkg/gatherers/clusterconfig/operators_pods_and_events_test.go`

## Privacy

## Changelog

None.

## References

https://issues.redhat.com/browse/CCXDEV-4600
https://bugzilla.redhat.com/show_bug.cgi?id=1961067

### DIFF
--- a/pkg/gatherers/clusterconfig/const.go
+++ b/pkg/gatherers/clusterconfig/const.go
@@ -21,6 +21,8 @@ var (
 	registryScheme = runtime.NewScheme()
 	// logTailLines sets maximum number of lines to fetch from pod logs
 	logTailLines = int64(100)
+	// logTailLinesLong sets the maximum number of lines to fetch from long pod logs
+	logTailLinesLong = int64(400)
 
 	defaultNamespaces           = []string{"default", "kube-system", "kube-public"}
 	datahubGroupVersionResource = schema.GroupVersionResource{
--- a/pkg/gatherers/clusterconfig/operators_pods_and_events.go
+++ b/pkg/gatherers/clusterconfig/operators_pods_and_events.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


pkg/gather/clusterconfig/operators_pods_and_events.go
```diff
@@ -250,33 +255,75 @@ func getContainerLogs(ctx context.Context,
...
+	client corev1client.CoreV1Interface,
+	pod *corev1.Pod,
+	name string,
+	isPrevious bool,
+	buf *bytes.Buffer,
+	bufferSize *int64) (string, error) {
+	if buf == nil {
+		buf = &bytes.Buffer{}
+	}
+	err := fetchPodContainerLog(ctx, client, pod, buf, name, isPrevious, bufferSize)
```
rluders, line 298: Note: I know that this `fetchPodContainerLog` by default uses the `logTailLines` that will block the committed changes to work properly. I didn't change it yet, 'cause I'm exploring some ideas around it. :)

pkg/gather/clusterconfig/operators_pods_and_events.go
```diff
@@ -37,6 +38,9 @@ type CompactedEventList struct {
 	Items []CompactedEvent `json:"items"`
 }
 
+// Used to detect the possible stack trace on logs
+var stackTraceRegex = regexp.MustCompile(`((.*)(/{0,9}){0,9}).([0-9]{0,9})`)
```
tremes, line 42: I am not very good in the regex expressions TBH, but this seems too wide. It's matching logs without any stacktrace from what I've seen.

tremes: You can try to degrade the monitoring operator (as we do in our tests) and run your code against it. You will see that a lot of logs match your expression, but there is no stacktrace present.

natiiix (approved):
/lgtm
I tested locally that the IO runs and generates archives without errors.
Could you please fix the PR description a bit? It contains the default paths from the template.

PR #422, opened 2021-05-10
Bug 1959650: Gather SDI-related MachineConfigs

natiiix:
This PR implements an enhancement gathering two MachineConfigs related to the SAP Data Intelligence workload. This enhancement will not be backported past v4.8.

**Description from Jira:**

> After discussions with a SAP team representative, it was determined that two MachineConfigs should be gathered from clusters running SDI. There are also relevant MachineConfigPools, which are already gathered by past data enhancements.

> The first MC is called `75-worker-sap-data-intelligence`, but on clusters schedulable master nodes, it may also appear as `75-master-sap-data-intelligence`.

> The second MC is more complicated because it is generated from other resources. It is called `99-sdi-generated-containerruntime` on the testing cluster, but its name may be different on other clusters. Instead of being filtered by name, the gatherer should look for a `sdi-pids-limit` owner reference.

**Implementation note:** Most SAP-related enhancements first check for the `datahubs` resource to check if the cluster is running an SDI payload and collect nothing if the resource is not present on the cluster. In the case of this enhancement, this check is intentionally omitted because there should be no way for any of the gathering conditions to be met on a non-SDI cluster and the MachineConfigs are a part of core configuration, which could be present even if the `datahubs` resource was missing.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

- `docs/insights-archive-sample/config/machineconfigs/75-worker-sap-data-intelligence.json`
- `docs/insights-archive-sample/config/machineconfigs/99-sdi-generated-containerruntime.json`

## Documentation

- `docs/gathered-data.md` - `SAPMachineConfigs` section

## Unit Tests

- `pkg/gather/clusterconfig/sap_machine_configs_test.go`

## Privacy

After a discussion with a SAP team representative, it was determined that the selected MachineConfigs should not contain any sensitive information under normal circumstances.

## Changelog

No.

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-4829
Bugzilla Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1959650


0sewa0:
unfortunately I don't have a SAP cluster so I can't really test it, but I went through the code and run the extensive unittest for this change and everything works as expected.
/lgtm

PR #423, opened 2021-05-12
Gather Profiling data (POC)

0sewa0:
A somewhat simple gatherer for gathering profiling data. (only from the apiserver for now) 

**IMPORTANT**: if you want to test it first look at: https://github.com/openshift/insights-operator/pull/423/files#diff-34a5d4667465a696c64a82b0febc3d9b2f373d1923ed1ecfb8330f26b78517ccR31

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample archive

- TODO

## Documentation

- TODO

## Unit Tests

- TODO

## Privacy

### DIFF
--- a/pkg/gatherers/profiling/profiling_gatherer.go
+++ b/pkg/gatherers/profiling/profiling_gatherer.go
@@ -1,0 +1,82 @@
+package profiling
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/openshift/insights-operator/pkg/gatherers"
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+const (
+	ProfileCPU      = "debug/pprof/profile"
+	ProfileTypeHeap = "debug/pprof/heap"
+)
+
+var profilingGathererPeriod = time.Hour * 12
+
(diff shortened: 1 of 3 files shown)


pkg/gatherers/profiling/profiling_gatherer.go
```diff
@@ -0,0 +1,82 @@
...
+
+// GET
+// /api/debug/pprof/profile?seconds=10
+func (g *Gatherer) GetProfiles(ctx context.Context, profile string, seconds int) ([]byte, error) {
+	profilingRESTClient, err := rest.RESTClientFor(g.gatherKubeConfig)
+	if err != nil {
+		klog.Warningf("Unable to load profiling client, profiling will not be collected: %v", err)
+		return nil, nil
+	}
+	data, err := profilingRESTClient.Get().AbsPath(profile).
```
tremes, line 74: Is this really profiling some API server (e.g kube-api)? Where or what is the full URL? I am probably missing something, but it seems to me as profiling of IO itself.

0sewa0: as far as I know this profiles the kube api-server (that is why you don't see any URL really, because it uses the "base" URL of the cluster, and adds the pprof things to the end to connect to the profiler), but this feature is postponed for 4.9 due to auth issues

tremes: hm interesting :) OK thanks for info

tremes:
The corresponding epic was dropped and there are no plans to work on it at the moment. 
/close

PR #425, opened 2021-05-13
Bug 1966416: Do not exceed archive size limit

tremes:
This enforces the data size limit and also fixes a bug with growing metadata file in case of gathering failures.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample archive

TODO we will need to update the metadata json 

## Documentation

No doc update

## Unit Tests

Some tests added in:
- ` pkg/controller/periodic/periodic_test.go`
- `pkg/recorder/recorder_test.go `

## Privacy

## References


pkg/controller/periodic/periodic_test.go
```diff
@@ -62,12 +65,39 @@ func Test_Controller_CustomPeriodGathererNoPeriod(t *testing.T) {
...
+	for i := range mockRecorder.Records {
+		// find metadata record
+		if mockRecorder.Records[i].Name == recorder.MetadataRecordName {
+			metadataFound = true
+			b, err := mockRecorder.Records[i].Item.Marshal(context.Background())
+			assert.NoError(t, err)
+			metaData := make(map[string]interface{})
+			err = json.Unmarshal(b, &metaData)
+			assert.NoError(t, err)
+			assert.Equal(t, 1, len(metaData["status_reports"].(map[string]interface{})),
```
Sergey1011010, line 87: btw, there's `assert.Len` which fits little better

pkg/gather/gather.go
```diff
@@ -44,8 +44,9 @@ type GathererFunctionReport struct {
 
 // ArchiveMetadata contains the information about the archive and all its' gatherers
 type ArchiveMetadata struct {
-	// info about gathering functions
-	StatusReports []GathererFunctionReport `json:"status_reports"`
+	// info about gathering functions. We need to use map, because functions will repeat
```
Sergey1011010, line 47: We actually don't have to, it could stay an array, but internally we can use a map so that we use only the latest status. Btw, the motivation behind storing all the results was that they can be different(for whatever reason) and we would store all of them.

pkg/gather/gather.go
```diff
@@ -108,14 +109,17 @@ func CollectAndRecordGatherer(
 				errs = append(errs, fmt.Errorf(errStr))
 			}
 		}
-
+		var recordedRecs []record.Record
```
Sergey1011010, line 112: could actually be just a counter since we don't use the records themselves :)

pkg/gather/gather_test.go
```diff
@@ -392,18 +394,18 @@ func assertMetadataOneGatherer(
 	err := json.Unmarshal(metadataBytes, &archiveMetadata)
 	assert.NoError(t, err)
 
-	for i := range archiveMetadata.StatusReports {
-		archiveMetadata.StatusReports[i].Duration = 0
+	for _, v := range archiveMetadata.StatusReports {
+		v.Duration = 0
```
Sergey1011010, line 398: It's actually gonna fail if the test gatherer takes >= 1 ms because this expression assigns the value to a copy of the status report, you need to do something like `archiveMetadata.StatusReports[key].Duration = 0`. Go sometimes isn't obvious..

pkg/gather/gather_test.go
```diff
@@ -392,18 +394,18 @@ func assertMetadataOneGatherer(
...
 	}
 
 	assert.Equal(t, isGlobalObfuscationEnabled, archiveMetadata.IsGlobalObfuscationEnabled)
-	assert.ElementsMatch(t, statusReports, archiveMetadata.StatusReports)
+	assert.True(t, reflect.DeepEqual(statusReports, archiveMetadata.StatusReports))
```
Sergey1011010, line 402: Why? As far as I see ElementsMatch also uses DeepEqual for everything except bytes, but also ignores the order of the elements.

pkg/gather/mock_gatherers.go
```diff
@@ -178,3 +177,30 @@ func (g *MockCustomPeriodGathererNoPeriod) GatherShouldBeProcessed(context.Conte
 		},
 	}, nil
 }
+
+type MockFailingdGatherer struct {
```
Sergey1011010, line 181: a little typo `Failingd` :)

Sergey1011010:
Looks great, just a few things above

pkg/gather/gather.go
tremes: Well I think it's very confusing to store results of failed gatherer which result is not recorded (in the archive), but maybe it's a topic for discussion.
OK for using the map only internally.

pkg/gather/gather_test.go
tremes: There was some failure AFAIK....if we keep it as array then we can keep the orginal definition I guess.

tremes:
Updated. Please check @Sergey1011010

pkg/controller/gather_job.go
```diff
@@ -84,14 +84,24 @@ func (d *GatherJob) Gather(ctx context.Context, kubeConfig, protoKubeConfig *res
...
-	return gather.RecordArchiveMetadata(allFunctionReports, rec, anonymizer)
+	return gather.RecordArchiveMetadata(mapToArray(allFunctionReports), rec, anonymizer)
+}
+
+func mapToArray(m map[string]gather.GathererFunctionReport) []gather.GathererFunctionReport {
+	a := make([]gather.GathererFunctionReport, len(m))
+	for _, v := range m {
+		a = append(a, v)
+	}
+	return a
```
tremes, line 106: It would be nice to move this to `utils` package and reuse...but it causes cyclic dep again AFAIK.

Sergey1011010:
@tremes, sorry, didn't notice, looks great, thanks :)

Sergey1011010 (approved):
/lgtm

Just tested, works as expected: in case some gathering function produces too big result, we get an error to the metadata.

PR #433, opened 2021-05-20
Conditional gathering

Sergey1011010:
This PR implements a new gatherer needed by [CCXDEV-4431](https://issues.redhat.com/browse/CCXDEV-4431). Right now it collects logs and image streams of namespace "openshift-cluster-samples-operator" if alert "SamplesImagestreamImportFailing" is firing, but it designed in a way that it can be further extended to run X gatherer with Y params on Z condition. 

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/conditional/namespaces/openshift-cluster-samples-operator/imagestreams/example.json`
- `docs/insights-archive-sample/conditional/namespaces/openshift-cluster-samples-operator/pods/cluster-samples-operator-8ffb9b45f-49mjr/containers/cluster-samples-operator-watch/logs/last-100-lines.log`
- `docs/insights-archive-sample/conditional/namespaces/openshift-cluster-samples-operator/pods/cluster-samples-operator-8ffb9b45f-49mjr/containers/cluster-samples-operator/logs/last-100-lines.log`

## Documentation

- `path/to/documentation.md`

## Unit Tests

## Privacy

Yes. The collected image streams are anonymized here https://github.com/openshift/insights-operator/pull/433/files#diff-3244b427f7e3f71fcf8c9788e7648b1ae57233dd685b1cb61807698c1bcddf75R70

## Breaking Changes

It's not a breaking change, but some code can potentially misfunction because the new format for function names in metadata was added, for example `conditional/image_stream_definitions_of_namespace/namespace=openshift-cluster-samples-operator` where the last part is the gatherer parameters

## References

https://issues.redhat.com/browse/CCXDEV-4431


pkg/gatherers/conditional/gather_logs_of_namespace.go
```diff
@@ -0,0 +1,88 @@
...
+		common.LogContainersFilter{
+			Namespace:     namespace,
+			LabelSelector: labelSelector,
+		},
+		common.LogMessagesFilter{
+			TailLines: tailLines,
+		},
+		func(namespace string, podName string, containerName string) string {
+			return fmt.Sprintf(
+				"%v/namespaces/%v/pods/%v/containers/%v/logs/%v",
```
tremes, line 78: I think it would be better to preserve same structure as in `clusterconfig` gatherer (something like `%v/pod/%s/logs/%s/%s)`. WDYT?

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -0,0 +1,380 @@
...
+			gatherImageStreamsOfNamespace: map[string]interface{}{
+				"namespace": "openshift-cluster-samples-operator",
+			},
+		},
+	},
+}
+
+const (
+	canConditionalGathererFail = false
+	conditionalGathererPeriod  = time.Hour * 6
```
tremes, line 87: Why 6 hours? I think it would make sense to use the default/configured interval (i.e the same as in `clusterconfig` gatherer)

pkg/gatherers/conditional/gather_logs_of_namespace.go
Sergey1011010: Well, in my opinion the new structure is more clear cuz we have the types of the resources embedded in the path, something similar to REST, also the old structure has a bug that only the logs of one container are saved (because the first version didn't take into account multiple containers). It would be a bad idea to change the structure for existing gatherers, but here it's a totally new one, so we can make it more clear.

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: Well, later we may load the gathering rules from an external API, so there may be many things to gather. But I think it's a question to the PMs, we can as well collect it as often as `clusterconfig` gatherer, it all depends on how fast we want to get this conditionally gathered data.

tremes: My thinking is following. Alerts are gathered in `clusterconfig` anyway so from data analysis point of view you will see that an alert XY is firing, but you will not have related conditional data, because the `conditional` gatherer didn't run this time. This seems to me confusing.

pkg/gatherers/conditional/gather_logs_of_namespace.go
tremes: OK fair point. I think we should also think about the data processing aspect. I am wondering if it can happen (in the future) that same (or very similar) data is gathered in `clusteconfig` as well as in this new `conditional` gatherer. This could cause some issues.....

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -0,0 +1,380 @@
...
+	"github.com/openshift/insights-operator/pkg/record"
+	"github.com/openshift/insights-operator/pkg/utils"
+)
+
+// gatheringFunctionBuilders lists all the gatherers which can be run on some condition. Gatherers can have parameters,
+// like namespace or number of log lines to fetch, see the docs of the functions.
+var gatheringFunctionBuilders = map[gatheringFunctionName]gathererFunctionBuilderPtr{
+	gatherLogsOfNamespace:         (*Gatherer).BuildGatherLogsOfNamespace,
+	gatherImageStreamsOfNamespace: (*Gatherer).BuildGatherImageStreamsOfNamespace,
+}
```
tremes, line 32: We will need to build this map dynamically based on incoming `gatheringRules` definition in the future right? am I understanding it correctly?

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -0,0 +1,380 @@
...
+
+	data, err := metricsClient.Get().AbsPath("federate").
+		Param("match[]", `ALERTS{alertstate="firing"}`).
+		DoRaw(ctx)
+	if err != nil {
+		return err
+	}
+
+	var parser expfmt.TextParser
+	metricFamilies, err := parser.TextToMetricFamilies(bytes.NewReader(data))
```
tremes, line 220: This `err` is ignored.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -0,0 +1,380 @@
...
+
+// updateAlertsCache updates the cache with firing alerts
+func (g *Gatherer) updateAlertsCache(ctx context.Context) error {
+	if g.metricsGatherKubeConfig == nil {
+		return nil
+	}
+
+	metricsClient, err := rest.RESTClientFor(g.metricsGatherKubeConfig)
+	if err != nil {
+		return err
```
tremes, line 203: It's failing here with `open /var/run/configmaps/service-ca-bundle/service-ca.crt: no such file or directory` for some reason when running against my quicklab cluster. Looking....

tremes: Ahh now I see it....we basically ignore this error and just log a warning. See https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/clusterconfig/recent_metrics.go#L42

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: Yeah, but for the conditional gatherer we can look at the metadata if it actually collected something for this archive. No problem removing it, tho, it doesn't make much difference, just the conditional gatherer will run more often.

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: Not this one, but the next one, `gatheringRules`. We'll get a json, unmarshall it to []gatheringRule, check that the data makes sense and use it in the `GetGatheringFunctions`.

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: Yeah, the conditional gatherer requires the metrics to be available https://github.com/openshift/insights-operator#debugging-prometheus-metrics-without-valid-ca

tremes: Can we update it please? I think this is not very convenient (first for running IO locally when you don't have the metrics config). The new conditional gatherer is marked as `canFail=false`, which means that if metrics are not available then IO will go disabled after some time (because this new gatherer will repeatedly fail). So I would suggest one of the following: 

- only log the error here and don't return it (the same we do in our existing metrics gatherer)
- mark the new conditional gatherer as `canFail=true`

Sergey1011010: Hm, yeah, that makes sense, thanks

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -0,0 +1,368 @@
...
+
+	for _, metric := range metricFamily.GetMetric() {
+		if metric == nil {
+			klog.Info("metric is nil")
+			continue
+		}
+
+		for _, label := range metric.GetLabel() {
+			if label == nil {
+				klog.Info("label is nil")
```
tremes, line 229: I guess seeing this message (and similar above) in the IO log will not be very helpful :) on the other hand I am not having any better suggestion at hand (maybe saying something that the metric family with given name references some empty metrics or something like that). It's not blocker for merging.

pkg/gatherers/conditional/types.go
```diff
@@ -0,0 +1,38 @@
...
+type conditionWithParams struct {
+	Type   conditionType          `json:"type"`
+	Params map[string]interface{} `json:"params"`
+}
+
+// GatheringFunctionParams is a type to store gathering function parameters
+type GatheringFunctionParams map[string]interface{}
+
+// gatheringFunctionName defines functions of conditional gatherer
+type gatheringFunctionName string
```
tremes, line 23: It looks like this type could be used more, but it would probably require some more changes. Again not a blocker.

pkg/gatherers/conditional/gather_logs_of_namespace.go
```diff
@@ -0,0 +1,88 @@
...
+//   - namespace string - namespace from which to collect image streams
+//   - label_selector string - a label selector to filter some pods (default to all pods)
+//   - tail_lines int64 - a number of log lines to keep for each container
+//
+// The Kubernetes API:
+//          https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/pod_expansion.go#L48
+// Response see:
+//          https://docs.openshift.com/container-platform/4.6/rest_api/workloads_apis/pod-core-v1.html#apiv1namespacesnamespacepodsnamelog
+//
+// * Location in archive: conditional/namespaces/{namespace}/pods/{pod_name}/containers/{container_name}/logs/last-{n}-lines.log
```
tremes, line 25: OK so we shouldn't forget about our generated docs. Can you please run `make docs`? But this will not be generated anyway. Does it make sense to start the name of the method with `Build` (the gendoc scans only methods starting with `Gather` AFAIK)? 
We can also add here:
```
* Since versions:
//   * 4.8+
``` 
Same for the above `gather_image_stream_definitions_of_namespace.go` gatherer.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -0,0 +1,368 @@
...
+				gatheringFunctions[funcName] = function
+			}
+		}
+	}
+
+	return gatheringFunctions, nil
+}
+
+// GatherConditionalGathererRules stores the gathering rules in insights-operator/conditional-gatherer-rules.json
+func (g *Gatherer) GatherConditionalGathererRules(context.Context) ([]record.Record, []error) {
```
tremes, line 149: This will go to our `docs/gathered-data.md` as I just noticed :) Not sure we want it there....but probably not a big deal. WDYT @jholecek-rh pls? :)

pkg/gatherers/conditional/types.go
Sergey1011010: What do you mean used more?

pkg/gatherers/conditional/gather_logs_of_namespace.go
Sergey1011010: I called it `BuildX` because these are the types of gatherers that can have parameters and they are a little different than normal ones. Lemme check the docs generator.

pkg/gatherers/conditional/types.go
tremes: E.g https://github.com/openshift/insights-operator/pull/433/files#diff-57e9d48447e382f53bafe213301e653fd5989c354c9d751e7f6386be25459e89R251 could return `map[gatheringFunctionName]gatherers.GatheringClosure, []error` and there are similar places...but it's just a detail....

pkg/gatherers/conditional/conditional_gatherer.go
jholecek-rh: We want it there, imho.

I'm just worried about the terminology. Insights rule developers are part of the primary audience for `docs/gathered-data.md`. They will have to pause to realize that it is not talking about Insights rules (like I did right now). I just cannot think of a better name...

tremes: Good points @jholecek-rh. Thanks!

tremes:
I have already reviewed & verified this one several times. Great job!
/lgtm

tremes:
Adding lgtm again.
/lgtm

tremes:
Waiting for https://github.com/openshift/origin/pull/26264
/hold

PR #439, opened 2021-06-01
Bug 1967833: Adds tasks pool to tasks_processing

0sewa0:
Adds a task pool in the task_processing.go so that the resource usage doesn't grow linearly with the number of added gatherers.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

No

### DIFF
--- a/pkg/gather/tasks_processing.go
+++ b/pkg/gather/tasks_processing.go
@@ -36,9 +36,14 @@ func HandleTasksConcurrently(ctx context.Context, tasks []Task) chan GatheringFu
 	// run all the tasks in the background and close the channel when they are finished
 	go func() {
 		var wg sync.WaitGroup
-		workerNum := 4 * runtime.NumCPU()
-		klog.V(4).Infof("number of workers: %d", workerNum)
 		tasksChan := make(chan Task)
+
+		// set number of workers according to the CPU, 1 worker per task max
+		workerNum := 4 * runtime.NumCPU()
+		if len(tasks) < workerNum {
+			workerNum = len(tasks)
+		}
+		klog.V(4).Infof("number of workers: %d", workerNum)
 
 		// create workers
 		for i := 0; i < workerNum; i++ {
--- a/pkg/gather/tasks_processing.go
+++ b/pkg/gather/tasks_processing.go
@@ -2,11 +2,13 @@ package gather
 
 import (
 	"context"
+	"runtime"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/gather/tasks_processing.go
```diff
@@ -29,31 +29,42 @@ type GatheringFunctionResult struct {
...
 func HandleTasksConcurrently(ctx context.Context, tasks []Task) chan GatheringFunctionResult {
-	// TODO: consider using tasks pool with limited number of tasks run at the same time
-	// so that memory usage doesn't grow linearly with the number of added gatherers
-
 	resultsChan := make(chan GatheringFunctionResult)
 
 	// run all the tasks in the background and close the channel when they are finished
 	go func() {
 		var wg sync.WaitGroup
+		workerNum := 5 // TODO: Get this from config
```
Sergey1011010, line 37: I think we want to find some better number by benchmarking for example. Cuz 5 threads should be pretty slow. Probably it should be runtime.NumCPU() multiplied by some number.

tremes: I don't think it's necessary to make it configurable. WDYT?

Sergey1011010: Probably not, but it should be at least moved to a constant.

0sewa0: I choose `5` because it was just the first random number that came to mind :sweat_smile:

I think making it configurable would be cool if anyone wanted to configure it  (ie. the user) but I doubt anyone would actually do that, so it might be a waste of effort. 
Making it "smart" by looking at the number of CPU's and amount of RAM would be also cool but I have no idea what kind of thresholds would I give. (ie. what is the cpu/ram number after which its ok to add a few extra workers) Also making it smart scares me a bit. :smile:

Sergey1011010: We don't need to look at the amount of RAM, it'd be too complicated, we just need to choose the number (which can be multiplied by `runtime.NumCPU()`) big enough so that the operator processes everything in reasonable time, but not too big so that it doesn't consume way too much memory on average.

rluders (approved):
yep, it works.
/lgtm

PR #441, opened 2021-06-01
Bug 1961701: Use configured interval as the event time limit & check series if

tremes:
lastTimeStamp is zero

This PR updates the logic of events gathering to use the gathering interval to get the oldest event time and also looks at event series time in case of `lastTimeStamp` being zero.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No doc update

## Unit Tests

No new test

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/gatherers/clusterconfig/operators_pods_and_events.go
+++ b/pkg/gatherers/clusterconfig/operators_pods_and_events.go
@@ -71,7 +71,7 @@ func (g *Gatherer) GatherClusterOperatorPodsAndEvents(ctx context.Context) ([]re
 		return nil, []error{err}
 	}
 
-	records, err := gatherClusterOperatorPodsAndEvents(ctx, gatherConfigClient, gatherKubeClient.CoreV1())
+	records, err := gatherClusterOperatorPodsAndEvents(ctx, gatherConfigClient, gatherKubeClient.CoreV1(), g.interval)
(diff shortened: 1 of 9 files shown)


pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -193,14 +198,24 @@ func gatherNamespaceEvents(ctx context.Context, coreClient corev1client.CoreV1In
 		return nil, err
 	}
 	// filter the event list to only recent events
-	oldestEventTime := time.Now().Add(-maxEventTimeInterval * 2)
+	oldestEventTime := time.Now().Add(-interval)
```
rluders, line 201: So here we go back to collecting the data within the normal range. Is this intentional?

tremes: Yes I think it makes sense. If you change the default conf and the gathering runs e.g every 4 hours then you will be interested in the events from the last 4 hours (and not only 2).

rluders (approved):
Just checking the code, it does make sense.
/lgtm

0sewa0:
/lgtm
(just a rebase)

PR #442, opened 2021-06-01
Bug 1966717: include full timestamps in the logs

Sergey1011010:
## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `*.log` files now have full timestamps

## Unit Tests

No changes

## Breaking Changes

Yes, the new format of timestamps in the logs

## References

https://issues.redhat.com/browse/CCXDEV-4897

### DIFF
--- a/pkg/gatherers/clusterconfig/gather_logs.go
+++ b/pkg/gatherers/clusterconfig/gather_logs.go
@@ -37,7 +37,7 @@ type logMessagesFilter struct {
 //     - containerNameRegexFilter to filter containers in the pod (keep empty to not filter)
 //   - logMessagesFilter allows you to specify
 //     - messagesToSearch to filter the logs by substrings (case-insensitive)
-//       or regex (add `(?i)` in the beginning to make search case-insensitive)
+//       or regex (add `(?i)` in the beginning to make search case-insensitive). Leave nil to not filter.
 //     - regexSearch which makes messagesToSearch regex patterns, so you can accomplish more complicated search
 //     - sinceSeconds which sets the moment to fetch the logs from (current time - sinceSeconds)
 //     - limitBytes which sets the maximum amount of logs that can be fetched
@@ -85,6 +85,7 @@ func gatherLogsFromContainers(
 				Container:    container,
 				SinceSeconds: &messagesFilter.sinceSeconds,
 				LimitBytes:   &messagesFilter.limitBytes,
+				Timestamps:   true,
 			})
 
 			logs, err := filterLogs(ctx, request, messagesFilter.messagesToSearch, messagesFilter.isRegexSearch)
@@ -129,6 +130,10 @@ func filterLogs(
 
 	for scanner.Scan() {
 		line := scanner.Text()
+		if messagesToSearch == nil {
+			result += line + "\n"
+		}
+
 		for _, messageToSearch := range messagesToSearch {
 			if regexSearch {
 				matches, err := regexp.MatchString(messageToSearch, line)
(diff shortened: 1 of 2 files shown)


pkg/gatherers/clusterconfig/gather_logs.go
```diff
@@ -129,6 +130,10 @@ func filterLogs(
 
 	for scanner.Scan() {
 		line := scanner.Text()
+		if messagesToSearch == nil {
+			result += line + "\n"
+		}
```
tremes, line 135: This opens the new functionality for this function. It would be nice to unify it with https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/clusterconfig/operators_pods_and_events.go#L254 
Just saying..... otherwise LGTM, but we need to check with @jholecek-rh about potential impact on the rules.

jholecek-rh:
@Sergey1011010 Is it possible to tell the timestamp format from `insights-operator/gathers.json`? I'm looking for something for rules to easily detect if they can expect full timestamps or the original ones. It would be really helpful to distinguish why a rule is not hitting:

1. It didn't have the required data (the full timestamp).
2. It didn't find any log messages of interest (a regex didn't match any line).

I'm also wondering how to reflect this breaking change in the documentation. We will make sure that all existing rules will support both the old and the new format. However, we need to make sure that even new rules will do so.

pkg/gatherers/clusterconfig/gather_logs.go
Sergey1011010: Yeah, I started refactoring to have only one place to gather logs, but then decided it doesn't worth the efforts right now.

Sergey1011010:
@jholecek-rh after merging this PR all the logs in the archives will have the new format. So, only the old archives are gonna stick to the old format. It's possible to use a simple regex to detect the format.

tremes:
Just few changes
/lgtm

jholecek-rh:
@Sergey1011010 Let's be clear. "Only old archives" means all archives we've received and all archives we will have received until clusters upgrade to 4.8.something.That's a lot of archives/clusters. :slightly_smiling_face:

Timestamps in log files are an archive-wide "configuration". To distinguish the two cases that I described, we need an Insights component that would serve as an archive-wide indicator of the log file structure.

Option 1:
* We know the structure when writing the archive.
* We pass the information in the archive.
* The component picks it up. 

Clean, robust and easy.

Option 2:
* We know the structure when writing the archive.
* We don't pass it on in the archive.
* The component will
  * Depend on all logs in the archive because logs are collected conditionally and we don't know which logs are there
  * Need to handle the case that there are no log files at all
  * Base the decision on an (undocumented?) assumption that the first line of any log file determines the structure of all log files

Workable, not an easy way to consume the data, not exactly clean, possibly less robust.

This is my last try. I will not push further. :slightly_smiling_face:

jholecek-rh:
@Sergey1011010 How about the documentation? Any ideas how to make consumers of the data aware that we are receiving archives with two different log file structures? Will we expect them to "just know"(tm) and wait for the bright distant future where all (most) clusters will have updated? :innocent:

Sergey1011010:
@jholecek-rh I think it should be pretty easy to make a regex inside the function processing log file that would check which timestamp format it is. We can add some flag somewhere in metadata, but not sure if we want that. @tremes WDYT?

Sergey1011010:
Random thought: we could also include the list of files produces by the function (if any), but I'm not sure if I like this idea, it could simplify processing.

tremes:
Adding some flag to the metadata seems a bit redundant IMO. Couldn't the rule processing just distinguish the cluster versions and act accordingly? Because this update is/will be for >4.8.0.

tremes:
Wrt documentation - we should have updated the sample archive at least right @Sergey1011010 ? I just realized.....

jholecek-rh:
OK, deciding based on the version seems OK. Is it going to be >4.8.0 (i.e. >=4.8.1) or >=4.8.0?

Sergey1011010:
@tremes yup, forgot about it, there it is https://github.com/openshift/insights-operator/pull/447

PR #444, opened 2021-06-04
MemoryRecord name can be obfuscated & fix case of duplicate records

tremes:
`MemoryRecord` name can be partially obfuscated as well (when includes cluster base domain i.e node names, hostsubnet names could contain `<CLUSTER_BASE_DOMAIN>` string). This means that such record couldn't be found in https://github.com/openshift/insights-operator/blob/master/pkg/recorder/recorder.go#L171 and the size was growing....and now we have enforced the size limit -> basically empty archives after some time when obfuscation is enabled, because the size limit was exceeded. 

Also do not check the length of the data fingerprint (since it will be always 0 because we don't use fingerprints) -> fix the size counting in case of same duplicate records (yes it can happen as I found out).

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No update

## Unit Tests

No new test

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/recorder/recorder.go
+++ b/pkg/recorder/recorder.go
@@ -169,14 +169,6 @@ func (r *Recorder) copy() record.MemoryRecords {
 func (r *Recorder) clear(records record.MemoryRecords) {
 	r.lock.Lock()
 	defer r.lock.Unlock()
-	size := int64(0)
-	for _, record := range records {
-		existing, ok := r.records[record.Name]
-		if !ok || existing.Data == nil || existing.At != record.At || existing.Fingerprint != record.Fingerprint {
-			continue
-		}
-		size += int64(len(existing.Data))
-		existing.Data = nil
-	}
-	r.size -= size
+	r.records = make(map[string]*record.MemoryRecord, len(records))
+	r.size = 0
 }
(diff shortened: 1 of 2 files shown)


rluders (approved):
/lgtm

pkg/recorder/recorder.go
```diff
@@ -85,7 +85,7 @@ func (r *Recorder) Record(rec record.Record) error {
 		return fmt.Errorf("record %s(size=%d) exceeds the archive size limit %d and will not be included in the archive",
 			recordName, recordSize, r.maxArchiveSize)
 	}
-	r.records[recordName] = memoryRecord
+	r.records[memoryRecord.Name] = memoryRecord
```
rluders, line 88: @tremes I think that this thing could be a problem, 'cause the way that we flush things. You don't want to use the `memoryRecord.Name` you should definitely use the `recordName`. There are other implications to change this line.

/remove-lgtm

rluders: Just one example, that you can see down below, the function `has` checks it by the Filename, not the RecordName. There is a reason to use the `Filename` instead of the `Name`, but TBH I don't record right now, I think that it is related to the data struct between the `MemoryRecord` and the `Record`.

rluders (approved):
/lgtm

PR #445, opened 2021-06-04
Bug 1967843: Fixes one small bug

rluders:
This PR fixes a nil reference to a variable at the `gather_logs.go` that could cause some issues.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No.

## Documentation

No.

## Unit Tests

No.

## Privacy

## Breaking Changes

No.

## References

https://bugzilla.redhat.com/show_bug.cgi?id=1967843

### DIFF
--- a/pkg/gatherers/clusterconfig/gather_logs.go
+++ b/pkg/gatherers/clusterconfig/gather_logs.go
@@ -132,6 +132,7 @@ func filterLogs(
 		line := scanner.Text()
 		if messagesToSearch == nil {
 			result += line + "\n"
+			continue
 		}
 
 		for _, messageToSearch := range messagesToSearch {


Sergey1011010:
Well, it actually can't, `nil` slices are empty and the for loop won't run, but no problem adding `continue`, it can make it a little more clear that we don't want to do anything else in case `messagesToSearch` is `nil`.

rluders:
> Well, it actually can't, `nil` slices are empty and the for loop won't run, but no problem adding `continue`, it can make it a little more clear that we don't want to do anything else in case `messagesToSearch` is `nil`.

Yep, I agree... it is more to guarantee that the flow is doing what we expect and keep it all clear and under control.

PR #446, opened 2021-06-04
add egress ips support to anonymizer

Sergey1011010:
## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

No changes

## Documentation

No changes

## Unit Tests

No changes

## Privacy

No new data was collected

## Breaking Changes

Not really

## References

https://issues.redhat.com/browse/CCXDEV-4864

### DIFF
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -84,7 +84,7 @@ type ConfigProvider interface {
 
 // NewAnonymizer creates a new instance of anonymizer with a provided config observer and sensitive data
 func NewAnonymizer(clusterBaseDomain string, networks []string, secretsClient corev1client.SecretInterface) (*Anonymizer, error) {
-	networks = append(networks, "127.0.0.1/8")
+	networks = append(networks, "127.0.0.0/8")
 
 	cidrs, err := k8snet.ParseCIDRs(networks)
 	if err != nil {
@@ -140,8 +140,6 @@ func NewAnonymizerFromConfigClient(
 		return nil, err
 	}
 
-	secretsClient := kubeClient.CoreV1().Secrets(secretNamespace)
-
 	if installConfig, exists := clusterConfigV1.Data["install-config"]; exists {
 		networkRegex := regexp.MustCompile(Ipv4NetworkRegex)
 		networks = append(networks, networkRegex.FindAllString(installConfig, -1)...)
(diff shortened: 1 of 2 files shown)


pkg/anonymization/anonymizer.go
```diff
@@ -128,6 +132,19 @@ func NewAnonymizerFromConfigClient(
...
 	}
 
+	// egress subnets
+
+	hostSubnets, err := networkClient.HostSubnets().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, hostSubnet := range hostSubnets.Items {
```
rluders, line 142: I'm pretty sure that the lint will complain about these two `for`, 'cause it will say that you should use `index` or `pointer`.

rluders (changes requested):

pkg/anonymization/anonymizer.go
Sergey1011010: Um, why?

rluders: Check these links:
https://go-critic.github.io/overview#rangeValCopy-ref
https://stackoverflow.com/questions/20185511/range-references-instead-values

tremes:
@Sergey1011010 would it be possible to test it (in our unit tests) please?

Sergey1011010:
@tremes added the test

pkg/anonymization/anonymizer.go
```diff
@@ -83,7 +84,7 @@ type ConfigProvider interface {
 
 // NewAnonymizer creates a new instance of anonymizer with a provided config observer and sensitive data
 func NewAnonymizer(clusterBaseDomain string, networks []string, secretsClient corev1client.SecretInterface) (*Anonymizer, error) {
-	networks = append(networks, "127.0.0.1/8")
+	networks = append(networks, "127.0.0.0/8")
```
tremes, line 87: Why this change please? It's not clear to me at the first glance. We need to update also in the package doc - https://github.com/openshift/insights-operator/pull/446/files#diff-9180eb430c735bed4f79b7892293543fe58bd8bc57493f4f823ca84faa073f8eR12

Sergey1011010: It was a mistake in the original code, not really a problem cuz anonymizations would still work correctly, but /8 means everything after the first byte(127) should be zeros so that it's a valid mask.

tremes: Ahh I see. Thx

tremes:
Reviewed & tested. Please remove BZ Id and I also think it would be nice to backport it to 4.8. Thanks!!
/lgtm

PR #449, opened 2021-06-11
Gather all MachineConfig definitions

tremes:
This PR updates the original `SAPMachineConfig` gatherer to gather all the `MachineConfig` definitions.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

TBD

## Documentation

Updated in `docs/gathered-data.md `

## Unit Tests

Updated in:
- `pkg/gatherers/clusterconfig/machine_configs_test.go`

## Privacy

## Breaking Changes

No

## References


pkg/gatherers/clusterconfig/machine_configs.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	records := []record.Record{}
+	var errs []error
+	for i := range mcList.Items {
+		mc := mcList.Items[i]
+		// preservce the original behavior - that is gathering of complete SAP machineconfig definitions
+		if !isSAPMachineConfig(mc) {
```
tremes, line 78: @natiiix Do we need to preserve this logic? I would say no, because the gatherer wasn't called anyway (see the update in `pkg/gatherers/clusterconfig/clusterconfig_gatherer.go`) and more importantly wasn't released in any production version AFAIK.

docs/gathered-data.md
```diff
@@ -298,6 +297,23 @@ Response see:
...
 
 
+## MachineConfigs
+
+collects MachineConfigs definitions.
+
+Gathered MachineConfigs at the time of implementation of the gatherer:
+* `75-worker-sap-data-intelligence`
+* `75-master-sap-data-intelligence`
+* `99-sdi-generated-containerruntime`
```
tremes, line 307: I think I will remove this.

pkg/gatherers/clusterconfig/machine_configs.go
```diff
@@ -0,0 +1,101 @@
...
+			// remove the sensitive content by overwriting the values
+			err := unstructured.SetNestedField(mc.Object, "intentionally overwritten by Insights Operator", "spec", "config", "storage", "files")
```
tremes, line 80: This is the question how to omit the sensitive data. This will replace the `spec.config.storage.files` with the string `intentionally overwritten by Insights Operator`. I can see other options:

- put just empty struct there
- put XXXX to resemble our legacy anonymization

WDYT please?

tremes: OK so it can't be an empty struct, but I can pass `nil` simply. Any thoughts please @Sergey1011010, @jholecek-rh ?

jholecek-rh: @tremes A preferred solution from the rule development point of view would be:
1. The object remains "valid" after the redaction. As it is a struct , it should remain a struct, not become a string. `nil` is OK.
2. The rule has a way to tell if it was redacted (e.g. gatherers configuration) and that the value cannot be trusted (e.g. a rule checking that it is not empty should not hit on `nil` in that case).

Does it make sense/is it applicable in this context?

tremes: I am not sure I fully understand your second point (is `nil` still ok or is it in contradiction with point 1 ?:)). What do you mean by `gatherers configuration` please?

Sergey1011010: I think replacing with `nil` would be fine. We also need to mention it in the docs somewhere.

tremes: OK thanks. It is already redacted here regardless any configuration. The point here is that we know that `spec.config.storage.files` will very likely contain sensitive data (in plain string) as for example certificates and this data would be very difficult to obfuscate (so we can say it's kinda a new approach how to work with sensitive data).

jholecek-rh: Got it. Let's make it `nil` and let's add a note about it in the gatherer's documentation.

tremes: Thanks. Both are already done. Resolving the conversation.

pkg/gatherers/clusterconfig/machine_configs.go
natiiix: I agree that it most likely didn't make its way into production yet and it's completely fine by me if we remove this part.

pkg/gatherers/clusterconfig/clusterconfig_gatherer.go
```diff
@@ -67,6 +67,7 @@ var gatheringFunctions = map[string]gatheringFunction{
 	"crds":                              importantFunc((*Gatherer).GatherCRD),
 	"host_subnets":                      importantFunc((*Gatherer).GatherHostSubnet),
 	"machine_sets":                      importantFunc((*Gatherer).GatherMachineSet),
+	"machine_configs":                   failableFunc((*Gatherer).GatherMachineConfigs),
```
natiiix, line 70: Should it be failable if we're gonna collect all of them?

natiiix (commented):
I'm gonna test it locally, and if it works fine, I'll approve it afterwards.

pkg/gatherers/clusterconfig/clusterconfig_gatherer.go
tremes: Well I think most of our gatherers should be "failable", but then there's of course the question whether we need this flag or not. This flag is only saying that if the specific gatherer failed and is marked as important then repeat the gathering cycle in some shorter time. Given this, I don't consider `MachineConfig` resource so important.

natiiix: Makes sense in that context.

pkg/gatherers/clusterconfig/machine_configs_test.go
```diff
@@ -32,7 +32,7 @@ func Test_SAPMachineConfigs(t *testing.T) {
 		machineConfigGroupVersionResource: "MachineConfigsList",
 	})
 
-	records, errs := gatherSAPMachineConfigs(context.Background(), machineConfigClient)
+	records, errs := gatherMachineConfigs(context.Background(), machineConfigClient)
```
natiiix, line 35: Please also change the name of the test function to match the gatherer function. It confused me a lot when running the tests because I thought I was still on some kind of an old branch or running into caching issues.

tremes: Good catch. Thanks

natiiix (approved):
/lgtm
Tested locally and it seems to do exactly what is described. :+1:

PR #456, opened 2021-06-16
Better pod log gathering with offset for stacktrace messages

rluders:
This PR implements a better way to gather the pod container logs by using an offset limit based on the line where the first stack trace message is found.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

No.

## Documentation

No.

## Unit Tests

No.

## Privacy

## Changelog

No.

## Breaking Changes

No.

## References

https://issues.redhat.com/browse/CCXDEV-4600

### DIFF
--- a/pkg/gatherers/clusterconfig/const.go
+++ b/pkg/gatherers/clusterconfig/const.go
@@ -14,9 +14,11 @@ const (
 var (
 	registryScheme = runtime.NewScheme()
 	// logMaxLines sets maximum number of lines of the log file
-	logMaxLines = int64(100)
+	logMaxTailLines = int64(100)
+	// logMaxLongTailLines sets maximum number of lines of the long log file
+	logMaxLongTailLines = int64(200)
 	// logLinesOffset sets the maximum offset if a stacktrace message was found in the logs
-	logLinesOffset = int64(20)
+	logLinesOffset = int64(10)
 
 	defaultNamespaces           = []string{"default", "kube-system", "kube-public"}
 	datahubGroupVersionResource = schema.GroupVersionResource{
--- a/pkg/gatherers/clusterconfig/const.go
+++ b/pkg/gatherers/clusterconfig/const.go
@@ -16,9 +16,9 @@ var (
 	// logMaxLines sets maximum number of lines of the log file
 	logMaxTailLines = int64(100)
 	// logMaxLongTailLines sets maximum number of lines of the long log file
-	logMaxLongTailLines = int64(200)
+	logMaxLongTailLines = int64(2000)
 	// logLinesOffset sets the maximum offset if a stacktrace message was found in the logs
-	logLinesOffset = int64(10)
+	logLinesOffset = int64(20)
 
 	defaultNamespaces           = []string{"default", "kube-system", "kube-public"}
 	datahubGroupVersionResource = schema.GroupVersionResource{
--- a/pkg/gatherers/clusterconfig/const.go
+++ b/pkg/gatherers/clusterconfig/const.go
@@ -14,9 +14,9 @@ const (
 var (
 	registryScheme = runtime.NewScheme()
 	// logTailLines sets maximum number of lines to fetch from pod logs
-	logTailLines = int64(100)
-	// logTailLinesLong sets the maximum number of lines to fetch from long pod logs
-	logTailLinesLong = int64(400)
+	logTailLines = int64(4000)
+	// logLinesOffset sets the maximum offset if a stacktrace message was found in the logs
+	logLinesOffset = int64(100)
 
 	defaultNamespaces           = []string{"default", "kube-system", "kube-public"}
 	datahubGroupVersionResource = schema.GroupVersionResource{
--- a/pkg/gatherers/clusterconfig/const.go
+++ b/pkg/gatherers/clusterconfig/const.go
@@ -18,7 +18,7 @@ var (
 	// logMaxLongTailLines sets maximum number of lines of the long log file
 	logMaxLongTailLines = int64(2000)
 	// logLinesOffset sets the maximum offset if a stacktrace message was found in the logs
-	logLinesOffset = int64(20)
+	logLinesOffset = 20
 
 	defaultNamespaces           = []string{"default", "kube-system", "kube-public"}
 	datahubGroupVersionResource = schema.GroupVersionResource{
--- a/pkg/gatherers/clusterconfig/const.go
+++ b/pkg/gatherers/clusterconfig/const.go
@@ -13,10 +13,10 @@ const (
 
 var (
 	registryScheme = runtime.NewScheme()
-	// logTailLines sets maximum number of lines to fetch from pod logs
-	logTailLines = int64(4000)
+	// logMaxLines sets maximum number of lines of the log file
+	logMaxLines = int64(100)
 	// logLinesOffset sets the maximum offset if a stacktrace message was found in the logs
-	logLinesOffset = int64(100)
+	logLinesOffset = int64(20)
 
 	defaultNamespaces           = []string{"default", "kube-system", "kube-public"}
 	datahubGroupVersionResource = schema.GroupVersionResource{
--- a/pkg/gatherers/clusterconfig/operators_pods_and_events.go
+++ b/pkg/gatherers/clusterconfig/operators_pods_and_events.go
@@ -316,21 +316,18 @@ func getContainerLogs(ctx context.Context,
 
 // getLogWithStacktracing search for the first stack trace line and offset it by logLinesOffset
 func getLogWithStacktracing(logArray []string) string {
-	var log []string
-	var found bool
-
+	var limit int
 	for idx := range logArray {
 		line := logArray[idx]
-		log = append(log, line)
-		if !found {
-			if len(log) > logLinesOffset {
-				log = log[1:]
+		if found := stackTraceRegex.MatchString(line); found {
+			limit = idx - logLinesOffset
+			if limit < 0 {
+				limit = 0
 			}
-			found = stackTraceRegex.MatchString(line)
+			break
 		}
 	}
-
-	return strings.Join(log, "\n")
+	return strings.Join(logArray[limit:], "\n")
 }
 
 // getContainerLogString fetch the container log from API and return it as String
--- a/pkg/gatherers/clusterconfig/operators_pods_and_events.go
+++ b/pkg/gatherers/clusterconfig/operators_pods_and_events.go
@@ -41,7 +41,7 @@ type CompactedEventList struct {
 }
 
 // Used to detect the possible stack trace on logs
-var stackTraceRegex = regexp.MustCompile(`.*/[^/]*:\d+.*0[xX][0-9a-fA-F]+`)
+var stackTraceRegex = regexp.MustCompile(`\.go:[0-9]+\s\+0x`)
 
 // GatherClusterOperatorPodsAndEvents collects information about all pods
 // and events from namespaces of degraded cluster operators. The collected
--- a/pkg/gatherers/clusterconfig/operators_pods_and_events.go
+++ b/pkg/gatherers/clusterconfig/operators_pods_and_events.go
@@ -282,15 +282,12 @@ func getContainerLogs(ctx context.Context,
 
 		logName := logFilename(c.Name, isPrevious)
 
-		// Fetch full container logs and continue on error
-		logString, err := getContainerLogString(ctx, client, pod, c.Name, isPrevious, buf, nil)
+		// Fetch container logs and continue on error
+		logString, err := getContainerLogString(ctx, client, pod, c.Name, isPrevious, buf, &logMaxTailLines)
 		if err != nil {
 			klog.V(2).Infof("Error: %q", err)
 			continue
 		}
-
-		// We need an array to slice the log
-		logArray := strings.Split(logString, "\n")
 
 		if found := stackTraceRegex.MatchString(logString); found {
 			klog.V(2).Infof(
@@ -299,15 +296,14 @@ func getContainerLogs(ctx context.Context,
 				pod.Name,
 				pod.Namespace,
 				isPrevious)
-			logArray = getLogWithStacktracing(logArray)
+			logString, err = getContainerLogString(ctx, client, pod, c.Name, isPrevious, buf, &logMaxLongTailLines)
+			if err != nil {
+				klog.V(2).Infof("Error: %q", err)
+				continue
+			}
+			// find the stack trace and get its
+			logString = getLogWithStacktracing(strings.Split(logString, "\n"))
 		}
-
-		// Set the log limits by number of lines
-		maxLines := int(logMaxLines)
-		if len(logArray) < maxLines {
-			maxLines = len(logArray)
-		}
-		logString = strings.Join(logArray[:maxLines], "\n")
 
 		records = append(records, record.Record{
 			Name: fmt.Sprintf("config/pod/%s/logs/%s/%s", pod.Namespace, pod.Name, logName),
@@ -318,8 +314,8 @@ func getContainerLogs(ctx context.Context,
 	return records
 }
 
-// getLogWithStacktracing search for the first stack trace line and offset it
-func getLogWithStacktracing(logArray []string) []string {
+// getLogWithStacktracing search for the first stack trace line and offset it by logLinesOffset
+func getLogWithStacktracing(logArray []string) string {
 	var log []string
 	var found bool
 
@@ -339,12 +335,13 @@ func getLogWithStacktracing(logArray []string) []string {
 		// offset limit value
 		log = append(log, line)
 		if len(log) > int(logLinesOffset) {
-			log[0] = "" // to avoid memory leaks
+			// to avoid memory leaks (https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order)
+			log[0] = ""
 			log = log[1:]
 		}
 	}
 
-	return log
+	return strings.Join(log, "\n")
 }
 
 // getContainerLogString fetch the container log from API and return it as String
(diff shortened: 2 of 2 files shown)


pkg/gatherers/clusterconfig/const.go
```diff
@@ -14,9 +14,9 @@ const (
 var (
 	registryScheme = runtime.NewScheme()
 	// logTailLines sets maximum number of lines to fetch from pod logs
-	logTailLines = int64(100)
-	// logTailLinesLong sets the maximum number of lines to fetch from long pod logs
-	logTailLinesLong = int64(400)
+	logTailLines = int64(4000)
```
rluders, line 17: @tremes instead get the full log I'm getting a pretty reasonable amount of lines, which IMHO, would give us better performance. What do you think?

pkg/gatherers/clusterconfig/const.go
```diff
@@ -14,9 +14,9 @@ const (
...
+	// logLinesOffset sets the maximum offset if a stacktrace message was found in the logs
+	logLinesOffset = int64(100)
```
rluders, line 19: This is the amount of offset that we gonna collect above the line where we found the stack trace message.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -313,6 +309,37 @@ func getContainerLogs(ctx context.Context,
...
+		// if it didn't find the line yet, keep add it as offset limiting by the arbitrary
+		// offset limit value
+		log = append(log, line)
+		if int64(len(log)) > logLinesOffset {
+			log[0] = "" // to avoid memory leaks
+			log = log[1:]
+		}
+	}
+
+	return strings.Join(log, "\n"), found
```
rluders, line 340: If we don't find any stack tracing message... I think that would be nice to get a smaller log... What do you think?

pkg/gatherers/clusterconfig/const.go
natiiix: Seems like a good idea

pkg/gatherers/clusterconfig/const.go
natiiix: 4000 lines is quite a lot, so I believe we may want to add a size limit as well to avoid size problems caused by extremely long lines. It shouldn't be too difficult, right?

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -313,6 +309,37 @@ func getContainerLogs(ctx context.Context,
...
+	var found bool
+	for scanner.Scan() {
+		line := scanner.Text()
+		// if it already found the stack we just want to keep add the log till the end
+		if found {
+			log = append(log, line)
+			continue
+		}
+		// tries to find the stack message at the given line and add it to the log
+		if found = stackTraceRegex.MatchString(line); found {
```
tremes, line 327: Is the regex updated? I can't see it.

rluders: Ops! I reverted it by mistake before committing it.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -289,19 +290,14 @@ func getContainerLogs(ctx context.Context,
 			continue
 		}
 
-		// check for stack tracer
-		if found := stackTraceRegex.MatchString(logString); found {
+		logString, found := getLogWithStacktracing(logString)
```
tremes, line 293: So do we always scan the complete log? I don't it's necessary.

rluders: What is your suggestion?

tremes: Let me repeat the suggestion from https://issues.redhat.com/browse/CCXDEV-4600. 
> If the last 100 lines log (the commonly gathered log) is matching the regex (there's something looking like stacktrace) then load complete log to the memory and try to record the complete stacktrace

Does it make sense?

rluders: So, yep, but the problem in this approach is that the log gathering is already slow enough, and by adding this it will be even slower.

I agree that the result could be better, but imagine that for every pod container log we will check line by line for the stack tracing and then adding it to the final log... So, in this case, also imagine that the stack tracing is located at the first 300 log lines, but the log has 1000 lines, and we have some important info available at the end of the log.

Is it possible to happen? I don't know. Do we want to commit the rest of the log if we found a stack tracing and just focus at the stack tracing?

tremes: No we will not check line by line for every container. The idea is to check it line by line only when the `logString` is matching the regex. I don't think it will be that slow. The assumption here is that there won't be many logs matching the stacktrace regex. Anyway it's difficult to discuss about the performance when we don't have any real numbers. 

The case you're describing (stacktrace located at the first 300 log lines) is something we won't cover. The goal is not to have every possible stacktrace in the log. The goal is IMHO to have complete stacktrace when the last 100 log lines had something looking like stacktrace (of course then there's the question what to do if we found the beginning of the stacktrace there. I guess we can't do much about it...maybe only try to gather it again, but I am not sure).

Sergey1011010: @tremes I think we still want to limit the maximum number of fetched logs in case there was a stacktrace cuz there can be quite a lot of logs (if I got your idea correctly)

tremes: I haven't thought about limit for the stacktrace TBH. My first thought was to get the whole stacktrace and nothing else. If we think we still need to limit it then we can probably ask ourselves if it makes sense at all (because we can't predict which part of the stacktrace will be important).

pkg/gatherers/clusterconfig/const.go
rluders: It seems that we want to search using the full log... and also, we already use a size limit to the file.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: OK. I changed things a little bit. I think that now we do have a smaller log, and also a better result to focus at the stack trace message.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -282,28 +282,33 @@ func getContainerLogs(ctx context.Context,
 
 		logName := logFilename(c.Name, isPrevious)
 
-		// Fetch container logs and continue on error
-		logString, err := getContainerLogString(ctx, client, pod, c.Name, isPrevious, buf, &logTailLines)
+		// Fetch full container logs and continue on error
+		logString, err := getContainerLogString(ctx, client, pod, c.Name, isPrevious, buf, nil)
```
tremes, line 286: I was trying to suggest to avoid this logic (loading the full log every time).....

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -282,28 +282,33 @@ func getContainerLogs(ctx context.Context,
...
-			}
+			logArray = getLogWithStacktracing(logArray)
 		}
 
+		// Set the log limits by number of lines
+		maxLines := int(logMaxLines)
+		if len(logArray) < maxLines {
+			maxLines = len(logArray)
+		}
+		logString = strings.Join(logArray[:maxLines], "\n")
```
tremes, line 310: This seems wrong. We don't want beginning of the log.

rluders: Oh, wait. I inverted the logic.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: So...

1) Get the last 100 lines;
2) Look for stack trace messages;
3) If found one, fetch the log again with logTailLines + offset.

rluders: I think that the amount of lines is too small, TBH.

rluders: or...

1. Get the last 100 lines;
2. Look for stack trace messages;
3. If found one, fetch the full log, check for the first match line + offset.

tremes: Passing nil as the `logTailLines` param is basically give me the whole log if I am not mistaken. https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/core#PodLogOptions

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -313,6 +318,35 @@ func getContainerLogs(ctx context.Context,
...
+		if found = stackTraceRegex.MatchString(line); found {
+			log = append(log, line)
+			continue
+		}
+		// if it didn't find the line yet, keep add it as offset limiting by the arbitrary
+		// offset limit value
+		log = append(log, line)
+		if len(log) > int(logLinesOffset) {
+			log[0] = "" // to avoid memory leaks
+			log = log[1:]
```
tremes, line 343: Sorry. I don't understand this logic. You set the first line to "" and then you omit it. Why do we want to omit the first line?

rluders: If you don't clean up the value stored at the index before you remove it, you could have a memory leak.

tremes: First, please discuss why we have to omit the first line, and then maybe talk about a potential memory leak.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
Sergey1011010: Is there any limit for the number of log lines in openshift? Fetching potentially infinite amount of data isn't cool :) But even if there's I'd vote for still adding our own limit.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
Sergey1011010: @tremes I think the idea is that we always append lines to `log` and if we haven't found the stacktrace yet, but already have too many lines in the `log` we set the first line to an empty string and change a reference to an underlying array so that it now ignores this line, so we get something like circular buffer, but implemented in a very strange way :)

Maybe it's better to rewrite it like this:

```
find the index of the first string matching stackStaceRegex
take all the items from "this index we've just found minus logLinesOffset" till the end and copy it to a new array
```

Of course checking the corner cases like if there was nothing found. If I got the logic correctly.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: So, yes... That is why initially I had set a limit. I will changes some things later.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: Let's discuss it tomorrow.

rluders:
Let me explain what I did here:

I want to find if we have a possible stacktrace to the first fetched log messages (limited to 100 lines). If it doesn't found one, just record, otherwise, if it finds a possible stacktrace message, I'll request a second log, now with 200 lines, that I convert into an array (I need something that I could handle as lines), so, I'll check line by line if I can found the stacktrace message (why didn´t I get the index from the first check? 'cause that one will give me the index at the `[]byte` not the line number)... In this case, I'll keep always an offset (using a very simple circular-buffer-like), it should allow us to reduce the number of lines in the log, and keep a reasonable offset before the stacktrace the first line detected. Also, 'cause we don't know if the following information after the stacktrace message is important, I'm keeping it. So, it means that the stacktrace log, should never have more than 200 lines.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -313,6 +314,36 @@ func getContainerLogs(ctx context.Context,
...
+		if found = stackTraceRegex.MatchString(line); found {
+			log = append(log, line)
+			continue
+		}
+		// if it didn't find the line yet, keep add it as offset limiting by the arbitrary
+		// offset limit value
+		log = append(log, line)
+		if len(log) > int(logLinesOffset) {
+			// to avoid memory leaks (https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order)
+			log[0] = ""
```
tremes, line 339: We don't store a pointer or struct in the array, so I don't think there will be a memory leak, but I could be wrong.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -313,6 +314,36 @@ func getContainerLogs(ctx context.Context,
...
+		}
+		// tries to find the stack message at the given line and add it to the log
+		if found = stackTraceRegex.MatchString(line); found {
+			log = append(log, line)
+			continue
+		}
+		// if it didn't find the line yet, keep add it as offset limiting by the arbitrary
+		// offset limit value
+		log = append(log, line)
+		if len(log) > int(logLinesOffset) {
```
tremes, line 337: `logLinesOffset` can be defined as `int` right? And we will need no retyping.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: It is one line of code, @tremes. One line of code. I prefer to be sure if the cost is one line of code that doesn't impact anything.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: Yep. I think that doesn't make sense to use these consts as `int64` - I'll change it.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -313,6 +314,34 @@ func getContainerLogs(ctx context.Context,
...
+		}
+		// if it didn't find the line yet, keep add it as offset limiting by the arbitrary
+		// offset limit value
+		log = append(log, line)
+		if len(log) > logLinesOffset {
+			log = log[1:]
+		}
+	}
+
+	return strings.Join(log, "\n")
```
tremes, line 342: OK so I was looking at this again. It works fine, but maybe something like following would be more straightforward:
```

	for idx := range logArray {
		line := logArray[idx]
		log = append(log, line)
		if !found {
			if len(log) > logLinesOffset {
				log = log[1:]
			}
			found = stackTraceRegex.MatchString(line)
		}
	}
```
WDYT?

rluders: Looks good.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -313,6 +314,25 @@ func getContainerLogs(ctx context.Context,
...
+func getLogWithStacktracing(logArray []string) string {
+	var log []string
+	var found bool
+
+	for idx := range logArray {
+		line := logArray[idx]
+		log = append(log, line)
+		if !found {
+			if len(log) > logLinesOffset {
+				log = log[1:]
```
Sergey1011010, line 327: But doesn't it still store all the strings in the underlying array in memory?

Sergey1011010: TBH I don't like the whole idea of implementing circular buffer like this and I don't think we need a circular buffer either, it'd be more clear to first find the string with the stack trace and then take all the necessary log lines from the original array. It'll be logLinesOffset more iterations, but I think not a big deal.

rluders: @Sergey1011010 check it now.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -313,6 +314,22 @@ func getContainerLogs(ctx context.Context,
...
+		line := logArray[idx]
+		if found := stackTraceRegex.MatchString(line); found {
+			limit = idx - logLinesOffset
+			if limit < 0 {
+				limit = 0
+			}
+			break
+		}
+	}
+	return strings.Join(logArray[limit:], "\n")
```
tremes, line 330: I think it looks good. Maybe we could copy the slice here to avoid keeping the underlying array in memory. I am leaving the final approval to @Sergey1011010 :-P

rluders: Yeah. It also works, but I think that it is OK. The GC should handle it.

Sergey1011010: Yeah, it should be removed since we're returning the result of `join`. Btw, if nothing is found, it will return all the logs, it is expected, right?

Sergey1011010:
Looks great

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: Yep. It will return all the logs. But I think that would be really difficult to fit into this scenario, once we already detect the stack trace in the 100 last lines then we request the 2000 lines.

tremes: Yes this function is called only when the regex pattern is found in the log snippet so there should be a line matching it as well.

tremes:
I reviewed it many times and also tested few times. 
/lgtm

tremes:
We need to fix the integration test failures first.
/hold

PR #460, opened 2021-06-18
Remove managedFields from gathered resources

natiiix:
In OCP 4, JSON representations of resources returned by the API contain a metadata field called `managedFields`, which takes up a considerable amount of space without being of any use to us.

This PR introduces a serialization/marshalling filter, which automatically removes this field from resources stored in the IO archive by using `ResourceMarshaller` instead of `JSONMarshaller` for resources that have `managedFields`. This is performed in a somewhat type-safe way for compile-time transparency.

The result is an archive size reduction, both in the raw and compressed form, by approximately 10%.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

Many sample resource JSONs have been modified, so please see the "Files changed" tab for a complete list.

All resources affected by this change should now be updated in the sample archive. There should be no `managedFields` present in any of the sample resources.

## Documentation

No.

## Unit Tests

- `pkg/record/resource_marshaller_test.go`
  - `Test_ResourceMarshaller_GetExtension`
  - `Test_ResourceMarshaller_MarshalUnstructured`
  - `Test_ResourceMarshaller_MarshalPod`

## Privacy

This PR adds no new data to the archive, only removes the `managedFields` fields.

## Changelog

No.

## Breaking Changes

No.

The `managedFields` fields have been removed, but these should not be used by any of the software that processes the IO archives.

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-4899


pkg/record/record.go
```diff
@@ -30,11 +32,29 @@ type Marshalable interface {
 	GetExtension() string
 }
 
+// managedFieldsSetter is a universal interface of
+// resources that implement the SetManagedFields method.
+type managedFieldsSetter interface {
```
natiiix, line 37: This should probably be in one of our `types` packages instead. wdyt?

pkg/record/record.go
```diff
@@ -30,11 +32,29 @@ type Marshalable interface {
...
+}
+
 type JSONMarshaller struct {
 	Object interface{}
 }
 
 func (m JSONMarshaller) Marshal(_ context.Context) ([]byte, error) {
+	if mfs, ok := m.Object.(managedFieldsSetter); ok {
+		mfs.SetManagedFields([]metav1.ManagedFieldsEntry{})
+	} else if strMap, ok := m.Object.(map[string]interface{}); ok {
```
natiiix, line 48: If we passed the `unstructured.Unstructured` struct instead of the `unstructured.Unstructured.Object` map, the first condition would hit and the map-related code wouldn't have to be there at all.

tremes: I am failing to see how it would work with other types, which are not `Unstructured`. Also the `unstructured.Unstructured.Object` is not the same `Object` we have here...

natiiix: I'm sorry about the confusion. It's tricky to describe these kinds of ideas in text very well when everything is called `Object`.

Sergey1011010:
Haven't tested, but looks great

pkg/record/resource_marshaller_test.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Contains(jsonBytesRM, managedFieldsCheckBytes) {
+		t.Fatal("managedFields field has not been removed from the resource")
+	}
+
+	if len(jsonBytesRM) >= len(jsonBytesDirect) {
+		t.Fatalf("JSON from ResourceMarshaller is not smaller than directly marshalled JSON (%d >= %d)", len(jsonBytesRM), len(jsonBytesDirect))
+	}
```
tremes, line 84: The last two tests could be merged into one using two test cases I guess, but it's really a minor detail and this is OK as well.

tremes:
I didn't check the json files updates. I reviewed the changes in the code and randomly checked some files in the archive (IOW I ran it against cluster-bot cluster). The `config/configmaps/kube-system/cluster-config-v1.json` (I know it was added recently) still contains the `managedFields`, but I think it's not a blocker and we can update it in some next PR. Good job! Thank you!
/lgtm

PR #466, opened 2021-06-29
fix obfuscation translation table secret manifest

Sergey1011010:
Fixed manifests and tested by doing `oc apply -f 03-clusterrole.yaml` and deploying my own image with the latest version of the code.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No

## Documentation

No

## Unit Tests

No

## Privacy

No

## Changelog

No

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-4481

### DIFF
--- a/manifests/03-clusterrole.yaml
+++ b/manifests/03-clusterrole.yaml
@@ -10,9 +10,9 @@ roleRef:
   kind: ClusterRole
   name: system:auth-delegator
 subjects:
-- kind: ServiceAccount
-  namespace: openshift-insights
-  name: operator
+  - kind: ServiceAccount
+    namespace: openshift-insights
+    name: operator
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
@@ -27,9 +27,9 @@ roleRef:
   kind: Role
   name: extension-apiserver-authentication-reader
 subjects:
-- kind: ServiceAccount
-  namespace: openshift-insights
-  name: operator
+  - kind: ServiceAccount
+    namespace: openshift-insights
+    name: operator
 ---
 
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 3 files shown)


manifests/03-clusterrole.yaml
```diff
@@ -316,6 +316,36 @@ roleRef:
...
+  - apiGroups:
+      - ''
+    resources:
+      - secrets
+    verbs:
+      - create
+      - get
+      - watch
+      - list
+      - delete
```
tremes, line 338: So this was wrong in our previous version? I see we had:
```
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
  - list
  - create
  - update
  - delete
```

Sergey1011010: Hm, the previous one also works when manually applied. I gotta research more how manifests are applied :D

tremes: It's weird...your update works locally fine for me (after applying the manifest). Isn't it all about the kubeConfig vs protoKubeConfig? I would like to fully understand TBH, because there is basically the same problem (creating/updating a secret - just in different namespace) in https://github.com/openshift/insights-operator/pull/375 and from my local testing, it seems it doesn't require any additional permissions.......

Sergey1011010: > Isn't it all about the kubeConfig vs protoKubeConfig

Hm, not sure

tremes: It's not...I just verified.

manifests/03-clusterrole.yaml
```diff
@@ -316,6 +316,36 @@ roleRef:
...
-- kind: ServiceAccount
-  name: operator
+  - kind: ServiceAccount
+    name: operator
+    namespace: openshift-insights
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: openshift-insights-secrets-full-access
```
tremes, line 326: OK this will not work. I would suggest to keep the original role names/definitions we have in 4.8 and just and the required annotations (the same as other resources have) - see https://coreos.slack.com/archives/CEGKQ43CP/p1625119637397800.

tremes:
Majority is just formatting update. The rest looks good to me and hopefully it will finally fix the secret permissions. 
/lgtm

PR #469, opened 2021-07-01
Remove ParseJSONQuery function and replace it with unstructured

tremes:
functions

It seems we don't need to implement/rely on `ParseJSONQuery` and it's better to use `unstructured` API instead.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No update

## Unit Tests

Corresponding unit tests updated

## Privacy

## Breaking Changes

No

## References


pkg/utils/unstructerd_wrappers.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func NestedStringWrapper(obj map[string]interface{}, fields ...string) (string, error) {
+	s, ok, err := unstructured.NestedString(obj, fields...)
+	if !ok {
+		return "", fmt.Errorf("can't find %s", fields)
```
natiiix, line 12: I don't know how, but it works. I thought `%s` had to be a string, but apparently a slice is fine too. I don't think I've seen it used like this before instead of `%v`.

natiiix (approved):
Tested locally and everything seems to work fine, so I'm glad the confusing function is finally being removed.
/lgtm

pkg/utils/unstructerd_wrappers.go
tremes: Good point. It's true that it looks little bit weird e.g `can't find [spec displayName]`....I think I will try to update. Thanks

rluders (approved):
Tested locally.
/lgtm

PR #470, opened 2021-07-01
Conditional gathering config validation

Sergey1011010:
## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

No changes

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gather/gather_test.go`
- `pkg/gatherers/conditional/validation_test.go`
- `pkg/utils/errors_test.go`

## Privacy

No new data was collected

## Changelog

No

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-5145


pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -111,9 +109,16 @@ func (g *Gatherer) GetName() string {
 // GetGatheringFunctions returns gathering functions that should be run considering the conditions
 // + the gathering function producing metadata for the conditional gatherer
 func (g *Gatherer) GetGatheringFunctions(ctx context.Context) (map[string]gatherers.GatheringClosure, error) {
+	// later the config will be downloaded from an external source
+	errs := validateGatheringRules(gatheringRules)
+	if len(errs) > 0 {
+		klog.Error("got invalid config for conditional gatherer: %v", utils.SumErrors(errs))
```
tremes, line 115: It should be `klog.Errorf`. Anyway, I got the following message when I tried to introduce some validation failure (and then noticed it in the tests...lol):
```
conditional_gatherer.go:116] got invalid config for conditional gatherer: Key: 'GatheringRule.Conditions[].Type' Error:Field validation for 'Conditions[].Type' failed on the 'is_valid' tag
```
I don't find this message very helpful TBH, but this might change in the future. Maybe the `GatheringRule` will have a name....

pkg/gatherers/conditional/validation.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+
+	return allErrors
+}
+
+func validateGatheringFunctionsParamsAreStructs(gatheringFunctions GatheringFunctions) (errs []error) {
+	// a workaround to not panic https://github.com/go-playground/validator/issues/789
+	for key, val := range gatheringFunctions {
+		if reflect.TypeOf(val).Kind() != reflect.Struct {
```
tremes, line 42: I think this type of validation should be easily enforced with validating the JSON condition definition against JSON schema. My feeling is that we started from the wrong end here. I think we will receive JSON data from the new service so I guess these kinds of problems will be caught when doing `json.Unmarshal`.

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: Well, the underlying error has a lot of information, but it's not exposed by the default `Error()` function, I'll try to find a way to make it more comprehensible

pkg/gatherers/conditional/validation.go
Sergey1011010: Well, this PR lacks the unmarshalling thing, but it will be later added and it'll be not too much of code, I just wanted to introduce it in one of the next PRs. Yes, it's doing some extra validation, but it may be useful if we for example decide to keep some rules in the code of the operator and some in the external repo.

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: @tremes I think now (I turned them to JSONs) is the most we can get, we could also add the value itself to the error, but it maybe not the best idea, not sure.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -20,64 +21,16 @@ import (
 	"github.com/openshift/insights-operator/pkg/gatherers"
 	"github.com/openshift/insights-operator/pkg/gatherers/common"
 	"github.com/openshift/insights-operator/pkg/record"
+	"github.com/openshift/insights-operator/pkg/utils"
 )
 
+const gatheringRulesEndpoint = "http://localhost:8080/rules.json"
```
tremes, line 27: It would be good to read this value from config.

pkg/gatherers/conditional/parsing.go
```diff
@@ -0,0 +1,67 @@
...
+	result := GatheringRule{
+		GatheringFunctions: make(GatheringFunctions),
+	}
+
+	for i := 0; i < len(rule.Conditions); i++ {
+		condition := &rule.Conditions[i]
+
+		jsonParams, err := json.Marshal(condition.Params)
+		if err != nil {
+			return GatheringRule{}, err
```
tremes, line 38: Maybe we could use a reference and return nil here.

pkg/gatherers/conditional/validation.go
tremes: OK, but this option (having something in the code and something externally) now seems like a bit of overengineering to me TBH. I would leave this kind of validation up to the JSON schema (we can validate here against it as well).

pkg/gatherers/conditional/validation.go
```diff
@@ -0,0 +1,147 @@
...
+
+	for name, params := range gatheringRule.GatheringFunctions {
+		validateGatheringFunction(name, params, sl)
+	}
+}
+
+// validateCondition validates that a value of type ConditionWithParams is valid.
+// Valid value should have a valid Type and Params of the corresponding type
+func validateCondition(condition *ConditionWithParams, sl validator.StructLevel) {
+	if err := condition.Type.IsValid(); err != nil {
```
tremes, line 70: I would leave this validation up to the JSON schema as well. We can define a set/enum of valid condition types.

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: Good point

pkg/gatherers/conditional/validation.go
Sergey1011010: Hm, not really sure if the schema validation can do everything we need, I will take the task about schema validation for the repo and explore if it works for the operator too.

tremes: Me neither. I am just saying I would do as much as possible in the schema and leave some leftover here in the code. Because imagine one wants to introduce a new condition type. In this situation you would need an update in the IO code (possibly bugzilla, backporting, all the annoying stufff etc.) to pass the validation vs just updating the schema (which can be retrieved from the repository). 
It's just an idea and needs some real proof/verification.

Sergey1011010:
Replaced by https://github.com/openshift/insights-operator/pull/483

PR #472, opened 2021-07-14
Bug 1982170: Set also the summary operation when updating status

tremes:
Set also to operation so it can be distinguished in `status.go`. See the messy code here https://github.com/openshift/insights-operator/blob/master/pkg/controller/status/status.go#L156. We should probably set `degradingFailure` to false by default. There's already issue to refactor this method. 

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No doc update

## Unit Tests

No new test but we should think about some integration test if possible. 

## Privacy

## Breaking Changes

No

## References

https://bugzilla.redhat.com/show_bug.cgi?id=1982170


natiiix (approved):
Didn't test locally because this is just a simple fix.
/lgtm

tisnik (approved):
Seems legit!

PR #473, opened 2021-07-16
Bug 1988349: Insightsreport set corresponding clusteroperator condition correctly

tremes:
This PR registers the `insightsreport` status controller so that the operation can be distinguished and new condition (called `InsightsDownloadDegraded`) introduced 

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No doc update

## Unit Tests

I will work on new integration test.

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -24,12 +24,19 @@ import (
 	"k8s.io/klog/v2"
 )
 
-// How many upload failures in a row we tolerate before starting reporting
-// as UploadDegraded
-const uploadFailuresCountThreshold = 5
-
-// GatherFailuresCountThreshold defines how many gatherings can fail in a row before we report Degraded
-const GatherFailuresCountThreshold = 5
+const (
+	// OperatorDisabled defines the condition type when the operator's primary funcion has been disabled
+	OperatorDisabled configv1.ClusterStatusConditionType = "Disabled"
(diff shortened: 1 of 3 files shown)


pkg/controller/status/status.go
```diff
@@ -181,6 +188,11 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 				disabledReason = summary.Reason
 				disabledMessage = summary.Message
 			}
+		} else if summary.Operation == controllerstatus.DownloadingReport {
+			degradingFailure = false
+			klog.V(4).Info("Failed to download Insights report")
```
tremes, line 193: I will try to think about some other/better message here....on the other hand it's pretty clear from the previous log messages that the download failed....

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -209,7 +219,7 @@ func (r *Gatherer) RetrieveReport() {
 }
 
 // Run goroutine code for gathering the reports from Smart Proxy
-func (r *Gatherer) Run(ctx context.Context) {
+func (r *Controller) Run(ctx context.Context) {
```
rluders, line 222: `(c *Controller)`

rluders (changes requested):

PR #477, opened 2021-07-21
Bug 1984047: Do not use klog.Fatal

tremes:
`klog.Fatal` dumps goroutines stack traces and calls `os.Exit(255)`. This is not desired.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No documentation update

## Unit Tests

No new test

## Privacy

## Breaking Changes

No

## References

https://bugzilla.redhat.com/show_bug.cgi?id=1984047

### DIFF
--- a/pkg/cmd/start/start.go
+++ b/pkg/cmd/start/start.go
@@ -69,15 +69,15 @@ func NewGather() *cobra.Command {
 func runGather(operator *controller.GatherJob, cfg *controllercmd.ControllerCommandConfig) func(cmd *cobra.Command, args []string) {
 	return func(cmd *cobra.Command, args []string) {
 		if configArg := cmd.Flags().Lookup("config").Value.String(); len(configArg) == 0 {
-			klog.Error("error: --config is required")
+			klog.Exit("error: --config is required")
 		}
 		unstructured, _, _, err := cfg.Config()
 		if err != nil {
-			klog.Error(err)
+			klog.Exit(err)
 		}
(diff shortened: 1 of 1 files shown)


pkg/cmd/start/start.go
```diff
@@ -69,36 +69,36 @@ func NewGather() *cobra.Command {
 func runGather(operator *controller.GatherJob, cfg *controllercmd.ControllerCommandConfig) func(cmd *cobra.Command, args []string) {
 	return func(cmd *cobra.Command, args []string) {
 		if configArg := cmd.Flags().Lookup("config").Value.String(); len(configArg) == 0 {
-			klog.Fatalf("error: --config is required")
+			klog.Error("error: --config is required")
```
Sergey1011010, line 72: But don't we need to stop the execution in case of an error?

tremes: Well the execution is stopped, but it's true that perhaps `klog.Exit` would be better in this case.

tremes: Thanks

natiiix (approved):
After the fix mentioned by my colleague, I see no further issues. I have not tested this locally because it would be hard to simulate all the error cases, but both functions terminate the application, with the exception of the last line where `Fatal` was changed to `Error`, so there is no reason to believe that this could break IO in any way.

/lgtm

PR #478, opened 2021-07-27
Bug 1987192: Set the disabled state only when the token is removed from the

tremes:
pull-secret

The disabled state was not consistent previously. We set it in the following situations:

- token is not valid (or ingress returned 403 or 401 for some reason). Disabled was set together with `UploadDegraded` state, but the gathering continued. 
- gathering failed multiple times in a row. Disabled was set, but it didn't have any effect. Gathering continued after some time
- token was removed from the `pull-secret` -> disabled was set -> gathering as well as uploading was stopped

This PR changes the disabled setting to happen only in the last situation. That's the only situation when it makes sense IMO. 

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No update

## Documentation

No update

## Unit Tests

No new unit test, but it requires integ. test update https://gitlab.cee.redhat.com/ccx/insights-operator-tests/-/merge_requests/49

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -151,36 +151,19 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 			continue
 		}
 
-		degradingFailure := true
+		degradingFailure := false
 
 		if summary.Operation == controllerstatus.Uploading {
 			if summary.Count < uploadFailuresCountThreshold {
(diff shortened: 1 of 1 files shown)


rluders (approved):
/lgtm

Sergey1011010 (approved):
/lgtm

PR #483, opened 2021-08-06
Json schema validation

Sergey1011010:
This PR implements validation of gathering rules for conditional gatherer on the operator side

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

No changes

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gather/gather_test.go`
- `pkg/gatherers/conditional/parsing_test.go`
- `pkg/gatherers/conditional/validation_test.go`
- `pkg/utils/errors_test.go`

## Privacy

No new data was collected

## Changelog

No

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-5145

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -29,6 +29,7 @@ type Serialized struct {
 		Interval string `json:"interval"`
 		Disabled bool   `json:"disabled"`
 	}
+	ConditionalGathererRulesEndpoint string `json:"conditional_gatherer_rules_endpoint"`
 }
 
 // Controller defines the standard config for this operator.
@@ -60,8 +61,9 @@ type Controller struct {
 	Password string
 	Token    string
 
-	HTTPConfig HTTPConfig
-	OCMConfig  OCMConfig
+	HTTPConfig                       HTTPConfig
+	OCMConfig                        OCMConfig
+	ConditionalGathererRulesEndpoint string `json:"conditional_gatherer_rules_endpoint"`
 }
 
 // HTTPConfig configures http proxy and exception settings if they come from config
@@ -115,6 +117,7 @@ func ToController(s *Serialized, cfg *Controller) (*Controller, error) { // noli
 	cfg.Impersonate = s.Impersonate
 	cfg.Gather = s.Gather
 	cfg.EnableGlobalObfuscation = s.EnableGlobalObfuscation
+	cfg.ConditionalGathererRulesEndpoint = s.ConditionalGathererRulesEndpoint
 
 	if len(s.Interval) > 0 {
 		d, err := time.ParseDuration(s.Interval)
@@ -198,6 +201,7 @@ func ToDisconnectedController(s *Serialized, cfg *Controller) (*Controller, erro
 	cfg.Impersonate = s.Impersonate
 	cfg.Gather = s.Gather
 	cfg.EnableGlobalObfuscation = s.EnableGlobalObfuscation
+	cfg.ConditionalGathererRulesEndpoint = s.ConditionalGathererRulesEndpoint
 
 	if len(s.Interval) > 0 {
 		d, err := time.ParseDuration(s.Interval)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -29,7 +29,7 @@ type Serialized struct {
 		Interval string `json:"interval"`
 		Disabled bool   `json:"disabled"`
 	}
-	ConditionalGathererRulesEndpoint string `json:"conditional_gatherer_rules_endpoint"`
+	ConditionalGathererRulesEndpoint string `json:"conditionalGathererRulesEndpoint"`
 }
 
 // Controller defines the standard config for this operator.
@@ -63,7 +63,7 @@ type Controller struct {
 
 	HTTPConfig                       HTTPConfig
 	OCMConfig                        OCMConfig
-	ConditionalGathererRulesEndpoint string `json:"conditional_gatherer_rules_endpoint"`
+	ConditionalGathererRulesEndpoint string
 }
 
 // HTTPConfig configures http proxy and exception settings if they come from config
--- a/pkg/gather/gather.go
+++ b/pkg/gather/gather.go
@@ -55,14 +55,13 @@ type ArchiveMetadata struct {
 func CreateAllGatherers(
 	gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config,
 	anonymizer *anonymization.Anonymizer, controller *config.Controller,
-	configurator configobserver.Configurator,
 ) []gatherers.Interface {
 	clusterConfigGatherer := clusterconfig.New(
 		gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig, anonymizer, controller.Interval,
 	)
 	workloadsGatherer := workloads.New(gatherProtoKubeConfig)
 	conditionalGatherer := conditional.New(
-		gatherProtoKubeConfig, metricsGatherKubeConfig, configurator.Config().ConditionalGathererRulesEndpoint,
+		gatherProtoKubeConfig, metricsGatherKubeConfig, controller.ConditionalGathererRulesEndpoint,
 	)
 
 	return []gatherers.Interface{clusterConfigGatherer, workloadsGatherer, conditionalGatherer}
--- a/pkg/gather/gather.go
+++ b/pkg/gather/gather.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"runtime"
-	"sort"
 	"strings"
 	"time"
 
@@ -56,12 +55,15 @@ type ArchiveMetadata struct {
 func CreateAllGatherers(
 	gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config,
 	anonymizer *anonymization.Anonymizer, controller *config.Controller,
+	configurator configobserver.Configurator,
 ) []gatherers.Interface {
 	clusterConfigGatherer := clusterconfig.New(
 		gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig, anonymizer, controller.Interval,
 	)
 	workloadsGatherer := workloads.New(gatherProtoKubeConfig)
-	conditionalGatherer := conditional.New(gatherProtoKubeConfig, metricsGatherKubeConfig)
+	conditionalGatherer := conditional.New(
+		gatherProtoKubeConfig, metricsGatherKubeConfig, configurator.Config().ConditionalGathererRulesEndpoint,
+	)
 
 	return []gatherers.Interface{clusterConfigGatherer, workloadsGatherer, conditionalGatherer}
 }
@@ -128,11 +130,11 @@ func CollectAndRecordGatherer(
 			FuncName:     fmt.Sprintf("%v/%v", gathererName, result.FunctionName),
 			Duration:     result.TimeElapsed.Milliseconds(),
 			RecordsCount: recordedRecs,
-			Errors:       errorsToStrings(result.Errs),
+			Errors:       utils.ErrorsToStrings(result.Errs),
 			Panic:        result.Panic,
 		})
 	}
-	return functionReports, sumErrors(errs)
+	return functionReports, utils.SumErrors(errs)
 }
 
 // RecordArchiveMetadata records info about archive and gatherers' reports
@@ -219,31 +221,4 @@ func getListOfEnabledFunctionForGatherer(gathererName string, allFunctionsList [
 	}
 
 	return false, result
-}
-
-// sumErrors simply sorts the errors and joins them with commas
-func sumErrors(errs []error) error {
-	if len(errs) == 0 {
-		return nil
-	}
-
-	var errStrings []string
-	for _, err := range errs {
-		errStrings = append(errStrings, err.Error())
-	}
-
-	sort.Strings(errStrings)
-	errStrings = utils.UniqueStrings(errStrings)
-
-	return fmt.Errorf("%s", strings.Join(errStrings, ", "))
-}
-
-// errorsToStrings turns error slice to string slice
-func errorsToStrings(errs []error) []string {
-	var result []string
-	for _, err := range errs {
-		result = append(result, err.Error())
-	}
-
-	return result
 }
--- a/pkg/gatherers/conditional/conditional_gatherer.go
+++ b/pkg/gatherers/conditional/conditional_gatherer.go
@@ -246,7 +246,14 @@ func (g *Gatherer) createGatheringClosures(
 		if err != nil {
 			errs = append(errs, err)
 		} else {
-			name := getConditionalGatheringFunctionName(string(functionName), functionParams)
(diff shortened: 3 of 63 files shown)


pkg/gather/gather.go
```diff
@@ -56,12 +55,15 @@ type ArchiveMetadata struct {
...
+	configurator configobserver.Configurator,
 ) []gatherers.Interface {
 	clusterConfigGatherer := clusterconfig.New(
 		gatherKubeConfig, gatherProtoKubeConfig, metricsGatherKubeConfig, anonymizer, controller.Interval,
 	)
 	workloadsGatherer := workloads.New(gatherProtoKubeConfig)
-	conditionalGatherer := conditional.New(gatherProtoKubeConfig, metricsGatherKubeConfig)
+	conditionalGatherer := conditional.New(
+		gatherProtoKubeConfig, metricsGatherKubeConfig, configurator.Config().ConditionalGathererRulesEndpoint,
+	)
```
tremes, line 66: I don't think there's a need to pass the `configobserver.Configurator` instance to the `CreateAllGatherers` function. You already have the `controller` here so you can call `controller.ConditionalGathererRulesEndpoint`.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -88,18 +41,29 @@ type Gatherer struct {
...
-func New(gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config) *Gatherer {
-	imageKubeConfig := rest.CopyConfig(gatherProtoKubeConfig)
-	imageKubeConfig.QPS = common.ImageConfigQPS
-	imageKubeConfig.Burst = common.ImageConfigBurst
+func New(gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config, gatheringRulesEndpoint string) *Gatherer {
+	var imageKubeConfig *rest.Config
+	if gatherProtoKubeConfig != nil {
+		imageKubeConfig = rest.CopyConfig(gatherProtoKubeConfig)
+		imageKubeConfig.QPS = common.ImageConfigQPS
+		imageKubeConfig.Burst = common.ImageConfigBurst
```
tremes, line 55: Just out of curiosity - why do you set these?

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -88,18 +41,29 @@ type Gatherer struct {
...
 
 	return &Gatherer{
 		gatherProtoKubeConfig:   gatherProtoKubeConfig,
 		metricsGatherKubeConfig: metricsGatherKubeConfig,
 		imageKubeConfig:         imageKubeConfig,
+		gatheringRulesEndpoint:  gatheringRulesEndpoint,
+		httpClient: &http.Client{
+			Transport: httpcache.NewMemoryCacheTransport(),
+			Timeout:   time.Minute,
+		},
```
tremes, line 66: Just note. It would be probably good to move it to `insightsclient` once there's some auth required.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -273,10 +254,32 @@ func (g *Gatherer) createGatheringClosures(
...
+
 // getConditionalGatheringFunctionName creates a name of the conditional gathering function adding the parameters
 // after the name. For example:
 //   "conditional/logs_of_namespace/namespace=openshift-cluster-samples-operator,tail_lines=100"
-func getConditionalGatheringFunctionName(funcName string, gatherParams map[string]interface{}) string {
+func getConditionalGatheringFunctionName(funcName string, gatherParamsInterface interface{}) string {
+	gatherParams, err := utils.StructToMap(gatherParamsInterface)
+	if err != nil {
+		// will only happen when non a struct is passed which means code is completely broken and panicking is ok
+		panic(err)
```
tremes, line 280: I am affraid this can terminate our main go routine (that's undesirable....but correct me if I am wrong) so I think it would be safer to return the error.

pkg/config/config.go
```diff
@@ -60,8 +61,9 @@ type Controller struct {
 	Password string
 	Token    string
 
-	HTTPConfig HTTPConfig
-	OCMConfig  OCMConfig
+	HTTPConfig                       HTTPConfig
+	OCMConfig                        OCMConfig
+	ConditionalGathererRulesEndpoint string `json:"conditional_gatherer_rules_endpoint"`
```
tremes, line 66: This should be just string - no need to define the json marshalling info

pkg/config/config.go
```diff
@@ -29,6 +29,7 @@ type Serialized struct {
 		Interval string `json:"interval"`
 		Disabled bool   `json:"disabled"`
 	}
+	ConditionalGathererRulesEndpoint string `json:"conditional_gatherer_rules_endpoint"`
```
tremes, line 32: `json:"conditional_gatherer_rules_endpoint"`doesn't correspond to the [config in the yaml files](https://github.com/openshift/insights-operator/pull/483/files#diff-073e48cb952955fd3f328a1110efe362d38fad66b985b21c5b1a1865425312c1R19) and thus it won't work.

pkg/gatherers/conditional/conditional_gatherer_test.go
```diff
@@ -16,15 +17,15 @@ import (
...
+	return New(nil, nil, "")
 }
 
 func Test_Gatherer_Basic(t *testing.T) {
 	gatherer := newEmptyGatherer()
 	assert.Equal(t, "conditional", gatherer.GetName())
 	gatheringFunctions, err := gatherer.GetGatheringFunctions(context.TODO())
-	assert.NoError(t, err)
-	assert.Greater(t, len(gatheringFunctions), 0)
+	assert.EqualError(t, err, `unable to load gathering rules: Get "": unsupported protocol scheme ""`)
```
tremes, line 27: Maybe it would be slightly more realistic to pass some `gatheringRulesEndpoint` to the `New` function instead of an empty string.

pkg/gatherers/conditional/gathering_rule.schema.json
```diff
@@ -0,0 +1,127 @@
...
+    "required": [
+        "conditions",
+        "gathering_functions"
+    ],
+    "properties": {
+        "conditions": {
+            "type": "array",
+            "description": "The list of conditions on which the rule is triggered",
+            "minItems": 0,
+            "maxItems": 8,
```
tremes, line 38: Does it make sense to limit it now?

pkg/gatherers/conditional/gathering_rule.schema.json
```diff
@@ -0,0 +1,127 @@
...
+                            "params"
+                        ],
+                        "properties": {
+                            "type": {
+                                "type": "string",
+                                "title": "Type",
+                                "description": "Type of the condition alert_is_firing",
+                                "const": "alert_is_firing"
+                            },
+                            "params": {
```
tremes, line 60: Hmm now when I am thinking about this, it could be a bit confusing. `params` word sounds more like an array of params to me (which is not the case if I understand it correctly).

Do we allow more alert names (within one condition)? I guess there could be such condition in the future. How would it be defined?

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: If I remember correctly, getting image streams didn't work without these

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: lemme check

pkg/gatherers/conditional/gathering_rule.schema.json
Sergey1011010: I think it's better to limit everything we can

pkg/gatherers/conditional/gathering_rule.schema.json
Sergey1011010: It's called params because for other type of alert it can be different. We can define a rule requiring 2 alerts to fire like this:
```{
  "conditions": [
    {
      "type": "alert_is_firing",
      "params": {
        "name": "SamplesImagestreamImportFailing"
      }
    },
    {
      "type": "alert_is_firing",
      "params": {
        "name": "AnotherAlert"
      }
    }
  ],
  "gathering_functions": {
    "logs_of_namespace": {
      "namespace": "openshift-cluster-samples-operator",
      "tail_lines": 100
    },
    "image_streams_of_namespace": {
      "namespace": "openshift-cluster-samples-operator"
    }
  }
}```

File: pkg/gatherers/conditional/gathering_functions.go [hunk 10]
```diff
@@ -0,0 +1,52 @@
+package conditional
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// GatheringFunctions is a type to map gathering function name to its params
+type GatheringFunctions = map[GatheringFunctionName]interface{}
```
tremes, line 9: Wouldn't be good to define some type for the params? I guess we could use type embedding and the generic type could have hopefully something common for all the params... I guess name at least.

Sergey1011010: It seems redundant to include all the possible things in one type

pkg/gather/gather.go
Sergey1011010: fixed

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: changed to returning an error

pkg/config/config.go
Sergey1011010: fixed

pkg/gatherers/conditional/conditional_gatherer_test.go
Sergey1011010: fixed

pkg/gatherers/conditional/conditional_gatherer.go
tremes: OK thanks. Maybe we could mention it in a comment.

pkg/gatherers/conditional/gathering_rule.schema.json
tremes: OK thanks. Yes. It's probably just a detail, but maybe we could think about changing the `params` attribute to array..... I guess it would still have to be an array of objects.

pkg/gatherers/conditional/parsing.go
```diff
@@ -0,0 +1,65 @@
...
+		newParams, err := gatheringFunction.NewParams(jsonParams)
+		if err != nil {
+			return GatheringRule{}, err
+		}
+
+		result.GatheringFunctions[gatheringFunction] = newParams
+	}
+
+	return result, nil
+}
```
tremes, line 65: It would be nice to provide some comments for the functions here.

pkg/gatherers/conditional/gathering_functions.go
```diff
@@ -0,0 +1,52 @@
...
+	Namespace string `json:"namespace"`
+	// A number of log lines to keep for each container
+	TailLines int64 `json:"tail_lines"`
+}
+
+// GatherImageStreamsOfNamespaceParams defines parameters for image streams of namespace gatherer
+type GatherImageStreamsOfNamespaceParams struct {
+	// Namespace from which to collect image streams
+	Namespace string `json:"namespace"`
+}
```
tremes, line 52: These two types (`GatherImageStreamsOfNamespaceParams`, `GatherLogsOfNamespaceParams`) look quite similar. What about using something like:
```

type NamespaceParams struct {
	// Namespace from which to collect logs
	Namespace string `json:"namespace"`

	// A number of log lines to keep for each container
	TailLines int64 `json:"tail_lines,omitempty"`
}
```
I would like to avoid an explosion of *Params types. :)

pkg/gatherers/conditional/gathering_rule.schema.json
tremes: OK :)

pkg/gatherers/conditional/gathering_functions.go
tremes: Well I've been thinking about some inheritance (Java-ish :)) structure (in Go it's more composition though), but it looks like the params (e.g `GatherLogsOfNamespaceParams` and `AlertIsFiringConditionParams`) have nothing in common. So I guess let's keep it as it is.....but it's interesting problem... :)

pkg/gatherers/conditional/parsing.go
```diff
@@ -0,0 +1,65 @@
...
+	result := GatheringRule{
+		GatheringFunctions: make(GatheringFunctions),
+	}
+
+	for i := 0; i < len(rule.Conditions); i++ {
+		condition := &rule.Conditions[i]
+
+		jsonParams, err := json.Marshal(condition.Params)
+		if err != nil {
+			return GatheringRule{}, err
```
tremes, line 36: what about returning `nil, err` instead? Same for the error cases below.

pkg/gatherers/conditional/gathering_rule.schema.json
Sergey1011010: Why? Map fits better in my opinion, cuz we're not gonna have multiple names for an alert for example.

pkg/gatherers/conditional/gathering_functions.go
Sergey1011010: But tail lines has nothing to do with namespace, not sure if it's a good idea.

tremes: Hmm right :) I think I don't have any other idea right now how to make the params more generic. So not a blocker.

pkg/gatherers/conditional/gathering_rule.schema.json
tremes: I was thinking mainly of the size of the JSON file TBH. So something like:
```
"conditions": [
    {
      "type": "alert_is_firing",
      "params": [
        "name": "AlertA",
        "name": "AlertB", 
      ]
    },
  ],
```
This could be problematic from your point of view (why? validation?)?

Sergey1011010: Well, we can have something like:
```
  "conditions": [
    {
      "type": "alerts_are_firing",
      "params": [
        { "name": "AlertA" },
        { "name": "AlertB" }
      ]
    }
  ],
```
But we'd need to change the operator's code, the schema (it may become more complicated) and so on. The problem can be for some conditions which can't have multiple parameters of the same type, we'd need to decide on the behaviour when there's multiple params of the same type (can be use the latest, return an error, etc).

tremes: I see thanks. Good points. We don't have to address it right now, but I think we should keep that in mind (maybe new JIRA task?), because this requirement (having a condition with more alerts) is very likely to come IMO.

natiiix (commented):
I have looked throught the changes since the last time I have seen this PR a couple of weeks back. I definitely prefer this approach to accessing the conditional gatherer params compared to the old method. I don't really have any worthwhile criticism; it all seems fine and a logical improvement of the code to me.

PR #484, opened 2021-08-10
Bug 1991860: Don't try to record an empty Record if gatherClusterConfigV1 fails

mtrmac:
Currently, `gatherClusterConfigV1` returns `(Record{}, error)` on failure, and its caller appends the empty `Record` into its results; that later crashes in `Record.Filename()` because `Record.Item` is `nil`.

Instead, be more similar to the sibling function and return a `[]Record{}`, which allows `gatherClusterConfigV1` to return no records.

## Categories
- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive
See https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2695/pull-ci-openshift-machine-config-operator-master-e2e-aws-disruptive/1421081909886193664

## Unit Tests
N/A, the existing code doesn’t have unit tests of this nature.

## Privacy
N/A, no newly collected information.

## Changelog
?

## Breaking Changes
No

## References
None

### DIFF
--- a/pkg/gatherers/clusterconfig/cluster_config_v1_config_map.go
+++ b/pkg/gatherers/clusterconfig/cluster_config_v1_config_map.go
@@ -15,10 +15,10 @@ import (
 
 // gatherClusterConfigV1 gathers "cluster-config-v1" from "kube-system" namespace leaving only "install-config" from data.
 // "install-config" is anonymized.
-func gatherClusterConfigV1(ctx context.Context, coreClient corev1client.CoreV1Interface) (record.Record, []error) {
+func gatherClusterConfigV1(ctx context.Context, coreClient corev1client.CoreV1Interface) ([]record.Record, []error) {
 	configMap, err := coreClient.ConfigMaps("kube-system").Get(ctx, "cluster-config-v1", metav1.GetOptions{})
 	if err != nil {
-		return record.Record{}, []error{err}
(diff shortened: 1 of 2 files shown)


tremes:
@mtrmac Thank you, however it would be nice if you assign the BZ to yourself when you're working on that (or plan to work on that). I have already prepared https://github.com/tremes/insights-operator/commit/e34887197e8c23b32e4640b019b412082e45b563

mtrmac:
> @mtrmac Thank you, however it would be nice if you assign the BZ to yourself when you're working on that

I’m afraid I’m not going to to take the lead on this; this is a three-levels-deep dependency on some other work I’m actually doing, and I know basically nothing about the insights operator.

Feel free to treat this as a third-party drive-by PR; if you want to disconnect it from the bug, reprioritize the bug to some other time, or fix it in some other way, any of that is perfectly fine with me.

tremes:
@mtrmac No problem. The change is fine. I was only little bit surprised :) OK so let me take your commit and add mine there and I will open a new PR if you're OK with that.

tremes:
I took the @mtrmac's commit and created new https://github.com/openshift/insights-operator/pull/485. Thanks a lot!

PR #486, opened 2021-08-16
Bug 1995913: Degraded status in the OCM controller

tremes:
This adds status checking to the OCM controller. If the response from the OCM API is an HTTP error other than 404 (which means that SCA certs are not allowed for the given organization) then try again with exponential backoff and after few attempts mark the operator as degraded.  
At the same time, the status doesn't have to be changed to degraded in a disconnected environment. This means that any non-HTTP error is not considered as a reason for marking the operator as degraded.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No documentation update

## Unit Tests

No new test

## Privacy

## Breaking Changes

No

## References

https://bugzilla.redhat.com/show_bug.cgi?id=1995913

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -180,7 +180,8 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 			downloadReason = summary.Reason
 			downloadMessage = summary.Message
 		} else if summary.Operation == controllerstatus.PullingSCACerts {
-			// summary.Count
+			klog.V(4).Infof("Failed to download the SCA certs within the threshold %d with exponential backoff. Marking as degraded.",
+				uploadFailuresCountThreshold)
 			degradingFailure = true
 			ocmErrorMsg = summary.Message
 			ocmErrorReason = summary.Reason
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -144,8 +144,7 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
 	var last time.Time
 	var reason string
 	var errs []string
-	var uploadErrorReason, uploadErrorMessage, disabledReason, disabledMessage, downloadReason, downloadMessage,
-		ocmErrorReason, ocmErrorMsg string
+	var uploadErrorReason, uploadErrorMessage, disabledReason, disabledMessage, downloadReason, downloadMessage string
 	allReady := true
 	for i, source := range c.Sources() {
 		summary, ready := source.CurrentStatus()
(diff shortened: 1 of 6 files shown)


pkg/ocm/ocm.go
```diff
@@ -156,3 +184,40 @@ func (c *Controller) updateSecret(s *v1.Secret, ocmData *ScaResponse) (*v1.Secre
...
+			klog.Errorf("%v. Trying again in %s", httpErr, bo.Step())
+			return false, nil
+		}
+		return true, nil
+	})
+	// exp. backoff timeouted -> error
+	if err != nil {
+		return nil, fmt.Errorf("timed out waiting for the successful response from %s", endpoint)
+	}
+	return data, nil
```
tremes, line 222: This method is little bit complicated (especially due the fact that 404 and offline env. are not considered as errors from the "degraded" perspective) compared to counting the threshold in the [controllerstatus](https://github.com/openshift/insights-operator/blob/master/pkg/controllerstatus/controllerstatus.go#L32) and the checking it in the `status.go` (similarly as [here](https://github.com/openshift/insights-operator/blob/master/pkg/controller/status/status.go#L164))

The approach with exp. backoff is more "agressive":
- if HTTP error other than 404 then try again in 5 min (if this is still true then try again in 10 min, 20min, 40min => mark as degraded)

Compared to the other approach:
- if HTTP error other than 404 then only increase the count and do nothing
-  try again in default period (which is 8h in this case) if there's still an error then increase the count again
- if count higher than some threshold (let's say 2 or 3) then mark IO as degraded

WIth the first approach we can go degraded in ~1h15min, with the second one it would be after 16h-24h. I am not sure what is better TBH. The certs are valid for a long time (1 year by default), but on the other hand, if the content changes then you may want to update the certs quickly.

pkg/controller/status/status.go
```diff
@@ -300,6 +310,21 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
...
+				Reason:             ocmErrorReason,
+				Message:            ocmErrorMsg,
+			})
+		} else {
+			setOperatorStatusCondition(&existing.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+				Type:   configv1.OperatorDegraded,
+				Status: configv1.ConditionFalse,
+				Reason: "AsExpected",
+			})
+		}
```
tremes, line 327: Hmmm this block is probably not necessary and makes our tests failing... need to check

pkg/ocm/ocm.go
tremes: I updated the initial backoff step to `c.configurator.Config().OCMConfig.Interval / 32` (that is 15 minutes by default) so the times are bit longer than mentioned above.

PR #492, opened 2021-09-07
ApiRequestCount conditional gathering

tremes:
This is adding a new conditional gatherer.  If the `APIRemovedInNextEUSReleaseInUse` alert is firing then it looks at its labels, get the resource, version and group from them and finds the corresponding `APIRequestCounts` "record" for this resource. 

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/conditional/api_request_counts.json`

## Documentation

- `docs/gathered-data.md `

## Unit Tests

- ` pkg/gatherers/conditional/gather_api_request_count_test.go`

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-5581


pkg/gatherers/conditional/gather_api_request_count.go
```diff
@@ -0,0 +1,107 @@
...
+	LastDayRequestCount int64  `json:"last_day_request_count"`
+}
+
+// BuildGatherApiRequestCounts creates a gathering closure which collects logs api requests counts for the
+// resources mentioned in
+// Params is of type AlertIsFiringConditionParams:
+//   - alert_name string - name of the firing alert
+//
+//
+// * Location in archive: conditional/api_request_counts.json
```
tremes, line 30: I have to distinguish it, because I will have a second similar alert....naming is hard. What about:
`conditional/<alert_name>/api_request_counts.json` 
WDYT pls @Sergey1011010 ?

tremes: Well I have only one alert so it's simpler, but I am still not sure this is the best location.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -237,17 +253,28 @@ func (g *Gatherer) updateAlertsCacheFromClient(ctx context.Context, metricsClien
...
+			alert.Labels[label.GetName()] = label.GetValue()
 		}
+		alerts, ok := g.firingAlerts[alert.Name]
+		if ok {
+			alerts = append(alerts, alert)
+		} else {
+			alerts = []Alert{alert}
+		}
+
+		g.firingAlerts[alert.Name] = alerts
```
natiiix, line 277: Could be simplified into `g.firingAlerts[alert.Name] = append(g.firingAlerts[alert.Name], alert)` :wink:

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -86,8 +102,9 @@ type Gatherer struct {
 	gatherProtoKubeConfig   *rest.Config
 	metricsGatherKubeConfig *rest.Config
 	imageKubeConfig         *rest.Config
-	firingAlerts            map[string]bool // golang doesn't have sets :(
-	gatheringRules          []GatheringRule
+	// there can be multiple instances of the same alert
+	firingAlerts   map[string][]Alert // golang doesn't have sets :(
```
Sergey1011010, line 106: Just a small nitpick: we can already delete "golang doesn't have sets :(" comment, it's already a proper map :)

pkg/gatherers/conditional/gather_api_request_count.go
```diff
@@ -0,0 +1,106 @@
...
+				return records, err
+			}
+			return records, nil
+		},
+		CanFail: canConditionalGathererFail,
+	}, nil
+}
+
+func (g *Gatherer) gatherAPIRequestCounts(ctx context.Context, alertName string) ([]record.Record, []error) {
+	var resources []string
```
Sergey1011010, line 54: Maybe better to make it a set to improve performance of the for loop lil bit

pkg/gatherers/conditional/gathering_functions.go
```diff
@@ -13,13 +13,19 @@ type GatheringFunctions = map[GatheringFunctionName]interface{}
...
-// GatherImageStreamsOfNamespace is a function collecting image streams of the provided namespace.
-// See file gather_image_streams_of_namespace.go
-const GatherImageStreamsOfNamespace GatheringFunctionName = "image_streams_of_namespace"
+	// GatherImageStreamsOfNamespace is a function collecting image streams of the provided namespace.
+	// See file gather_image_streams_of_namespace.go
+	GatherImageStreamsOfNamespace GatheringFunctionName = "image_streams_of_namespace"
+
+	// GatherAPIRequestCounts is a function collecting api request counts for the resources read
+	// from the corresponding alert
+	GatherAPIRequestCounts GatheringFunctionName = "api_request_counts"
```
Sergey1011010, line 27: Maybe we could call it api_request_counts_of_alert or something similar to be more specific?

pkg/gatherers/conditional/gathering_rule.schema.json
```diff
@@ -120,6 +120,19 @@
...
+                "^api_request_counts$": {
+                    "type": "object",
+                    "title": "GatherApiRequestCountsParams",
+                    "required": [
+                        "alert_name"
+                    ],
+                    "properties": {
+                        "alert_name": {
+                            "type": "string",
+                            "title": "AlertName"
```
Sergey1011010, line 133: I think it's better to validate it as well like in `alert_is_firing` condition where we have the pattern `^[a-zA-Z0-9_]{1,128}$`

pkg/utils/slices.go
```diff
@@ -40,3 +40,13 @@ func TakeLastNItemsFromByteArray(array []byte, desiredLength int) []byte {
 
 	return array
 }
+
+// IsInSlice checks if string is in the slice of strings
+func IsInSlice(v string, a []string) bool {
```
Sergey1011010, line 45: we actually have `utils.StringInSlice`

pkg/gatherers/conditional/gather_api_request_count.go
tremes: you mean the https://github.com/openshift/insights-operator/pull/492/files#diff-768929cfc4894c47306f40d5abfcac538dd69ef7d32f7b0a347dc31ce6919660R79 right? Yeah I think I can use a set. Thanks

pkg/gatherers/conditional/gathering_rule.schema.json
tremes: OK why not

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -237,17 +253,21 @@ func (g *Gatherer) updateAlertsCacheFromClient(ctx context.Context, metricsClien
...
 			}
-
 			if label.GetName() == "alertname" {
-				g.firingAlerts[label.GetValue()] = true
+				alert.Name = label.GetValue()
+				continue
 			}
+			alert.Labels[label.GetName()] = label.GetValue()
 		}
+		g.firingAlerts[alert.Name] = append(g.firingAlerts[alert.Name], alert)
```
natiiix, line 270: Is there any reason to have the `Name` as an explicit parameter of the `Alert` type when the only way to access it is by its name in the `firingAlerts` map? You can also get the name from the `Labels` and this way there's an extra level of a type nested into the map that doesn't really seem very necessary to me, as `Name` is the only parameter is has beside the label map. I thought you would just create a new type alias `AlertLabels` -> `map[string]string` and that `firingAlerts` would be `map[string][]AlertLabels`.

natiiix: Looking at the rest of the code, I don't see `Name` being actually used anywhere in a significant way, right? Couldn't the `Alert` type be simplified to just a string:string map to only store the labels, then?

tremes: I guess no special reason :) OK let me update it

Sergey1011010:
looks great

natiiix (approved):
Tested locally according to the steps provided using the Service Mesh operator, which resulted in `conditional/alerts/APIRemovedInNextEUSReleaseInUse/api_request_counts.json` file being successfully generated after an hour.
/lgtm

PR #495, opened 2021-09-13
Bug 2003657:  Respect user defined proxy's CA cert

tremes:
This checks if some proxy is defined for the Insights Operator and if so then:

1.  Looks at the cluster-wide proxy spec `clusterProxy.Spec.TrustedCA.Name` to get the  name of the config map with the CA cert
2.  Gets the corresponding config map from the `openshift-config` namespaces
3. Tries to read the `ca-bundle.crt` key from the config map
4. Adds the CA bundle cert to the CA cert pool used in the `insightsclient`

**How to reproduce**
It's quite easy to reproduce the actual problem (you can set some proxy in the cluster-wide proxy and observe the IO error), but it's a bit difficult to verify that this fix works. You can check the https://access.redhat.com/articles/4740031. In general you need to do the following:

1.  Deploy the mitmproxy to a cluster
2. Extract the `mitmproxy-ca-cert.pem` from the mitmproxy container (see `oc exec -n mitmproxy mitmproxy cat /root/.mitmproxy/mitmproxy-ca-cert.pem > /tmp/mitmproxy-ca-cert.pem` (It happened that the mitmproxy container was sometimes killed in my case. **Note that you have to extract the cert again in such case!**)
3. Create a new config map in `openshift-config` namespace containing the cert (see `oc create cm user-ca-bundle -n openshift-config --from-file=ca-bundle.crt=/tmp/mitmproxy-ca-cert.pem`
3. Patch the cluster-wide proxy `spec.TrustedCA` to point to the new config map (see ` oc patch proxy cluster --type=json -p '[{"op":"add","path":"/spec/trustedCA","value":{"name":"user-ca-bundle"}}]'`)
4. Configure the proxy env vars for the IO container:
```
MY_NO_PROXY=$(oc get network cluster --template '{{(index .status.serviceNetwork 0)}},{{(index .status.clusterNetwork 0).cidr}},localhost')
oc set env deployment/insights-operator HTTP_PROXY=http://mitmproxy.mitmproxy.svc:8080 HTTPS_PROXY=http://mitmproxy.mitmproxy.svc:8080 NO_PROXY=$MY_NO_PROXY
```
The IO upload should work fine and it should be visible in the mitmproxy log. Image containing this fix is available at quay.io/tremes/respect_proxy_ca
 

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No doc update

## Unit Tests

No test. We can think of some integration tests, but we would probably need to deploy mitmproxy container.

## Privacy

## Breaking Changes

No

## References

https://bugzilla.redhat.com/show_bug.cgi?id=2003657 (originally reported in https://bugzilla.redhat.com/show_bug.cgi?id=1995937)

### DIFF
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -31,6 +31,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachineryversion "k8s.io/apimachinery/pkg/version"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	"github.com/openshift/insights-operator/pkg/authorizer"
 )
@@ -120,9 +121,9 @@ func getTrustedCABundle() (*x509.CertPool, error) {
 }
 
 // clientTransport creates new http.Transport with either system or configured Proxy
-func clientTransport(authorizer Authorizer) http.RoundTripper {
+func (c *Client) clientTransport() http.RoundTripper {
 	clientTransport := &http.Transport{
-		Proxy: authorizer.NewSystemOrConfiguredProxy(),
+		Proxy: c.authorizer.NewSystemOrConfiguredProxy(),
 		DialContext: (&net.Dialer{
 			Timeout:   30 * time.Second,
 			KeepAlive: 30 * time.Second,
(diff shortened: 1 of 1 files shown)


natiiix (commented):
I've checked the code and discussed it a bit, but I didn't have time to test it locally yet, so I'm not giving a definitive approval at the moment. Sorry, but the process is quite complicated and I didn't have enough time today.

natiiix (approved):
I have tested the provided image on a testing cluster in combination with the `mitmproxy` and I was able to see that IO has successfully detected the proxy cetrificate and that it was correctly uploading archives.
/lgtm

PR #496, opened 2021-09-13
[release-4.9] ApiRequestCount conditional gathering (#492)

tremes:
This is adding a new conditional gatherer. If the APIRemovedInNextEUSReleaseInUse alert is firing then it looks at its labels, get the resource, version and group from them and finds the corresponding APIRequestCounts "record" for this resource.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/conditional/api_request_counts.json`

## Documentation

- `docs/gathered-data.md `

## Unit Tests

- `pkg/gatherers/conditional/gather_api_request_count_test.go`

## Privacy

## Breaking Changes

No

## References


pkg/gatherers/conditional/gather_api_request_count.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (g *Gatherer) gatherAPIRequestCounts(ctx context.Context,
+	dynamicClient dynamic.Interface, alertName string) ([]record.Record, []error) {
+	resources := make(map[string]struct{})
+	for _, labels := range g.firingAlerts[alertName] {
+		resourceName := fmt.Sprintf("%s.%s.%s", labels["resource"], labels["version"], labels["group"])
+		resources[resourceName] = struct{}{}
+	}
+
+	gvr := schema.GroupVersionResource{Group: "apiserver.openshift.io", Version: "v1", Resource: "apirequestcounts"}
```
rluders, line 65: I'm always confused about this one... 'cause we do have some resources declared inside the `const.go`, but in some cases we have it declared on its own gather file. Which is the right place?

tremes: Yes. Good point. Well we don't have any `const.go` in the conditional gatherer yet and we don't want to introduce a dependency to `clusterconfig` one..... I guess

rluders: Yeh, I don't block the merge, I already approved it, but... it would be nice if we could decide where do we want to store these declarations. Maybe we could change it in some other PR later.

PR #498, opened 2021-09-14
Bug 2012069: Refactoring Status controller

rluders:
This PR changes the `status.go` organizing the logic into smaller pieces making it a little bit better.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

None

## Documentation

None

## Unit Tests

None

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-5350
https://bugzilla.redhat.com/show_bug.cgi?id=2012069


pkg/controller/status/status.go
```diff
@@ -252,71 +289,15 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
...
 
-	// once the operator is running it is always considered available
-	setOperatorStatusCondition(&existing.Status.Conditions, configv1.ClusterOperatorStatusCondition{
-		Type:   configv1.OperatorAvailable,
-		Status: configv1.ConditionTrue,
-		Reason: "AsExpected",
-	})
-
-	// update the Progressing condition with a summary of the current state
+func (c *Controller) updateProcessingConditionWithSummary(existing *configv1.ClusterOperator,
```
tremes, line 299: It looks like it doesn't have to be a receiver method. Same for the `updateDisabledAndFailingConditions`

pkg/controller/status/status.go
```diff
@@ -359,19 +340,70 @@ func (c *Controller) merge(existing *configv1.ClusterOperator) *configv1.Cluster
...
+	} else {
+		setOperatorStatusCondition(&existing.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+			Type:   configv1.OperatorDegraded,
+			Status: configv1.ConditionFalse,
+			Reason: "AsExpected",
+		})
+	}
+}
+
+func setDisabledOperatorStatusCondition(existing *configv1.ClusterOperator, disabledMessage string, disabledReason string) {
```
tremes, line 392: These new functions (with `setUploadDegradedOperatorStatusCondition`,`setDegradedOperatorStatusCondition`,`setInsightsDownloadDegradedOperatorStatusCondition`) are basically the same. What about to use only one or just create the corresponding `ClusterOperatorStatusCondition` first and then only pass it to `setOperatorStatusCondition` function.

rluders: Yep! Thanks. This is probably the next step. I'm trying to do something about the `*reasons` and `*messages` to make it a little bit clean before cleaning up those methods.

tremes:
I think I would move [related objects settings](https://github.com/openshift/insights-operator/pull/498/files#diff-031531dfaf87d2bf33e39ad07de90f5edb6893005134968e24aad963c11ce237R225-R236) to separate function

pkg/controller/status/status.go
rluders: Agreed. In the beginning, I was planning to move the `last` to the struct properties, but... after doing it... well, it doesn't seem right.

tremes:
Some inspiration can be found in the [cluster-monitoring-operator](https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/client/status_reporter.go) :-)

pkg/controller/status/controller.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+// Controller is the type responsible for managing the statusMessage of the operator according to the statusMessage of the sources.
+// Sources come from different major parts of the codebase, for the purpose of communicating their statusMessage with the controller.
+type Controller struct {
+	name      string
+	namespace string
+
+	client     configv1client.ConfigV1Interface
+	coreClient corev1client.CoreV1Interface
```
tremes, line 53: This doesn't seem to be used here.

pkg/controller/status/controller.go
```diff
@@ -0,0 +1,411 @@
...
+	// prime the object if it does not exist
+	if clusterOperator == nil {
+		clusterOperator = &configv1.ClusterOperator{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: c.name,
+			},
+		}
+	}
+
+	ctrlStatus := newControllerStatus()
```
tremes, line 140: This means that every time when the `merge` is called then there's a new map created right? It doesn't seem right to me and I am not sure it will work correctly and I am also affraid of some leak.

pkg/controller/status/controller.go
```diff
@@ -0,0 +1,411 @@
...
+			cs.setCondition(configv1.OperatorDegraded, configv1.ConditionFalse, "AsExpected", "", metav1.Now())
+		}
+
+	default: // once we've initialized set Failing and Disabled as best we know
+		// handle when disabled
+		if ds := ctrlStatus.getStatus(DisabledStatus); ds != nil {
+			cs.setCondition(OperatorDisabled, configv1.ConditionTrue, ds.reason, ds.message, metav1.Now())
+		} else {
+			cs.setCondition(OperatorDisabled, configv1.ConditionFalse, "AsExpected", "", metav1.Now())
+		}
```
tremes, line 346: This if-else statement will be executed every time regardless `isInitializing` or not right? If so then we could move it out of the `switch` statement.

pkg/controller/status/controller.go
```diff
@@ -0,0 +1,411 @@
...
+		// handle when download fails
+		if ds := ctrlStatus.getStatus(DownloadStatus); ds != nil {
+			cs.setCondition(InsightsDownloadDegraded, configv1.ConditionTrue, ds.reason, ds.message, metav1.Time{Time: lastTransition})
+		} else {
+			cs.removeCondition(InsightsDownloadDegraded)
+		}
+	}
+}
+
+func updateProcessingConditionWithSummary(cs *conditions, ctrlStatus *controllerStatus,
```
tremes, line 372: I am missing go-doc here and above.

pkg/controller/status/controller.go
rluders: Shouldn't the GC take care of it? I think so... I didn't get any problem during my tests, but it wasn't a long run.
Maybe pass it to the struct would be a better approach. What do you think?

pkg/controller/status/controller.go
```diff
@@ -0,0 +1,410 @@
...
+	var errorReason string
+	var errs []string
+
+	allReady = false
+
+	for i, source := range c.Sources() {
+		summary, ready := source.CurrentStatus()
+		if !ready {
+			klog.V(4).Infof("Source %d %T is not ready", i, source)
+			allReady = true
```
tremes, line 182: This doesn't seem right. If the source is not ready then `allReady = true` doesn't make sense IMO.

tremes: Well it seems it's not working at the moment. From the CI:
>: operator conditions insights 
Operator unavailable (null): operator is not reporting conditions

I am also missing some unit tests for `conditions.go`, `controller.go`.

rluders: You are right. Nice catch, thank you.

pkg/controller/status/controller.go
```diff
@@ -0,0 +1,424 @@
...
+		if len(summary.Message) == 0 {
+			klog.Errorf("Programmer error: statusMessage source %d %T reported an empty message: %#v", i, source, summary)
+			continue
+		}
+
+		degradingFailure := false
+
+		if summary.Operation == controllerstatus.Uploading {
+			if summary.Count < uploadFailuresCountThreshold {
+				klog.V(4).Infof("Number of lastTransition upload failures %d lower than threshold %d. Not marking as degraded.",
```
tremes, line 199: > Number of lastTransition upload failures .... 
Please preserve the same message as before

pkg/controller/status/controller.go
```diff
@@ -0,0 +1,424 @@
...
+					summary.Count, uploadFailuresCountThreshold)
+			} else {
+				degradingFailure = true
+				klog.V(4).Infof("Number of lastTransition upload failures %d exceeded than threshold %d. Marking as degraded.",
```
tremes, line 203: > .... exceeded than threshold
I think this was wrong even before so we can fix it

pkg/controller/status/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	// cluster operator conditions
+	cs := newConditions(&clusterOperator.Status, metav1.Time{Time: now})
+	updateDisabledAndFailingConditions(cs, ctrlStatus, isInitializing, lastTransition)
+
+	// once the operator is running it is always considered available
+	cs.setCondition(configv1.OperatorAvailable, configv1.ConditionTrue, "AsExpected", "", metav1.Now())
+
+	updateProcessingConditionWithSummary(cs, ctrlStatus, isInitializing, lastTransition)
+
+	klog.V(4).Infof("The operator is healthy")
```
tremes, line 152: I think we shouldn't always print this to log. An example from my log shows it:
```
I0923 10:19:56.005695       1 controller.go:203] Number of lastTransition upload failures 5 exceeded than threshold 5. Marking as degraded.
I0923 10:19:56.005733       1 controller.go:336] The operator has some internal errors: Unable to report: gateway server reported unexpected error code: 404 (request=259e010ba9a44a07ae7ab367ab2d2b0e): 404 page not found
I0923 10:19:56.005742       1 controller.go:371] The operator has some internal errors: 
I0923 10:19:56.005747       1 controller.go:152] The operator is healthy
````
Also the line with `The operator has some internal errors: ` is missing something. I guess the actual error.

File: pkg/controller/status/conditions.go [hunk 11]
```diff
@@ -0,0 +1,119 @@
...
+			Reason:             "",
+		},
+		configv1.OperatorDegraded: {
+			Type:               configv1.OperatorDegraded,
+			Status:             configv1.ConditionUnknown,
+			LastTransitionTime: time,
+			Reason:             "",
+		},
+		configv1.OperatorUpgradeable: {
+			Type:               configv1.OperatorUpgradeable,
```
tremes, line 44: I think we didn't have this condition before and I think it's never set so we can probably remove it.

pkg/controller/status/controller_test.go
```diff
@@ -0,0 +1,82 @@
...
+func Test_Status_SaveInitialStart(t *testing.T) {
+	tests := []struct {
+		name                     string
+		clusterOperator          *configv1.ClusterOperator
+		expErr                   error
+		initialRun               bool
+		expectedSafeInitialStart bool
+	}{
+		{
+			name:                     "Non-initial run is has upload delayed",
```
tremes, line 27: This name is a little weird :) Please fix.

pkg/controller/status/controller.go
tremes: I am not sure, but I would rather not rely on it. We would need to test it and probably profile it. Anyway it should be feasible to create the map once and only maintain its content.

pkg/controller/status/controller.go
tremes: Ohh I realized it's twice there....we can keep only the first one then :)

rluders: OK. But now I'm wondering why the second one doesn't have the reason. :thinking: I guess that I'll take a look just to make sure.

pkg/controller/status/controller.go
```diff
@@ -131,6 +131,9 @@ func (c *Controller) merge(clusterOperator *configv1.ClusterOperator) *configv1.
 		clusterOperator = newClusterOperator(c.name, nil)
 	}
 
+	// make sure to start a clean status controller
+	c.ctrlStatus.reset()
```
rluders, line 135: @tremes So, after a few tests (and fighting with the e2e tests), it seems that this is the best solution. We keep an instance to the status controller into the controller (controller everywhere) and then it encapsulates the logic to handle the statuses. But, it needs to be reset every time that the merge is executed, otherwise, the operator status would be wrong.

pkg/controller/status/controller.go
tremes: This discussion was resolved, but the problem was not fully resolved. I am still seeing:
```
I1005 11:36:23.480363       1 controller.go:374] The operator has some internal errors: Unable to report: gateway server reported unexpected error code: 404 (request=8d7a0361f47641168e5d8da1b1598e0c): 404 page not found
I1005 11:36:23.480373       1 controller.go:157] The operator is healthy
```
I believe this was not case before to have one message saying there's some error and second message saying the operator is healthy.

pkg/controller/status/controller.go
tremes: yes I am looking at it and it seem we are reseting everything (also the line https://github.com/openshift/insights-operator/pull/498/files#diff-10c33b0f428af9470aef2747e42fc97d97150d338b151e836d9cdebbf4fbef42R149) eveyrtime....

pkg/controller/status/controller.go
natiiix: ^ That's indeed pretty weird and doesn't seem right, but I feel like IO sometimes really reports `The operator is healthy` even after printing an error (not necessarily from this code, but anything that happens before the message is logged).

pkg/controller/status/controller.go
rluders: Yeh, but in this case it is correct. We are "resetting" 'cause the `merge` method receives the `clusterOperator` when it is called.

pkg/controller/status/controller.go
rluders: You are right. I made a mistake here. Already fixed, let me just test it and check if I'm not forgetting anything. I got a little bit confused about the way that the `switch` should work, I guess...

rluders: Now, it should be fine.

pkg/controller/status/conditions_test.go
```diff
@@ -0,0 +1,468 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_conditions_entries(t *testing.T) {
+	time := metav1.Now()
+
+	type fields struct {
+		entryMap conditionsMap
+	}
```
tremes, line 16: Why do you need this type here? You should be able to use `conditions` IMO.

tremes: Same for all the tests below.

pkg/controller/status/conditions_test.go
```diff
@@ -0,0 +1,468 @@
...
+
+func Test_conditions_findCondition(t *testing.T) {
+	time := metav1.Now()
+
+	type fields struct {
+		entryMap conditionsMap
+	}
+	type args struct {
+		condition configv1.ClusterStatusConditionType
+	}
```
tremes, line 74: this type is little bit redundant IMO

pkg/controller/status/status_test.go
```diff
@@ -1,89 +1,185 @@
...
 )
 
-func Test_Status_SaveInitialStart(t *testing.T) {
+func Test_controllerStatus_getStatus(t *testing.T) {
+	type fields struct {
+		statusMap map[string]statusMessage
+	}
+	type args struct {
+		id string
+	}
```
tremes, line 14: Same here the type `fields` is not necessary IMO and the `args` is little bit redundant IMO.

pkg/controller/status/status_test.go
```diff
@@ -1,89 +1,185 @@
...
+			name: "Update existing status with same status message",
+			fields: fields{statusMap: map[string]statusMessage{
+				"upload":   {reason: "upload reason", message: "upload message"},
+				"disabled": {reason: "disabled reason", message: "disabled message"},
+			}},
+			args: args{id: DisabledStatus, reason: "disabled reason", message: "disabled message"},
+			want: &controllerStatus{statusMap: map[string]statusMessage{
+				"upload":   {reason: "upload reason", message: "upload message"},
+				"disabled": {reason: "disabled reason", message: "disabled message"},
+			}},
```
tremes, line 156: We could probably use the constants defined in `status.go`.

pkg/controller/status/conditions_test.go
rluders: Elaborate.

natiiix (commented):
Aside from the commend from @tremes about "The operator is healthy" being logged right after logging an error, I see nothing wrong. The code is alright and at least somewhat commented. I tested this version of IO locally and it ran just fine. Now giving a final approval yet to give others a chance to further review the PR.

pkg/controller/status/status.go
```diff
@@ -1,497 +1,62 @@
...
-func findOperatorStatusCondition(conditions []configv1.ClusterOperatorStatusCondition,
-	conditionType configv1.ClusterStatusConditionType) *configv1.ClusterOperatorStatusCondition {
-	for i := range conditions {
-		if conditions[i].Type == conditionType {
-			return &conditions[i]
-		}
-	}
-	return nil
+func (c *controllerStatus) isHealthy() bool {
+	return !(c.hasStatus(ErrorStatus) || c.hasStatus(DisabledStatus))
```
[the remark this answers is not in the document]
rluders: Yep. Correct. Already fixed it.

pkg/controller/status/conditions_test.go
tremes: I think it's not necessary to wrap `condition configv1.ClusterStatusConditionType` to a new type. IOW you can use directly the `configv1.ClusterStatusConditionType` IMO

rluders: I see... I prefer to keep the `args` and `fields` wrapped. It makes the code more standardized test structures.

tremes: `fields` and why pls? I can't see any benefit TBH. It adds only unnecessary parts to the test IMO. Using `conditions` you can omit the https://github.com/openshift/insights-operator/pull/498/files#diff-6cd7b045851c73c1120b823f3d07a763bfb7e20aa20e241ac5d3e27e862d57eaR119-R121 and just do `tt.conditions.findCondition(tt.args.condition)` then.

PR #500, opened 2021-09-14
Bug 2004814: OCM controller - change type of the secret

tremes:
This changes the type of the `etc-pki-entitlement` secret to opaque. It also change the attribute names to end with `pem` suffix so that it can be easily consumed by the Build API. 

## Categories

- [X]  Bugfix
- [ ] Enhancement
- [ ] Backporting
- [ ] Others (CI, Infrastructure, Documentation)

## Sample Archive

No new data

## Documentation

No doc update

## Unit Tests

Corresponding unit tests updated.

## Privacy

Yes. There are no sensitive data in the newly collected information.

## Breaking Changes

No

## References

https://issues.redhat.com/browse/???
https://bugzilla.redhat.com/show_bug.cgi?id=2004814
https://access.redhat.com/solutions/???

PR #506, opened 2021-09-17
[release-4.9] Bug 2005338: OCM controller - change type of the secret (#500)

tremes:
This changes the type of the `etc-pki-entitlement` secret to opaque. It also change the attribute names to end with pem suffix so that it can be easily consumed by the Build API. if there is already a same secret of type TLS then delete it first and re-create as opaque secret. 

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

No doc update

## Unit Tests

Unit tests udated

## Privacy

## Breaking Changes

No

## References

https://bugzilla.redhat.com/show_bug.cgi?id=2005338

### DIFF
--- a/pkg/ocm/ocm.go
+++ b/pkg/ocm/ocm.go
@@ -20,8 +20,10 @@ import (
 )
 
 const (
-	targetNamespaceName = "openshift-config-managed"
-	secretName          = "etc-pki-entitlement" //nolint: gosec
+	targetNamespaceName    = "openshift-config-managed"
+	secretName             = "etc-pki-entitlement" //nolint: gosec
+	entitlementAttrName    = "entitlement.pem"
+	entitlementKeyAttrName = "entitlement-key.pem"
 )
 
 // Controller holds all the required resources to be able to communicate with OCM API
@@ -146,6 +148,18 @@ func (c *Controller) checkSecret(ocmData *ScaResponse) error {
 		return err
 	}
 
+	if scaSec.Type == v1.SecretTypeTLS {
+		err = c.coreClient.Secrets(targetNamespaceName).Delete(c.ctx, scaSec.GetName(), metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+		_, err = c.createSecret(ocmData)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
 	_, err = c.updateSecret(scaSec, ocmData)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


rluders (approved):
/lgtm

rluders:
If CI approves...
/lgtm

pkg/ocm/ocm.go
```diff
@@ -146,37 +148,49 @@ func (c *Controller) checkSecret(ocmData *ScaResponse) error {
...
+		if err != nil {
+			return err
+		}
+		err = c.createSecret(ocmData)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
```
stleerh, line 162: Is it possible to do the TLS check before checking the existence of secret (line 140)?  That way you don't need to duplicate the code at lines 156-160.  Then the logic is:

If TLS secret exists, delete it.  
If no secret, create one.

tremes: Hmm... I don't think so (or I don't know how to do that). The `CoreV1Interface` doesn't allow you to get or list secrets of specific type (pls correct me if I am wrong)

stleerh: This is minor, and I am not familiar with the code, but wouldn't this simplify the logic?

\`func (c \*Controller) checkSecret(ocmData \*ScaResponse) error {  
   scaSec, err := c.coreClient.Secrets(targetNamespaceName).Get(c.ctx, secretName, metav1.GetOptions{})  
   if err != nil {  
       return err  
   }

   if scaSec.Type == v1.SecretTypeTLS {  
       delErr := c.coreClient.Secrets(targetNamespaceName).Delete(c.ctx, scaSec.GetName(), metav1.DeleteOptions{})  
       if delErr != nil {  
           return delErr  
       }  
   }

   // if the secret doesn't exist then create one  
   if errors.IsNotFound(err) {  
       return c.createSecret(ocmData)  
   }\`

tremes: if the secret is deleted then I have to created. Other than that your snippet is moving the ` if errors.IsNotFound(err)` few lines down, which is only "cosmetic" change IMO (and I think it's better to handle the error immediately).

It's true that I can return `c.createSecret(ocmData)` and save few lines with the error checking. I can update it if you want...but again a minor change IMO.

stleerh: Okay, will let this go.

tremes: OK thanks a lot.

PR #509, opened 2021-09-22
Conditional gatherer of logs of unhealthy pods

natiiix:
This PR adds a new conditional gatherer that gathers pod logs based on firing alerts. It looks for `KubePodCrashlooping` and `KubePodNotReady`, and collects the appropriate log from their corresponding pods/containers.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `pkg/gatherers/conditional/gathering_rule.schema.json`

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gatherers/conditional/gather_logs_of_unhealthy_pods_test.go`

## Privacy

We are already gathering pod logs elsewhere and this gatherer should only be triggered for pods/alerts from `openshift-*` namespaces.

## Changelog

No.

## Breaking Changes

No, there are only additions; no modification to existing gathering.

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-5499


pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
```diff
@@ -0,0 +1,99 @@
...
+				continue
+			}
+			alertPod, ok := alertLabels["pod"]
+			if !ok {
+				klog.Warningf("alert is missing 'pod' label")
+				continue
+			}
+			alertContainer, ok := alertLabels["container"]
+			if !ok {
+				klog.Warningf("alert is missing 'container' label")
```
tremes, line 74: Minor suggestion. I think it would be good to propagate these messages to our metadata as errors (including the alert name).

tremes:
@sferich888 I think the answer is yes. We already have some gatherers collecting various container logs and I believe some of them are really important - e.g https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md#clusteroperatorpodsandevents. Other examples are:
- https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md#openshiftsdncontrollerlogs
- https://github.com/openshift/insights-operator/blob/master/docs/gathered-data.md#openshiftsdnlogs
- and more

Also note that these logs in this PR are gathered only when the corresponding alert is firing.

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
```diff
@@ -0,0 +1,108 @@
...
+			errs = append(errs, newErr)
+			continue
+		}
+		// The container label may not be present for all alerts (e.g., KubePodNotReady).
+		alertContainer := alertLabels["container"]
+
+		containerFilter := ""
+		if alertContainer != "" {
+			containerFilter = fmt.Sprintf("^%s$", alertContainer)
+		}
```
tremes, line 77: It's just cosmetic detail, but maybe doing:
```
		if alertContainer, ok := alertLabels["container"]; ok {
			containerFilter = fmt.Sprintf("^%s$", alertContainer)
		}
```
would be more straightforward

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
```diff
@@ -0,0 +1,108 @@
...
+			common.LogMessagesFilter{
+				TailLines: params.TailLines,
+				Previous:  params.Previous,
+			},
+			func(namespace string, podName string, containerName string) string {
+				logKind := "current"
+				if params.Previous {
+					logKind = "previous"
+				}
+				return fmt.Sprintf("%s/unhealthy_logs/%s/%s/%s/%s.log", g.GetName(), namespace, podName, containerName, logKind)
```
tremes, line 94: I think this path doesn't fit very well to what we have in https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/conditional/gather_logs_of_namespace.go#L69. Would it make sense to use the same path?

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	if len(errs) > 0 {
+		t.Fatalf("unexpected error(s) returned by the log gathering function: %v", errs)
+	}
+	if len(rec) != 1 {
+		t.Fatalf("unexpected number of records (expected: 1, actual: %d)", len(rec))
+	}
+
+	if rec[0].Name != "conditional/unhealthy_logs/test-namespace/test-pod/test-container/previous.log" {
+		t.Fatalf("unexpected 'Name' of the second log record: %q", rec[0].Name)
```
tremes, line 97: Again cosmetic one:
> ....of the previous log record
maybe better....

tremes:
I didn't try to reproduce the alerts to trigger this new gatherer. I added few minor suggestions, but it looks very good! Thanks. You need to satisfy the linting.

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
```diff
@@ -0,0 +1,108 @@
...
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/insights-operator/pkg/gatherers"
+	"github.com/openshift/insights-operator/pkg/gatherers/common"
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// BuildGatherLogsOfUnhealthyPods collects either current or previous logs for pods firing one of the configured alerts.
+//
+// * Location in archive: conditional/unhealthy_logs/<namespace>/<pod>/<container>/[current|previous].log
```
Sergey1011010, line 18: so far, in conditional gatherer we use format where we put the type of the item first and then the name (for example `conditional/namespaces/{namespace}/pods/{pod_name}/containers/{container_name}/logs/last-{n}-lines.log`), I'd suggest to follow this concept

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
```diff
@@ -0,0 +1,108 @@
...
+	}, nil
+}
+
+func (g *Gatherer) gatherLogsOfUnhealthyPods(
+	ctx context.Context, coreClient v1.CoreV1Interface, params GatherLogsOfUnhealthyPodsParams,
+) ([]record.Record, []error) {
+	errs := []error{}
+	records := []record.Record{}
+
+	alertInstances, ok := g.firingAlerts[params.AlertName]
```
Sergey1011010, line 52: Shouldn't it be considered an error?

pkg/gatherers/conditional/gathering_rule.schema.json
```diff
@@ -134,6 +134,31 @@
...
+                "^logs_of_unhealthy_pods$": {
+                    "type": "object",
+                    "title": "GatherLogsOfUnhealthyPodsParams",
+                    "required": [
+                        "alert_name",
+                        "tail_lines",
+                        "previous"
+                    ],
+                    "properties": {
+                        "alert_name": {
```
Sergey1011010, line 147: Add validation please (`"pattern": "^[a-zA-Z0-9_]{1,128}$"`)

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
tremes: ok but what about previous/current log? Would it be `conditional/namespaces/{namespace}/pods/{pod_name}/containers/{container_name}/logs/previous.log` or `conditional/namespaces/{namespace}/pods/{pod_name}/containers/{container_name}/logs/previous-last-{n}-lines.log`?

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
tremes: What exactly? That the alert couldn't be found in the map?

pkg/gatherers/conditional/gathering_functions.go
```diff
@@ -27,6 +27,8 @@ const (
 	GatherAPIRequestCounts GatheringFunctionName = "api_request_counts_of_resource_from_alert"
 )
 
+const GatherLogsOfUnhealthyPods GatheringFunctionName = "logs_of_unhealthy_pods"
```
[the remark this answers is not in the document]
natiiix: Well, I would like if this gatherer could be used in general for any case of an alert where we want to gather logs from the corresponding pod, so the only specification I could add would be that it's based on alerts.

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
tremes: I agree the condition evaluation is basically the same (check if it's in the map) and the map should be read-only once  it's created.

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
Sergey1011010: I'd do `conditional/namespaces/{namespace}/pods/{pod_name}/containers/{container_name}/previous-logs/last-{n}-lines.log`, but `previous-last-{n}-lines.log` would also work

tremes: ok both works for me as well. So the suggestion is to update it to use one of these suggested paths and then deal with possible duplicates in a new spike right? I think I am fine with this approach. @natiiix WDYT?

pkg/gatherers/conditional/gather_logs_of_unhealthy_pods.go
tremes: well ok. Fair point....it's more a config error for me, but we can add some message to inform about this.

rluders (approved):
/lgtm

PR #514, opened 2021-10-05
Increment the "insightsclient_request_recvreport_total" metric only w…

tremes:
…hen a new report is correctly received

This is updating/fixing the logic when the `insightsclient_request_recvreport_total` prometheus metric is incremented. 

The problem was that it's not accurate to increment this metric in case of same downloaded report (basically after [this check](https://github.com/openshift/insights-operator/blob/master/pkg/insights/insightsreport/insightsreport.go#L130)). At the same time we need to update the metric in every error case, which complicates it a little. 

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No doc update

## Unit Tests

No new test

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -287,7 +287,7 @@ func (c *Client) Send(ctx context.Context, endpoint string, source Source) error
 }
 
 // RecvReport perform a request to Insights Results Smart Proxy endpoint
-func (c Client) RecvReport(ctx context.Context, endpoint string) (io.ReadCloser, error) {
+func (c Client) RecvReport(ctx context.Context, endpoint string) (*http.Response, error) {
 	cv, err := c.getClusterVersion()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/insights/insightsclient/insightsclient.go
```diff
@@ -424,6 +428,11 @@ func ocmErrorMessage(url *url.URL, r *http.Response) error {
 	}
 }
 
+// IncrementRecvReportMetric increments the "insightsclient_request_recvreport_total" metric for the given HTTP status code
+func (c *Client) IncrementRecvReportMetric(statusCode int) {
```
rluders, line 432: Does it need to be part of the `Client`? IMO, I'll make it just a function that receives the `metricsName` and the `statusCode`, and also makes it private to the package. What do you think?

rluders (approved):
/lgtm

PR #517, opened 2021-10-12
Collecting node logs

rluders:
This PR implements a new data enhancement to collect the node logs from the journal unit (kubelet).

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/config/node/logs/ip-10-0-135-228.us-east-2.compute.internal.log`

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gatherers/clusterconfig/node_logs_test.go` (WIP)

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-5831


pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,130 @@
...
+// nodeLogResourceURI creates the resource path URI to be fetched
+func nodeLogResourceURI(client rest.Interface, name string) string {
+	uri := client.Get().
+		Name(name).
+		Resource("nodes").SubResource("proxy", "logs").
+		Suffix("journal").URL().Path
+
+	if strings.HasSuffix("journal", "/") {
+		uri += "/"
+	}
```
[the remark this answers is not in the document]
rluders: Well, not really. It's 'cause the path may omit the end slash, so, basically, I'm just checking it and if there is no ending slash adding it. I don't think that left it without the ending slash may cause any issues, but... there is no harm.

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,130 @@
...
+	for {
+		// gzipped stream
+		_, err = io.CopyN(buf, gz, 1024)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return "", err
+		}
+	}
```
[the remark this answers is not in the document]
rluders: In this case, I'm handling two different possibilities. The first one that you see, using the `io.Copy` handles the non-gzipped request, so if for some reason, even asking for the gzip content, it fails to uncompress it, 'cause it is not compressed, so we can simply use the `io.Copy`. Otherwise, we have it as gzipped content, `gosec` says that there is a possibility of DDoS attacks. After some research, I found out that is recommended to use the `io.CopyN` to solve it.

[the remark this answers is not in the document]
tremes: Well OK, but we don't expect a malicious content from the API server right? :) I would be in favor of ignoring the `gosec` in this case (although it's a detail again) or we can consume the stream in some different way.....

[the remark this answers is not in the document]
rluders: I have no strong opinion about `io.Copy` or `io.CopyN` usage. Both works for me. Using just one could simplify the code a little bit, I guess, but I will have to split it into a separated function, I guess... 'cause we still need to handle the possibility to have the non-gzipped response, even always asking for the gzipped. Let me see what can be done.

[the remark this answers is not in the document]
rluders: BTW, I added one small check to `io.EOF`, 'cause I noticed that if EOF we don't want to handle it as an error...

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,149 @@
...
+	return client.Get().RequestURI(uri).
+		SetHeader("Accept", "text/plain, */*").
+		SetHeader("Accept-Encoding", "gzip").
+		Param("tail", strconv.Itoa(tail)).
+		Param("unit", unit)
+}
+
+// nodeLogString retrieve the data from the stream, decompress it (if necessary) and return the string
+func nodeLogString(req *rest.Request, out *bytes.Buffer, size int) (string, error) {
+	in, err := req.Stream(context.TODO())
```
tremes, line 105: I think we should not use `context.TODO()`, but rather use the context passed to the gatherer.

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,149 @@
...
+
+	_, err = gzip.NewReader(bytes.NewBuffer(head))
+	if err != nil {
+		// not gzipped stream
+		_, err = io.Copy(out, buf)
+		if err != nil && err != io.EOF {
+			return "", err
+		}
+
+		return out.String(), nil
```
tremes, line 125: So if it's not gzip stream then we return only first 1024 bytes? It seems a bit inconsistent comparing to filtering below.

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,149 @@
...
+		return "", err
+	}
+
+	// nolint: gosec
+	_, err = io.Copy(out, r)
+	if err != nil {
+		return "", err
+	}
+
+	scanner := bufio.NewScanner(out)
```
tremes, line 139: I am failing to see why you copy the buffer above. I think you should be able to do `bufio.NewScanner(in)` here or not?

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,149 @@
...
+	messagesToSearch := []string{
+		"\\[\\d+\\]: E\\d{4,} [0-9]{1,2}:[0-9]{1,2}:[0-9]{1,2}", //  Errors from log
```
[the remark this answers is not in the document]
tremes: It looks like the most important part is ` E\\d{4,}` (can there be more than 4 numbers?). I think it makes sense to match the time after this string, but maybe we can omit the `"\\[\\d+\\]:`.

pkg/gatherers/clusterconfig/node_logs.go
rluders: Nope, it will not. The returned buffer has the same size for both cases, gzipped or un-gzipped. The peek is only used to validate if the stream is gzipped or not. It basically the same logic that the `oc` uses to get it done.

pkg/gatherers/clusterconfig/node_logs.go
rluders: Waaait a minute. I think that I forgot something here.

rluders: But, answering your question. Nope, 'cause I need to handle both un-gzipped and gzipped sources, but I think that I forgot to standardize the method return, this is why you are getting it wrong.

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			return "", err
+		}
+
+		// nolint: gosec
+		_, err = io.Copy(out, r)
+		if err != nil {
+			return "", err
+		}
+	}
```
tremes, line 135: Sorry but I still don't understand why all this is needed. I think you can do just something like:
```
	r, err := gzip.NewReader(in)
	if err != nil {
		scanner = bufio.NewScanner(in)
	} else {
		scanner = bufio.NewScanner(r)
	}
```
or not?

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// GatherNodeLogs fetches the node logs from journal unit
+//
+// Response see https://docs.openshift.com/container-platform/4.8/rest_api/node_apis/node-core-v1.html#apiv1nodesnameproxypath
+//
+// * Location in archive: config/nodes/logs/
+// * See: docs/insights-archive-sample/config/nodes/logs
+// * Id in config: node_logs
```
tremes, line 38: It would be good to add the OCP version

pkg/gatherers/clusterconfig/node_logs.go
rluders: It doesn't work, u need to "uncompress" the streaming first. Otherwise, it will crash. This logic is basically the same that is being used for the `oc` command. I trust them.

pkg/gatherers/clusterconfig/node_logs.go
rluders: Good point.

pkg/gatherers/clusterconfig/node_logs.go
tremes: Well it seems to work fine in my cluster...

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	"bufio"
+	"bytes"
+	"compress/gzip"
+	"context"
+	"fmt"
+	"io"
+	"strconv"
+
+	"github.com/openshift/insights-operator/pkg/gatherers/common"
```
Sergey1011010, line 12: I thought we had a convention where we have 3 groups of imports: from the standard library, external packages, operator's code

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,149 @@
...
+	return gatherNodeLogs(ctx, clientSet.CoreV1())
+}
+
+func gatherNodeLogs(ctx context.Context, client corev1client.CoreV1Interface) ([]record.Record, []error) {
+	nodes, err := client.Nodes().List(ctx, metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/master"})
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	records, errs := nodeLogRecords(ctx, client.RESTClient(), nodes)
```
Sergey1011010, line 55: Can be just `return nodeLogRecords(...)`

Sergey1011010 (commented):
looks great

pkg/gatherers/clusterconfig/node_logs.go
rluders: OK, so, maybe I can simplify it. Let me give it a try.

pkg/gatherers/clusterconfig/node_logs.go
rluders: What do you mean by that?

pkg/gatherers/clusterconfig/node_logs.go
rluders: Nope. I tested doesn't crash, but what I get are empty files.

pkg/gatherers/clusterconfig/node_logs.go
Sergey1011010: I mean that I would group imports like this:
```
    "bufio"
	"bytes"
	"compress/gzip"
	"context"
	"fmt"
	"io"
	"strconv"

	corev1 "k8s.io/api/core/v1"
	"k8s.io/klog/v2"
	"k8s.io/client-go/rest"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/client-go/kubernetes"
	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"

	"github.com/openshift/insights-operator/pkg/gatherers/common"
	"github.com/openshift/insights-operator/pkg/recorder"
	"github.com/openshift/insights-operator/pkg/utils/marshal"
```
but maybe we don't have any convention for it

pkg/gatherers/clusterconfig/node_logs.go
tremes: Can you please share your code? It worked for me and the files were not empty

pkg/gatherers/clusterconfig/node_logs.go
tremes: I think we don't...but it's true that it's grouped differently than here. I mean into 3 groups (as Serhii mentioned) where the last two somehow intersect. Isn't this part of `go fmt` or our linting??

rluders: Gotcha! I can check if it's possible to add this validation to the lining. I think that the pre-commit should format the code before pushing it. I need to check why it is not happening.

rluders: OK. So, a quick investigation point that we have two different styles of coding format to import: [check here](https://github.com/golang/go/issues/37719) and [here](https://goinbigdata.com/goimports-vs-gofmt/). If we want to use the one by goimports I'll check how to do linting it and add it to the pre-commit script later. For now, the better approach is to configure the IDE to do it.

pkg/gatherers/clusterconfig/node_logs.go
rluders: Fixed.

tremes: Thank you.

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	defer in.Close()
+
+	r, err := gzip.NewReader(in)
+	var scanner *bufio.Scanner
+	if err != nil {
+		scanner = bufio.NewScanner(in)
+	} else {
+		scanner = bufio.NewScanner(r)
+	}
```
tremes, line 103: We should also close `r` please.

rluders: done

pkg/gatherers/clusterconfig/node_logs.go
```diff
@@ -0,0 +1,115 @@
...
+		SetHeader("Accept-Encoding", "gzip").
+		Param("tail", strconv.Itoa(tail)).
+		Param("unit", unit)
+}
+
+// nodeLogString retrieve the data from the stream, decompress it (if necessary) and return the string
+func nodeLogString(ctx context.Context, req *rest.Request) (string, error) {
+	in, err := req.Stream(ctx)
+	if err != nil {
+		return "", err
```
tremes, line 93: This can probably lead to an empty file/record, but it's probably not a big deal.

tremes:
Reviewed and tested few times. Thank you!

pkg/gatherers/clusterconfig/node_logs.go
rluders: I guess that It is possible. Don't you think that it should be handled by the recorder? I mean, the recorder should not record empty files. Right?

tremes: Probably...

PR #524, opened 2021-10-20
Cluster version condition

Sergey1011010:
## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No changes

## Documentation

No changes

## Unit Tests

## Privacy

No new data was collected

## Changelog

No

## Breaking Changes

Yes

## References

https://issues.redhat.com/browse/CCXDEV-5773

### DIFF
--- a/pkg/gatherers/conditional/conditional_gatherer.go
+++ b/pkg/gatherers/conditional/conditional_gatherer.go
@@ -3,6 +3,10 @@
 // they can be fetched from outside, checked that they make sense (we want to check the parameters, for example if
 // a rule tells to collect logs of a namespace on firing alert, we want to check that the namespace is created
 // by openshift and not by a user). Conditional gathering isn't considered prioritized, so we run it every 6 hours.
+//
+// To add a new condition, follow the next steps:
+// 1. Add structures to conditions.go
+// 2. Change areAllConditionsSatisfied function in conditional_gatherer.go
 package conditional
 
 import (
@@ -12,7 +16,10 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/blang/semver/v4"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/prometheus/common/expfmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
@@ -158,13 +165,15 @@ type Gatherer struct {
 	gatherProtoKubeConfig   *rest.Config
 	metricsGatherKubeConfig *rest.Config
 	imageKubeConfig         *rest.Config
+	gatherKubeConfig        *rest.Config
 	// there can be multiple instances of the same alert
 	firingAlerts   map[string][]AlertLabels
 	gatheringRules []GatheringRule
+	clusterVersion string
 }
 
 // New creates a new instance of conditional gatherer with the appropriate configs
-func New(gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config) *Gatherer {
+func New(gatherProtoKubeConfig, metricsGatherKubeConfig, gatherKubeConfig *rest.Config) *Gatherer {
 	var imageKubeConfig *rest.Config
 	if gatherProtoKubeConfig != nil {
 		// needed for getting image streams
@@ -177,6 +186,7 @@ func New(gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config) *Gatherer 
 		gatherProtoKubeConfig:   gatherProtoKubeConfig,
 		metricsGatherKubeConfig: metricsGatherKubeConfig,
 		imageKubeConfig:         imageKubeConfig,
+		gatherKubeConfig:        gatherKubeConfig,
 		gatheringRules:          defaultGatheringRules,
 	}
 }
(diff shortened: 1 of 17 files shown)


pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -272,12 +303,41 @@ func (g *Gatherer) updateAlertsCacheFromClient(ctx context.Context, metricsClien
 	return nil
 }
 
+func (g *Gatherer) updateVersionCache(ctx context.Context, configClient configv1client.ConfigV1Interface) error {
+	clusterVersion, err := configClient.ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	g.clusterVersion = clusterVersion.Spec.DesiredUpdate.Version
```
tremes, line 312: I think this is not correct. Is it? `DesiredUpdate` attribute/field is optional. I think we want `clusterVersion.Status.Desired.Version`

tremes:
@Sergey1011010 Please update the PR template. This is not an enhancement.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -247,10 +265,19 @@ func (g *Gatherer) updateAlertsCache(ctx context.Context) error {
...
+	if err := g.updateAlertsCache(ctx, metricsClient); err != nil { //nolint:govet
+		return err
+	}
+
+	configClient, err := configv1client.NewForConfig(g.gatherKubeConfig)
+	if err != nil {
+		return err
+	}
+
+	return g.updateVersionCache(ctx, configClient)
```
tremes, line 277: I guess this logic here basically means that if the `updateAlertsCache` fails (we couldn't connect to the metrics for some reason) then the `updateVersionCache` is not executed. Is that right? Does it mean that the conditions for cluster version will not be evaluated when we can't connect to the metrics?

pkg/gatherers/conditional/gathering_rule.schema.json
```diff
@@ -74,6 +74,40 @@
...
                             }
                         }
+                    },
+                    {
+                        "type": "object",
+                        "title": "ConditionWithParams",
+                        "description": "cluster_version_matches condition",
+                        "required": [
+                            "type",
+                            "params"
```
tremes, line 84: shouldn't this be `cluster_version_matches` ? And same below

natiiix: @tremes Could you please clarify what you mean? Because as far as I can see, the description is consistent with the `alert_is_firing` condition description field and the `required` fields match the fields defined below. Am I missing something?

tremes: AFAIK the validation was failing for me. Maybe I was doing something wrong or just missed something, but I think the naming here has to match the naming in https://github.com/openshift/insights-operator/pull/524/files#diff-4e4038b9223da9895b1e6eda486f7d60a3453f9e39e36aefc3d35a3c80d14c78R7 otherwise the validation fails.

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: > I guess this logic here basically means that if the updateAlertsCache fails (we couldn't connect to the metrics for some reason) then the updateVersionCache is not executed. Is that right?

Yes, but there's no point to update version cache if we don't have valid alerts cache since current rules contain alerts anyway. 

> Does it mean that the conditions for cluster version will not be evaluated when we can't connect to the metrics?

Yup, conditional gatherer won't do it's job if it can't initialize any of the caches. We'd get the error in the metadata tho.

tremes: > Yes, but there's no point to update version cache if we don't have valid alerts cache since current rules contain alerts anyway.
I am not sure I follow. We don't have any conditional rules with version yet. I think it's difficult to predict if they will have any alerts as well. If they will then it probably makes sense, but if they will not then it seems as unnecessary limitation.

Sergey1011010: Updated changing the logic little bit so we would get more info in the archive

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -177,10 +186,18 @@ func New(gatherProtoKubeConfig, metricsGatherKubeConfig *rest.Config) *Gatherer
...
+		gatherKubeConfig:        gatherKubeConfig,
 		gatheringRules:          defaultGatheringRules,
 	}
 }
 
+// GatheringRuleMetadata stores information about gathering rules
+type GatheringRuleMetadata struct {
+	Rule         GatheringRule `json:"rule"`
+	Errors       []error       `json:"errors"`
+	WasTriggered bool          `json:"was_triggered"`
```
tremes, line 198: Great. I like this idea!

tremes:
@Sergey1011010 Please fix the linting and then I will approve. Good job! Thanks!

Sergey1011010:
@tremes done

pkg/gatherers/conditional/conditional_gatherer.go
tremes: Thanks

tremes:
I reviewed few times and it looks good to me. I like the information in the metadata! Thanks!
/lgtm

PR #529, opened 2021-11-02
Gather DVO metrics

natiiix:
This PR implements an enhancement for gathering and filtering metrics from Deployment Validation Operator's metrics service.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/config/dvo_metrics`

## Documentation

- `docs/gathered-data.md` - "DVOMetrics" section

## Unit Tests

- `pkg/utils/read_lines_with_prefix_test.go` contains a unit test for an important part of the functionality

## Privacy

## Changelog

No.

## Breaking Changes

No

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-5856


pkg/gatherers/clusterconfig/dvo_metrics.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/insights-operator/pkg/record"
+	"github.com/openshift/insights-operator/pkg/utils/marshal"
+)
+
+func (g *Gatherer) GatherDVOMetrics(ctx context.Context) ([]record.Record, []error) {
+	apiURL, err := url.Parse("http://deployment-validation-operator-metrics.deployment-validation-operator.svc:8383")
```
tremes, line 16: I think it would be better to find the service name and port programmatically. In other words, rely only on the namespace name. 
Maybe we could also suggest renaming of the service - simply to `metrics`.

natiiix: @tremes Yeah, I suppose we shouldn't access it like this, but is there really any reason why/how it could change to anything else? Perhaps it would be worth bringing up on the next DVO sync. Dynamic service discovery seems to me like an additional complexity that may not really be necessary.

tremes: Well my point was that you would never know when someone will change the name or the port in the [service](definition)- As for the name, it's true that we will have to rely either on the fact that there's only one service, or simply try to find a service whose name contains `metrics`.

natiiix: Okay, so I hopefully added that functionality in f201c50.

pkg/gatherers/clusterconfig/clusterconfig_gatherer.go
```diff
@@ -48,6 +48,7 @@ func failableFunc(function gathererFuncPtr) gatheringFunction {
 var gatheringFunctions = map[string]gatheringFunction{
 	"pdbs":                              failableFunc((*Gatherer).GatherPodDisruptionBudgets),
 	"metrics":                           failableFunc((*Gatherer).GatherMostRecentMetrics),
+	"dvo_metrics":                       importantFunc((*Gatherer).GatherDVOMetrics),
```
tremes, line 51: I think we can mark this as failable. I also think we should create a task for removing this flag, because it's not necessary IMO

pkg/gatherers/clusterconfig/dvo_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+				klog.Errorf("Unable to read metrics lines with DVO prefix: %v", err)
+				return nil, []error{err}
+			}
+
+			allDVOMetricsLines = append(allDVOMetricsLines, prefixedLines...)
+		}
+	}
+
+	records := []record.Record{
+		{Name: "config/dvo_metrics_filtered", Item: marshal.RawByte(allDVOMetricsLines)},
```
tremes, line 96: I would call it just `dvo_metrics`

pkg/gatherers/clusterconfig/dvo_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, []error{err}
+	}
+
+	allDVOMetricsLines := []byte{}
+
+	for _, service := range serviceList.Items {
+		if !dvoMetricsServiceNameRegex.MatchString(service.Name) {
+			continue
+		}
+		for _, port := range service.Spec.Ports {
```
tremes, line 63: Hmmm this is interesting. I didn't think about the option of service with more ports specified....Maybe it will be slightly better to expect (i.e fail otherwise) that there's only one port specified than doing/trying to do more requests otherwise. WDYT?

pkg/gatherers/clusterconfig/dvo_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+			apiURL := url.URL{
+				Scheme: "http",
+				Host:   fmt.Sprintf("%s.%s.svc:%d", service.Name, dvoNamespace, port.Port),
+			}
+			metricsRESTClient, err := rest.NewRESTClient(&apiURL, "/", rest.ClientContentConfig{}, rateLimiter, http.DefaultClient)
+			if err != nil {
+				klog.Warningf("Unable to load metrics client, no metrics will be collected: %v", err)
+				return nil, nil
```
tremes, line 71: We can return the error here as well.

pkg/gatherers/clusterconfig/clusterconfig_gatherer.go
natiiix: Oh yeah, my bad, you're definitely right about that.

pkg/gatherers/clusterconfig/dvo_metrics.go
natiiix: Yep, I've already changed that back. The only reason I temporarily changed the name was so that I could see the cluster was actually running the correct version of IO, because I suspected it wasn't pulling the latest version of the image (which was true due to an incorrectly set image pull policy on the deployment).

pkg/gatherers/clusterconfig/dvo_metrics.go
natiiix: I don't really know. Is it a problem to make more than just one request? I'm just thinking there could be some kind of an odd hybrid metrics service that exposes multiple ports instead of having multiple endpoints, or something wild like that.

tremes:
I reviewed and experimented with this one. It works good! Thank you!
/lgtm

sferich888:
@davemulford to review for PX /lgtm

PR #532, opened 2021-11-04
Bug 2020250: Replacing deprecated ioutil

rluders:
This PR replaces the deprecated `ioutil` as recommended by [golang 1.16 changelog](https://golang.org/doc/go1.16#ioutil) and makes our lint happy.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

n/a

## Documentation

n/a

## Unit Tests

n/a

## Privacy

## Breaking Changes

n/a

## References

### DIFF
--- a/pkg/recorder/diskrecorder/diskrecorder.go
+++ b/pkg/recorder/diskrecorder/diskrecorder.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
@@ -146,8 +147,10 @@ func (d *DiskRecorder) Summary(_ context.Context, since time.Time) (io.ReadClose
 		return nil, false, nil
 	}
 	recentFiles := make([]string, 0, len(files))
+
+	var fileInfo fs.FileInfo
 	for _, file := range files {
-		fileInfo, err := file.Info()
+		fileInfo, err = file.Info()
 		if err != nil {
 			return nil, false, err
 		}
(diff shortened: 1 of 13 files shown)


pkg/recorder/diskrecorder/diskrecorder.go
```diff
@@ -144,10 +144,11 @@ func (d *DiskRecorder) Summary(_ context.Context, since time.Time) (io.ReadClose
 	}
 	recentFiles := make([]string, 0, len(files))
 	for _, file := range files {
-		if isNotArchiveFile(file) {
+		fileInfo, _ := file.Info()
```
tremes, line 147: We should not omit the error. Same above.

pkg/recorder/diskrecorder/diskrecorder.go
```diff
@@ -102,17 +102,21 @@ func (d *DiskRecorder) SaveAtPath(records record.MemoryRecords, path string) (re
...
 	if err != nil {
 		return err
 	}
 	count := 0
 	var errors []string
 	for _, file := range files {
-		if isNotArchiveFile(file) {
+		fileInfo, err := file.Info()
+		if err != nil {
+			errors = append(errors, err.Error())
```
tremes, line 114: Don't we want to rather log an error and continue when the file doesn't exist? I guess `fileInfo` will be `nil` if the file doesn't exist and it will probably cause a panic later on or is that wrong assumption?

rluders: Let me think...

tremes:
Reviewed and looks reasonable. Thanks
/lgtm

PR #534, opened 2021-11-04
Bug 2005816: make projectid and region anonymization consistent

Sergey1011010:
## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browse/CCXDEV-5970

### DIFF
--- a/pkg/gatherers/clusterconfig/image_registries.go
+++ b/pkg/gatherers/clusterconfig/image_registries.go
@@ -137,14 +137,15 @@ func anonymizeImageRegistry(config *registryv1.Config) *registryv1.Config {
 	if config.Spec.Storage.IBMCOS != nil {
 		anonymizeIBMCOSStorage(config.Spec.Storage.IBMCOS)
 	}
+
 	if config.Status.Storage.S3 != nil {
 		anonymizeS3Storage(config.Status.Storage.S3)
 	}
-	if config.Status.Storage.GCS != nil {
-		anonymizeGCSStorage(config.Status.Storage.GCS)
-	}
 	if config.Status.Storage.Azure != nil {
 		anonymizeAzureStorage(config.Status.Storage.Azure)
+	}
+	if config.Status.Storage.GCS != nil {
+		anonymizeGCSStorage(config.Status.Storage.GCS)
 	}
 	if config.Status.Storage.Swift != nil {
 		anonymizeSwiftStorage(config.Status.Storage.Swift)
@@ -152,6 +153,7 @@ func anonymizeImageRegistry(config *registryv1.Config) *registryv1.Config {
 	if config.Status.Storage.IBMCOS != nil {
 		anonymizeIBMCOSStorage(config.Status.Storage.IBMCOS)
 	}
+
 	// kubectl.kubernetes.io/last-applied-configuration annotation contains complete previous resource definition
 	// including the sensitive information as bucket, keyIDs, etc.
 	if lac, ok := config.Annotations[lacAnnotation]; ok {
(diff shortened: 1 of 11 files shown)


pkg/gatherers/clusterconfig/image_registries.go
```diff
@@ -137,21 +137,23 @@ func anonymizeImageRegistry(config *registryv1.Config) *registryv1.Config {
...
 	}
-	if config.Status.Storage.GCS != nil {
-		anonymizeGCSStorage(config.Status.Storage.GCS)
-	}
 	if config.Status.Storage.Azure != nil {
 		anonymizeAzureStorage(config.Status.Storage.Azure)
 	}
+	if config.Status.Storage.GCS != nil {
+		anonymizeGCSStorage(config.Status.Storage.GCS)
+	}
```
tremes, line 149: The changes here are not required right :)

tremes:
@Sergey1011010 This won't be so easy. The `projectID` and `region` is on more places. E.g `./config/configmaps/kube-system/cluster-config-v1.json` and then nodes, machinesets,..... as mentioned in the BZ.

natiiix (commented):
It seems quite alright to me just by looking at the code, but shouldn't there be at least some tests, or is this one of those cases where it's very difficult to create a unit test due to the fake API limitations?

Sergey1011010:
@natiiix well, it just extends the existing anonymization which wasn't tested before as well.

pkg/utils/anonymize/unstructured.go
```diff
@@ -0,0 +1,22 @@
...
+	if err != nil {
+		return err
+	}
+	if !found {
+		return fmt.Errorf("unable to find field '%v'", fields)
+	}
+
+	valueStr, _ := value.(string)
+	return unstructured.SetNestedField(data, String(valueStr), fields...)
+}
```
tremes, line 22: I was doing something very similar in `pkg/utils/unstructerd_wrappers.go` (but without  anonymization/setting the field) so maybe we could move it there. WDYT?

tremes:
@Sergey1011010 Correct, but it would be nice to add at least something - e.g the infrastucture gatherer (there's no dynamic client used) and `cluster_config_v1_config_map.go` is yours right :-P :) (maybe we could merge it with the `configs_map.go` gatherer, which also has some more "advanced" tests)

tremes:
There's also existing BZ for this so you can update the PR please.

pkg/utils/anonymize/unstructured.go
Sergey1011010: I think it's relates more to anonymize package than to utils.

Sergey1011010:
@tremes do you mean this one?

tremes:
@Sergey1011010 yes sir. Thanks

pkg/gatherers/clusterconfig/machine_sets.go
```diff
@@ -52,9 +55,27 @@ func gatherMachineSet(ctx context.Context, dynamicClient dynamic.Interface) ([]r
...
 		})
 	}
 
 	return records, nil
 }
+
+func anonymizeMachineset(data *unstructured.Unstructured) *unstructured.Unstructured {
+	fieldsToAnonymize := [][]string{
+		{"spec", "template", "spec", "providerSpec", "value", "projectID"},
+		{"spec", "template", "spec", "providerSpec", "value", "region"},
```
tremes, line 68: I was experimenting with adding one more line:
`{"spec", "template", "spec", "providerSpec", "value", "serviceAccounts"},`
but this will render empty `serviceAccounts` (because there's slice of strings not a string). It would be nice to anonymize the serviceAccounts as well IMO, because there's an email address.

Sergey1011010: done

tremes:
Reviewed & verified. Thank you!
/lgtm

PR #545, opened 2021-11-11
feat: alertmanager conditional log gathering

rluders:

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

Yes/No

## References


pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -79,6 +81,7 @@ var defaultGatheringRules = []GatheringRule{
 			},
 		},
 	},
+	// GatherAPIRequestCounts
```
rluders, line 84: BTW, I added these comments to help to identify the conditions for each gather. It was a little bit hard for me to read it.

natiiix: Makes sense and I agree that these lists tend to be difficult to read, especially as they grow longer.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -128,6 +132,39 @@ var defaultGatheringRules = []GatheringRule{
...
+				},
+			},
+		},
+		GatheringFunctions: GatheringFunctions{
+			GatherAlertmanagerLogs: GatherAlertmanagerLogsParams{
+				AlertName: "AlertmanagerClusterFailedToSendAlerts",
+				TailLines: 50,
+			},
+		},
+	},
```
natiiix, line 151: It's unfortunate that the conditional gatherer doesn't pass the alert name to the gatherer function is some way because it leads to this ugly code duplication that would be very easy to accidentally mess up (forgetting to set both alert name strings when copy-pasting). Not an issue with this PR, just a general note.

pkg/gatherers/conditional/gather_alertmanager_logs.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/openshift/insights-operator/pkg/record"
+
+	"github.com/openshift/insights-operator/pkg/gatherers"
+)
+
+// BuildGatherAlertmanagerLogs collects alertmanager logs for pods firing one the configured alerts.
+//
+// * Location in archive: conditional/namespaces/<namespace>/pods/<pod>/containers/<container>/logs/alertmanager-<event-name>.log
+// * Since versions:
+//   * 4.10+
```
natiiix, line 23: Most enhancements/gatherers also include the "Id in config" (internal code-name of the gatherer) and occassionally the path of a sample file in the sample archive. Just something to consider adding.

pkg/gatherers/conditional/gather_alertmanager_logs.go
```diff
@@ -0,0 +1,127 @@
...
+			common.LogMessagesFilter{
+				TailLines: params.TailLines,
+			},
+			func(namespace string, podName string, containerName string) string {
+				return fmt.Sprintf(
+					"%s/namespaces/%s/pods/%s/containers/logs/alertmanager-%s.log",
+					g.GetName(),
+					namespace,
+					podName,
+					strings.ToLower(params.AlertName),
```
natiiix, line 93: Any particular reason for converting the alert name to lowercase? Isn't it be more difficult to read that way?

pkg/gatherers/conditional/gather_alertmanager_logs.go
```diff
@@ -0,0 +1,127 @@
...
+
+func getAlertPodNamespace(labels AlertLabels) (string, error) {
+	namespace, ok := labels["namespace"]
+	if !ok {
+		newErr := fmt.Errorf("alert is missing 'namespace' label")
+		klog.Warningln(newErr.Error())
+		return "", newErr
+	}
+	return namespace, nil
+}
```
natiiix, line 127: We may want to add these to some kind of conditional gathering utils package/file because I'm doing something very similar in my conditional logs gatherer.

natiiix (commented):
Aside from a couple of small details, I don't see any particular issue with this PR. Please let me know if you want to make any of the suggested changes. Otherwise, I could probably approve it as is.

pkg/gatherers/conditional/gather_alertmanager_logs.go
```diff
@@ -0,0 +1,127 @@
...
+				Namespace:                alertPodNamespace,
+				FieldSelector:            fmt.Sprintf("metadata.name=%s", alertPodName),
+				ContainerNameRegexFilter: "^alertmanager$",
+			},
+			common.LogMessagesFilter{
+				TailLines: params.TailLines,
+			},
+			func(namespace string, podName string, containerName string) string {
+				return fmt.Sprintf(
+					"%s/namespaces/%s/pods/%s/containers/logs/alertmanager-%s.log",
```
tremes, line 89: The path is missing the container name here (same for the sample archive). It should be `"%s/namespaces/%s/pods/%s/containers/%s/logs/alertmanager-%s.log"` to follow the set precedent. 

We also call the log files as `last-100-lines.log` and similar.

pkg/gatherers/conditional/conditional_gatherer.go
rluders: Yeh, I was thinking exactly the same when I was implementing it. Maybe it would be a good idea to create a task to look for a better approach. What do you think?

pkg/gatherers/conditional/gather_alertmanager_logs.go
rluders: Personal preference. I don't like to handle "came-case" filenames. But I guess that [this one](https://github.com/openshift/insights-operator/pull/545#discussion_r755791973) overlaps this discussion.

pkg/gatherers/conditional/conditional_gatherer.go
tremes, line 151: The first condition seems a bit unnecessary IMO. The problem is that the `AlertmanagerClusterFailedToSendAlerts` doesn't provide any "pod" info and thus it always fails at https://github.com/openshift/insights-operator/pull/545/files#diff-9a6e0f2be79b4492a5ed592bd137bcb21a989988ccfc185f16796c550ceedd3fR70 and we also have https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/conditional/conditional_gatherer.go#L133 (not sure why) so I am not sure about all consequences....

rluders: Let me run some tests, but I think that you are right, we may not need the first alert.

rluders: I ran some tests and read some docs, and from what I can say I didn't notice any issues to remove the first alert and keep just the second one.

tremes:
I experimented with this one and reviewed couple of times. Looks good! Thank you!
/lgtm

PR #548, opened 2021-11-15
DO NOT MERGE: Add a name analysis command

smarterclayton:
STATUS: Draft, purely for exploration locally at this point

The `workloadinfo` command performs some transformation passes on
the config/workload_info.json file that makes it more amenable to
human exploration. The first implemented pass takes a dictionary of
names that represent the namespace, firstArg, or firstCommand fields
prior to hashing, creates an index between the hash and the source
name, and then replaces those inline into the workload info struct
which is serialized to stdout. The `--verbose` flag will print some
stats about how many names were identified. Any non-matching hashes
are left as is in the output.

A helper flag `--extract-names` was added to bulk transform exemplar
command lines (as might be found in image metadata or pod command and
args fields) into names for inlining.

Expose the necessary fields and types from pkg/gatherers/workloads
to avoid reuse.

This command would be used in the future to perform other
transformations for human use such as taking known maps of image
SHAs to metadata, or calculating other stats about usage from a
workload info file.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

Does not change gathering.

## Documentation

Not intended for public documentation yet.

## Unit Tests

No tests needed.

## Privacy

## Changelog

No.

## Breaking Changes

No.

## References


rluders:
@smarterclayton could you please make it a Draft? Just trying to not use the CI, unless really necessary. Thank you.

tremes:
Hi @smarterclayton any plans with this PR?

smarterclayton:
Yes, I want us to do analysis of the data set and this would be the best way to keep that close.

PR #551, opened 2021-11-24
adding gatherer for collecting silenced alerts

prashbnair:
Signed-off-by: Prashant Balachandran <pnair@redhat.com>

This PR adds a new gatherer which can collect silenced alerts from the alertmanager.

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Documentation
- `docs/gathered-data.md`

## References

https://issues.redhat.com/browse/CCXDEV-6423

### DIFF
--- a/manifests/03-clusterrole.yaml
+++ b/manifests/03-clusterrole.yaml
@@ -77,6 +77,26 @@ rules:
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: insights-operator-alertmanager
+  namespace: openshift-monitoring
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/single-node-developer: "true"
+roleRef:
+  kind: Role
+  name: monitoring-alertmanager-edit
+subjects:
+  - kind: ServiceAccount
+    namespace: openshift-insights
+    name: operator
+
+---
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: insights-operator
@@ -203,7 +223,7 @@ rules:
     verbs:
       - get
       - list
-      - watch    
+      - watch
   - apiGroups:
       - logging.openshift.io
     resources:
@@ -235,7 +255,7 @@ rules:
     verbs:
       - get
       - list
-      - watch    
+      - watch
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/pkg/gatherers/clusterconfig/silenced_alerts.go
+++ b/pkg/gatherers/clusterconfig/silenced_alerts.go
@@ -1,0 +1,46 @@
+package clusterconfig
+
+import (
+	"context"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+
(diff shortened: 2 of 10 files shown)


pkg/gatherers/clusterconfig/silenced_alerts.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		klog.Warningf("Unable to load alerts client, no alerts will be collected: %v", err)
+		return nil, nil
+	}
+
+	return gatherSilencedAlerts(ctx, alertsRESTClient)
+}
+
+func gatherSilencedAlerts(ctx context.Context, alertsClient rest.Interface) ([]record.Record, []error) {
+	data, err := alertsClient.Get().AbsPath("api/v2/alerts?silenced=true&active=false&inhibited=false").
```
tremes, line 31: Can you please elaborate on the params here? Why `active=false` and `inhibited=false`? :)

prashbnair: The default value for all query parameters is true. As we only need alerts that are silenced, only that parameter is set to true.
https://github.com/prometheus/alertmanager/blob/main/api/v2/openapi.yaml#L143

prashbnair:
/retitle adding gatherer for collecting silenced alerts

manifests/03-clusterrole.yaml
```diff
@@ -74,6 +74,13 @@ rules:
       - namespaces
     verbs:
       - get
+  # allow the operator to retrieve alerts from alert manager
```
simonpasquier, line 77: can't we bind to the [`monitoring-alertmanager-edit`](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/cluster-monitoring-operator/monitoring-alertmanager-edit-role.yaml) role instead?

simonpasquier (commented):
Alertmanager can be disabled by cluster admins, I think that the operator should detect this to avoid failing.

tremes:
@simonpasquier The way we will detect is that we receive an error and the operator is not degraded/failing in this case (only the particular gatherer), simply because some gatherers are expected to fail (sometimes, in various conditions).

manifests/03-clusterrole.yaml
prashbnair: Done

pkg/gatherers/clusterconfig/silenced_alerts.go
```diff
@@ -0,0 +1,46 @@
...
+		Param("inhibited", "false").
+		DoRaw(ctx)
+	if err != nil {
+		klog.Errorf("Unable to retrieve silenced alerts: %v", err)
+		return nil, []error{err}
+	}
+
+	records := []record.Record{
+		{Name: "config/silenced_alerts.json", Item: marshal.RawByte(data)},
+	}
```
tremes, line 43: If the `data` is empty this will be still recorded (with `[]`).If the data is empty, it is better not to record. So please add some simple check.

prashbnair: Done

tremes: thanks

tremes: Unfortunately this doesn't work, because the data is `[]` so `if (data != nil) && (len(data) != 0)` is always true

tremes:
Reviewed and tested. Works good! Thank you!

tremes:
Not sure what the change was, but it still looks good :)
/lgtm

PR #557, opened 2021-11-29
feat: limit number of containers per namespace

rluders:
This PR implements a way to limit the number of logs collected per container in the given namespace.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

none.

## Documentation

none.

## Unit Tests

- `pkg/gatherers/conditional/conditional_gatherer_test.go` (updated)

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-5544

### DIFF
--- a/pkg/gatherers/common/gather_logs.go
+++ b/pkg/gatherers/common/gather_logs.go
@@ -75,7 +75,7 @@ func CollectLogsFromContainers( //nolint:gocyclo
 		return nil, err
 	}
 
-	var skipped int
+	var skippedContainers int
 	var records []record.Record
 
 	for i := range pods.Items {
@@ -85,6 +85,12 @@ func CollectLogsFromContainers( //nolint:gocyclo
 		}
 		for j := range pods.Items[i].Spec.InitContainers {
 			containerNames = append(containerNames, pods.Items[i].Spec.InitContainers[j].Name)
+		}
+
+		containersLimited := containersFilter.MaxNamespaceContainers > 0 && len(records) >= containersFilter.MaxNamespaceContainers
+		if containersLimited {
+			skippedContainers += len(containerNames)
+			continue
 		}
 
 		pod := &pods.Items[i]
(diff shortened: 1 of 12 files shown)


pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -71,8 +71,9 @@ var defaultGatheringRules = []GatheringRule{
 		},
 		GatheringFunctions: GatheringFunctions{
 			GatherLogsOfNamespace: GatherLogsOfNamespaceParams{
-				Namespace: "openshift-cluster-samples-operator",
-				TailLines: 100,
+				Namespace:     "openshift-cluster-samples-operator",
+				TailLines:     100,
+				MaxContainers: 10,
```
rluders, line 76: It was really hard to figure out the right number here. So, I must say that I'm not pretty sure about it. I tried to consider the size of the log per namespace, but even so, hard to come with a number.

I believe that 10 is a reasonable number, but, maybe we need to tune it later.

pkg/gatherers/common/gather_logs.go
```diff
@@ -112,6 +115,12 @@ func CollectLogsFromContainers( //nolint:gocyclo
 				tailLines = nil
 			}
 
+			if containersFilter.MaxNamespaceContainers > 0 && len(records) >= containersFilter.MaxNamespaceContainers {
+				errs = append(errs, fmt.Errorf("skipping %s for %s. Max containers per namespace reached (max: %d)",
```
Sergey1011010, line 119: Isn't it gonna spam the errors? I think you can exit the loop here

pkg/gatherers/conditional/gathering_rule.schema.json
```diff
@@ -104,6 +104,12 @@
                             "title": "TailLines",
                             "minimum": 1,
                             "maximum": 4096
+                        },
+                        "max_containers": {
```
Sergey1011010, line 108: One of the initial purposes of the task was that there would be a hard limit on the operator side so that when we define the conditions externally, it can't collect too much.

pkg/gatherers/common/gather_logs.go
rluders: Yep, it will. But, at the same time, we want to know about the containers that we are skipping. So, another option would be to have a single error with the number of pods skipped. What do you think?

Sergey1011010: Single error with the number of skipped pods is much better

rluders: Done.

pkg/gatherers/conditional/gathering_rule.schema.json
rluders: How about now?

pkg/gatherers/common/gather_logs.go
```diff
@@ -116,8 +116,7 @@ func CollectLogsFromContainers( //nolint:gocyclo
 			}
 
 			if containersFilter.MaxNamespaceContainers > 0 && len(records) >= containersFilter.MaxNamespaceContainers {
-				errs = append(errs, fmt.Errorf("skipping %s for %s. Max containers per namespace reached (max: %d)",
-					containerName, containersFilter.Namespace, containersFilter.MaxNamespaceContainers))
+				skipped++
```
Sergey1011010, line 119: Isn't it `len(pods.Items) - containersFilter.MaxNamespaceContainers`?

rluders: It also works.

rluders: Yeh, maybe half correct?

``len(containerNames) - containersFilter.MaxNamespaceContainers``

rluders: Unless we want to get the number of pods, not the containers... which is fine as well...

rluders: Nah, nope. We need to know the number of containers, 'cause that is what we are limiting. So, counting the `pod.Items` isn't what we want.

rluders: Also, 'cause we have some regex rule that checks if the containerName matches some rules at L#94, well... it can cause some misleading.

rluders: I kind of used a similar logic to count the `containerNames` as well to break the loop when the limit reaches, so... it will avoid entering in some unnecessary for to found stuff.

pkg/gatherers/conditional/conditional_gatherer.go
tremes: Wasn't the idea to define this number within the condition?

rluders: Well.... https://github.com/openshift/insights-operator/pull/557#discussion_r758418597

natiiix (approved):
I haven't really tested it locally because it would require changing the code of a log gatherer and then triggering it on a testing cluster, which would be quite time-consuming, but the changes seem fairly straight-forward, so I have no complaints.
/lgtm

PR #563, opened 2021-12-08
feat: conditional log gathers into a single gather and PrometheusOperatorSyncFailed

rluders:
This PR implements a new conditional gathering for `PrometheusOperatorSyncFailed` as well it merges the conditional pod logs into one, making it easy to maintain.

In order to test it, you need to create a cluster with the cluster-bot using this [PR](https://github.com/openshift/cluster-monitoring-operator/pull/1514).

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/conditional/namespaces/openshift-monitoring/pods/prometheus-operator-5474fcdb86-nknz9/containers/kube-rbac-proxy/logs/last-50-lines.log`

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gatherers/conditional/gather_containers_logs_test.go` (renamed and updated)

## Privacy

## Breaking Changes

No.

## References

https://issues.redhat.com/browse/CCXDEV-6548
https://issues.redhat.com/browse/CCXDEV-6404


pkg/gatherers/conditional/gather_containers_logs.go
```diff
@@ -33,52 +34,67 @@ func (g *Gatherer) BuildGatherLogsOfUnhealthyPods(paramsInterface interface{}) (
...
+	params GatherContainersLogsParams,
+	coreClient corev1client.CoreV1Interface,
 ) ([]record.Record, []error) {
-	errs := []error{}
-	records := []record.Record{}
-
 	alertInstances, ok := g.firingAlerts[params.AlertName]
 	if !ok {
-		return nil, []error{fmt.Errorf("conditional gatherer triggered, but specified alert %q is not firing", params.AlertName)}
+		err := fmt.Errorf("condigional gather triggered, but specified alert %q is not firing", params.AlertName)
```
tremes, line 51: Typo `condigional`

pkg/gatherers/conditional/gather_containers_logs.go
```diff
@@ -33,52 +34,67 @@ func (g *Gatherer) BuildGatherLogsOfUnhealthyPods(paramsInterface interface{}) (
...
+			Namespace:     alertPodNamespace,
+			FieldSelector: fmt.Sprintf("metadata.name=%s", alertPodName),
+		}
 
 		// The container label may not be present for all alerts (e.g., KubePodNotReady).
-		containerFilter := ""
-		if alertContainer, ok := alertLabels["container"]; ok && alertContainer != "" {
-			containerFilter = fmt.Sprintf("^%s$", alertContainer)
+		if len(alertPodContainer) > 0 {
+			logContainersFilter.ContainerNameRegexFilter = fmt.Sprintf("^%s$", alertPodContainer)
```
tremes, line 86: Do we have to do this when the container name (`params.Container`) is given? This looks like we will always be searching the containers by the regex, not sure if it's necessary. 
I would also call the variable just `containerName` instead of `alertPodContainer`....just my 2 cents.

tremes: OK it's probably not a big deal, when I am looking at it again, but maybe we could think of updating the `CollectLogsFromContainers` to have some attribute/param for the container name.

rluders: I have no strong opinions about it. We could probably handle it in some other task.

docs/insights-archive-sample/conditional/namespaces/openshift-monitoring/pods/prometheus-operator-5474fcdb86-nknz9/containers/kube-rbac-proxy/logs/last-50-lines.log
```diff
@@ -0,0 +1,6 @@
+2021-12-16T14:24:33.921560406Z I1216 14:24:33.921420       1 main.go:151] Reading config file: /etc/kube-rbac-policy/config.yaml
```
rluders, line 1: Sorry about the logs, it is pretty hard to emulate the issue. That was kind of all that I could get from it.

natiiix (commented):
It looks alright to me, but I didn't have a chance to test it yet because it's quite complicated to generate many/most of the alerts. Anyways, I'd prefer if another person also took a look at this PR since it's quite a heavy change, so I'm not giving it a definitive approval just yet. However, I currently don't see any significant issue. :+1:

pkg/gatherers/conditional/common.go
```diff
@@ -25,3 +25,13 @@ func getAlertPodNamespace(labels AlertLabels) (string, error) {
 	}
 	return namespace, nil
 }
+
+func getAlertPodContainer(labels AlertLabels) (string, error) {
+	container, ok := labels["container"]
+	if !ok && len(container) > 0 {
+		newErr := fmt.Errorf("alert is missing 'container' label")
```
tremes, line 32: It would be good to tell which alert is missing that label.

pkg/gatherers/conditional/gather_containers_logs.go
```diff
@@ -33,52 +34,67 @@ func (g *Gatherer) BuildGatherLogsOfUnhealthyPods(paramsInterface interface{}) (
...
 			continue
 		}
+		var podContainer string
+		if len(params.Container) > 0 {
+			podContainer = params.Container
+		} else {
+			podContainer, err = getAlertPodContainer(alertLabels)
+			if err != nil {
+				errs = append(errs, err)
+			}
```
tremes, line 76: Hmm wouldn't be slightly better to call `continue` in case of error? Instead of checking https://github.com/openshift/insights-operator/pull/563/files#diff-d31fdac774b007a5e25edf343a316c64d1a2292ff5685f1e7c73fc4d63521841R85 again

tremes: Wouldn't you gather the logs from the whole namespace in such case?

rluders: I think that the idea is that exactly that. If we can't get the container name, go and grab it from all containers.

```// The container label may not be present for all alerts (e.g., KubePodNotReady).```

As the comment says it happens 'cause for some alerts, we don't have the container name. This implementation was done by @natiiix. Maybe he knows more details about it... but I believe that it is correct, otherwise we will have the alert being fired and not being able to collect the logs if the container name isn't present for that specific alert.

tremes: OK you're probably right. Not from all containers but probably from `FieldSelector: fmt.Sprintf("metadata.name=%s", podName),`

natiiix: Well, with the `KubePodNotReady` status, you cannot even get a container name because the pod and its containers don't really exist yet. It may be a bit of a special case, and we can't get any logs out of it anyways in the end, but that's why I added the extra check - in case there are other alerts with only information about the namespace and pod name, but without specifying the exact container that's causing the alert. I think it makes sense for alerts like that to exist.

tremes:
I reviewed couple of times. I think it works as expected. Thank you.
/lgtm

RickJWagner:
Reviewed this PR (and got context from CCXDEV-6548, CCXDEV-6404).  Looks good for PLM.

/label px-approved

tremes:
I went through it quickly because it was just a rebase. Thanks
/lgtm

PR #564, opened 2021-12-09
refactor: recucing configobserver.go cognitive complexity

rluders:
This PR reduces the cognitive complexity to `configobserver.go`.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No.

## Documentation

No.

## Unit Tests

- `pkg/config/configobserver/config_secret_test.go`

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-6266

### DIFF
--- a/pkg/config/configobserver/config_secret.go
+++ b/pkg/config/configobserver/config_secret.go
@@ -1,0 +1,142 @@
+package configobserver
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/insights-operator/pkg/config"
+)
+
+// ConfigSecret defines the configuration loaded from cluster secret
+type ConfigSecret struct {
+	config.Controller
+}
+
+// MinDuration defines the minimal report interval
+const MinDuration = 10 * time.Second
+
+// LoadSecretConfig loads the controller config with given secret data
+func LoadSecretConfig(secret *v1.Secret) (config.Controller, error) {
+	var cfg ConfigSecret
+	var err error
+
+	cfg.loadCredentials(secret.Data)
+	cfg.loadEndpoint(secret.Data)
+	cfg.loadHTTP(secret.Data)
+	cfg.loadReport(secret.Data)
+	cfg.loadOCM(secret.Data)
+
+	if intervalString, ok := secret.Data["interval"]; ok {
+		var duration time.Duration
+		duration, err = time.ParseDuration(string(intervalString))
+		if err == nil && duration < MinDuration {
+			err = fmt.Errorf("too short")
+		}
+		if err == nil {
+			cfg.Interval = duration
+		} else {
+			err = fmt.Errorf(
+				"insights secret interval must be a duration (1h, 10m) greater than or equal to ten seconds: %v",
+				err)
+			cfg.Report = false
+		}
+	}
+
+	return cfg.Controller, err
+}
+
+func (c *ConfigSecret) loadCredentials(data map[string][]byte) {
+	if username, ok := data["username"]; ok {
+		c.Username = string(username)
+	}
+	if password, ok := data["password"]; ok {
+		c.Password = string(password)
+	}
+}
+
+func (c *ConfigSecret) loadEndpoint(data map[string][]byte) {
+	if endpoint, ok := data["endpoint"]; ok {
+		c.Endpoint = string(endpoint)
+	}
+}
+
+func (c *ConfigSecret) loadHTTP(data map[string][]byte) {
+	if httpProxy, ok := data["httpProxy"]; ok {
+		c.HTTPConfig.HTTPProxy = string(httpProxy)
+	}
+	if httpsProxy, ok := data["httpsProxy"]; ok {
+		c.HTTPConfig.HTTPSProxy = string(httpsProxy)
+	}
+	if noProxy, ok := data["noProxy"]; ok {
+		c.HTTPConfig.NoProxy = string(noProxy)
+	}
+}
+
+func (c *ConfigSecret) loadReport(data map[string][]byte) {
+	if enableGlobalObfuscation, ok := data["enableGlobalObfuscation"]; ok {
+		c.EnableGlobalObfuscation = strings.EqualFold(string(enableGlobalObfuscation), "true")
+	}
+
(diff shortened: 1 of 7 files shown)


pkg/config/configobserver/configobserver.go
```diff
@@ -262,64 +104,85 @@ func (c *Controller) ConfigChanged() (configCh <-chan struct{}, closeFn func())
...
+		}
+	}
+
+	return secret, err
+}
+
+// Updates the stored tokens from the secrets in the cluster. (if present)
+func (c *Controller) updateToken(ctx context.Context) error {
+	klog.V(2).Infof("Refreshing configuration from cluster pull secret")
+	secret, err := c.fetchSecret(ctx, "pull-secret")
```
tremes, line 128: The error should be handled immediately.

pkg/config/configobserver/configobserver.go
```diff
@@ -262,64 +104,85 @@ func (c *Controller) ConfigChanged() (configCh <-chan struct{}, closeFn func())
...
+		return err
+	}
+	return nil
+}
+
+// Updates the stored configs from the secrets in the cluster. (if present)
+func (c *Controller) updateConfig(ctx context.Context) error {
+	var nextConfig config.Controller
+	klog.V(2).Infof("Refreshing configuration from cluster secret")
+	secret, err := c.fetchSecret(ctx, "support")
```
tremes, line 154: Same here

pkg/config/configobserver/config_secret.go
```diff
@@ -0,0 +1,142 @@
...
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/insights-operator/pkg/config"
+)
+
+// ConfigSecret defines the configuration loaded from cluster secret
+type ConfigSecret struct {
```
tremes, line 15: `ConfigSecret` name seems misleading to me TBH. I know we read the config from the secret (even though the secret is not required and I guess it's not present in most cases) but I think I would omit the word `secret` here.

pkg/config/configobserver/configobserver.go
```diff
@@ -262,64 +104,92 @@ func (c *Controller) ConfigChanged() (configCh <-chan struct{}, closeFn func())
...
+			nextConfig.Token = token
+			nextConfig.Report = true
+		}
+	}
+
+	c.setTokenConfig(&nextConfig)
+
+	if err != nil {
+		return err
+	}
```
tremes, line 150: This if block is not necessary now right?

natiiix (commented):
I was not very excited about how complicated the old code was, and I guess these changes make it a lot more readable. Plus, there are now more tests, which is definitely a good thing. I don't see anything wrong with the code other than the details pointer out by Tomas, which have already been addressed. Since he was the first to review it, I will let him decide if it's ready to be merged, but as far as I am concerned, it looks fine. :+1:

pkg/config/configobserver/config_test.go
```diff
@@ -0,0 +1,234 @@
...
+	"reflect"
+	"testing"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/openshift/insights-operator/pkg/config"
+)
+
+func TestConfigSecret_loadCredentials(t *testing.T) {
```
tremes, line 13: It would be good to remove the `Secret` also from the test names

pkg/config/configobserver/config_test.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func TestConfigSecret_loadEndpoint(t *testing.T) {
+	tests := []struct {
+		name string
+		data map[string][]byte
+		want *Config
+	}{
+		{
+			name: "Load HTTP Config",
```
tremes, line 50: Shouldn't this be `Load endpoint`? :)

pkg/config/configobserver/configobserver.go
```diff
@@ -262,64 +104,89 @@ func (c *Controller) ConfigChanged() (configCh <-chan struct{}, closeFn func())
...
+	klog.V(2).Infof("Refreshing configuration from cluster pull secret")
+	secret, err := c.fetchSecret(ctx, "pull-secret")
+	if err != nil {
+		return err
+	}
+
+	var nextConfig config.Controller
+	if secret != nil {
+		var token string
+		token, err = tokenFromSecret(secret)
```
tremes, line 136: nit: It would be simpler to go with `token, err := tokenFromSecret(secret)` and save one line :)

pkg/config/configobserver/configobserver.go
```diff
@@ -262,64 +104,89 @@ func (c *Controller) ConfigChanged() (configCh <-chan struct{}, closeFn func())
...
+		if err != nil {
+			return err
+		}
+		if len(token) > 0 {
+			nextConfig.Token = token
+			nextConfig.Report = true
```
tremes, line 142: Hmm I am actually not sure about this line right now. Does it actually fix the original logic? It looks like it doesn't break anything right.... :) I am just wondering why [this](https://github.com/openshift/insights-operator/blob/master/pkg/config/configobserver/configobserver.go#L107) was set regardless the token.

pkg/config/configobserver/configobserver.go
rluders: In this case, I'll be shadowing the `err` at L#130. I can give it a try, but almost sure that the lining will point the shadowing.

pkg/config/configobserver/configobserver.go
rluders: Well, I didn't notice any issues on my tests, but... maybe would be a good idea to keep the same logic.

pkg/config/configobserver/configobserver.go
tremes: Ahh ok np then...

pkg/config/configobserver/configobserver.go
tremes: Well....it was (and maybe still is :-)) really confusing. I guess that originally it was overwritten at https://github.com/openshift/insights-operator/blob/master/pkg/config/configobserver/configobserver.go#L190 

It looks like it's not overwritten after your update so it's hopefully OK.

tremes:
I didn't test this one but reviewed several times. Thanks you!
/lgtm

PR #565, opened 2021-12-14
info alert when the SCA is not available

tremes:
This adds a new info alert when the SCA is not available/enabled for the cluster

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Unit Tests

No unit test

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/manifests/08-prometheus_rule.yaml
+++ b/manifests/08-prometheus_rule.yaml
@@ -22,10 +22,10 @@ spec:
         severity: info
     - alert: SCANotAvailable
       annotations:
-        description: 'Simple content access is not enabled. See https://access.redhat.com/articles/simple-content-access to enable it.'
+        description: 'Simple content access (SCA) is not enabled. Once enabled, Insights Operator can automatically import the SCA certificates from Red Hat OpenShift Cluster Manager making it easier to use the content provided by your Red Hat subscriptions when creating container images. See https://docs.openshift.com/container-platform/latest/cicd/builds/running-entitled-builds.html for more information.'
         summary: Simple content access certificates are not available.
       expr: |
-       cluster_operator_conditions{name="insights", condition="SCANotAvailable", reason="NotFound"} == 1
+       max_over_time(cluster_operator_conditions{name="insights", condition="SCANotAvailable", reason="NotFound"}[5m]) == 1
       for: 5m
       labels:
         severity: info
--- a/manifests/08-prometheus_rule.yaml
+++ b/manifests/08-prometheus_rule.yaml
@@ -20,4 +20,13 @@ spec:
       for: 5m
       labels:
         severity: info
+    - alert: SCANotAvailable
+      annotations:
+        description: 'Simple content access is not enabled. See https://access.redhat.com/articles/simple-content-access to enable it.'
+        summary: Simple content access certificates are not available.
+      expr: |
+       cluster_operator_conditions{name="insights", condition="SCANotAvailable", reason="NotFound"} == 1
+      for: 5m
+      labels:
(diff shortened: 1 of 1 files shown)


tremes:
@siamaksade Can you please take a look?

manifests/08-prometheus_rule.yaml
```diff
@@ -20,4 +20,13 @@ spec:
       for: 5m
       labels:
         severity: info
+    - alert: SCANotAvailable
+      annotations:
+        description: 'Simple content access is not enabled. See https://access.redhat.com/articles/simple-content-access to enable it.'
```
sferich888, line 25: I don't see how linking to this is helpful. This page doesn't talk about how to use SCA with OCP at all.

sferich888: If I follow the controller logic correctly insights sets this up and establishes this thus this is simply an indication that the operator can't interact with OCM and subscription manager correctly: 

https://github.com/openshift/insights-operator/blob/master/pkg/ocm/ocm.go#L93-L136

sferich888 (changes requested):
I think I need to understand what is going on better in this before approving it.

manifests/08-prometheus_rule.yaml
tremes: It's not about how to use it, but about how to enable it. Yes but there's an agreement with the OCM API that when we get HTTP 404 then it basically means that the user didn't enable the SCA subscription and thus the alert (it's only for `Reason=NotFound`).

tremes: Anyway, I definetely welcome suggestions for a better description :-)

manifests/08-prometheus_rule.yaml
```diff
@@ -20,4 +20,13 @@ spec:
       for: 5m
       labels:
         severity: info
+    - alert: SCANotAvailable
```
simonpasquier, line 23: I would be in favor of a more meaningful name. Without prior knowledge of the insights operator, SCA is cryptic to me.

manifests/08-prometheus_rule.yaml
```diff
@@ -20,4 +20,13 @@ spec:
...
+      annotations:
+        description: 'Simple content access is not enabled. See https://access.redhat.com/articles/simple-content-access to enable it.'
+        summary: Simple content access certificates are not available.
+      expr: |
+       cluster_operator_conditions{name="insights", condition="SCANotAvailable", reason="NotFound"} == 1
```
simonpasquier, line 28: To avoid alert resets with failed scrapes, it's recommended to use an *_over_time() function (max_over_time() in this case):

```suggestion
       max_over_time(cluster_operator_conditions{name="insights", condition="SCANotAvailable", reason="NotFound"}[5m]) == 1
```

manifests/08-prometheus_rule.yaml
tremes: HI @simonpasquier thanks for reviewing this one. OK. Do you have any suggestion please? e.g `SimpleContentAccessNotAvailable`?

tremes: OK let me update it..thank you

natiiix (approved):
Tested by running the image from this PR on a testing cluster, setting the `scaEndpoint` to a URL that returns HTTP 404 and waiting for a couple of minutes. Then, once the invalid `scaEndpoint` was removed and IO restarted, the alert was no longer being generated after the old one expired.
/lgtm

PR #567, opened 2021-12-16
Remove unnecessary division into important and failable gatherers

tremes:
This is removing the unnecessary division into important and failable gatherers. Why? Because every gatherer can fail and in case of failure of an important gatherer, it is not worth repeating the gathering sooner IMO. This was the only significant difference when having this division. 

This requires thorough testing

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

no updates

## Documentation

- `path/to/documentation.md`

## Unit Tests

updated

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/controller/periodic/periodic.go
+++ b/pkg/controller/periodic/periodic.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/openshift/insights-operator/pkg/anonymization"
 	"github.com/openshift/insights-operator/pkg/config/configobserver"
-	"github.com/openshift/insights-operator/pkg/controller/status"
 	"github.com/openshift/insights-operator/pkg/controllerstatus"
 	"github.com/openshift/insights-operator/pkg/gather"
 	"github.com/openshift/insights-operator/pkg/gatherers"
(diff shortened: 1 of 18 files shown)


pkg/controller/periodic/periodic.go
```diff
@@ -129,33 +116,30 @@ func (c *Controller) Gather() {
...
-				})
-			return false, nil
-		})
+		utilruntime.HandleError(fmt.Errorf("%v failed after %s with: %v", name, time.Since(start).Truncate(time.Millisecond), err))
+		/* 			c.statuses[name].UpdateStatus(
+		controllerstatus.Summary{
+			Operation: controllerstatus.GatheringReport,
+			Reason:    "PeriodicGatherFailed",
+			Message:   fmt.Sprintf("Source %s could not be retrieved: %v", name, err),
+		}) */
```
tremes, line 142: I am not sure if I want to remove this as well....let's see while testing.

natiiix (approved):
I supported the idea of not using the confusing concept of important and failable gatherers for quite a while now, so I see no issue with it in this regard. I tried running this version of IO on a testing cluster and it appears to work as it should, didn't crash and generated a valid archive.
/lgtm

RickJWagner:
PLM reviews, approves.

/label px-approved

PR #571, opened 2022-01-11
Cluster transfer OCM controller

tremes:
This is adding a new controller for periodically checking OCM API for the cluster transfer.

Most of the time there will be no cluster transfer at all (right now we check only `accepted` cluster transfers). This means that the API will return HTTP 204 response and we do nothing. 

If there is one unique accepted cluster transfer then read the new secret data and check if the existing `pull-secret` requires an update or not (in other words we want to update it only once per accepted cluster transfer). If the `pull-secret` requires an update then update it. 

If there is by any coincidence more than 1 accepted cluster transfer then log the message, update the controller status and do nothing.

The documentation how to work with the new cluster transfer OCM API is https://gitlab.cee.redhat.com/service/uhc-account-manager/-/blob/master/docs/cluster_transfer_api.md

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Unit Tests

- `pkg/ocm/clustertransfer/cluster_transfer_test.go`

## Privacy

## Breaking Changes

No

## References


pkg/ocm/clustertransfer/cluster_transfer_test.go
```diff
@@ -0,0 +1,72 @@
...
+		name          string
+		defaultData   []byte
+		dataToUpdate  []byte
+		updatedResult []byte
+	}{
+		{
+			name:          "Updating cloud.openshift.com auth attributes and quay.io email",
+			defaultData:   []byte(`{"auths":{"cloud.openshift.com":{"auth":"eyJSb2xlIjoiwMjMsImlhdCI6MTY0MTg5MDAyM30","email":"test@test.com"},"quay.io":{"auth":"l3YM6DX9frFXhswvfuBv__dhrpDACa4F0E","email":"test-quay@test.com"}}}`),
+			dataToUpdate:  []byte(`{"auths":{"cloud.openshift.com":{"auth":"eyJSb2xlIjoiwMjMsImlhdCI6MTY0MTg5MDAyM30==","email":"updated@updated.com"},"quay.io":{"email":"test-quay@updated.com"}}}`),
+			updatedResult: []byte(`{"auths":{"cloud.openshift.com":{"auth":"eyJSb2xlIjoiwMjMsImlhdCI6MTY0MTg5MDAyM30==","email":"updated@updated.com"},"quay.io":{"auth":"l3YM6DX9frFXhswvfuBv__dhrpDACa4F0E","email":"test-quay@updated.com"}}}`),
```
tremes, line 28: These should probably go to files.

pkg/ocm/clustertransfer/cluster_transfer.go
```diff
@@ -0,0 +1,136 @@
...
+		return err
+	}
+
+	updatedData, err := jsonpatch.MergePatch(pullSecret.Data[v1.DockerConfigJsonKey], newData)
+	if err != nil {
+		return err
+	}
+
+	pullSecret.Data[v1.DockerConfigJsonKey] = updatedData
+	_, err = c.coreClient.Secrets("openshift-config").Update(c.ctx, pullSecret, metav1.UpdateOptions{})
```
tremes, line 97: Hmm we need to be a bit smarter here - we should check if there's really some diff and update only in this case.

natiiix (commented):
I have looked through the code, but I had some trouble getting a testing cluster running at the end of last week, so I was not able to perform the experiment yet, unfortunately. I don't have any issues with the code, but I can imagine _someone_ might complain about adding `TODO`s since there's been an effort to get rid of them not so long ago.

pkg/ocm/clustertransfer/cluster_transfer.go
```diff
@@ -0,0 +1,276 @@
...
+	existingData := c.pullSecret.Data[v1.DockerConfigJsonKey]
+
+	updatedData, err := jsonpatch.MergePatch(existingData, newData)
+	if err != nil {
+		return err
+	}
+
+	c.pullSecret.Data[v1.DockerConfigJsonKey] = updatedData
+	_, err = c.coreClient.Secrets("openshift-config").Update(c.ctx, c.pullSecret, metav1.UpdateOptions{})
+	if err != nil {
```
rluders, line 175: Can't you just return err?

pkg/ocm/clustertransfer/cluster_transfer.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	var data []byte
+	err := wait.ExponentialBackoff(bo, func() (bool, error) {
+		resp, err := c.client.Get(endpoint)
+		if err != nil {
+			return true, err
+		}
+		defer resp.Body.Close()
+		// TODO this will be moved to the insightsclient
```
rluders, line 201: when PoC gets approved, convert it to an issue.

pkg/ocm/clustertransfer/cluster_transfer.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	// there's no cluster transfer for the cluster - HTTP 204
+	if len(data) == 0 {
+		klog.Info("No cluster transfer exists")
+		c.updateStatus(true, "no cluster transfer content", "NoData", nil)
+		return
+	}
+
+	// TODO somewhere here we need to check if the ClusterTransfer is Accepted
```
rluders, line 110: when PoC gets approved, convert it to an issue.

config/pod.yaml
```diff
@@ -14,5 +14,8 @@ pull_report:
 ocm:
   scaEndpoint: https://api.openshift.com/api/accounts_mgmt/v1/certificates
   scaInterval: "8h"
+  # TODO
```
rluders, line 17: when PoC gets approved, convert it to an issue.

pkg/ocm/clustertransfer/cluster_transfer.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// New creates new instance of controllers
+func New(ctx context.Context, coreClient corev1client.CoreV1Interface, configurator Configurator) *Controller {
+	return &Controller{
+		Simple:       controllerstatus.Simple{Name: "clusterTransferController"},
+		coreClient:   coreClient,
+		ctx:          ctx,
+		configurator: configurator,
+		// TODO change to insightsclient
```
rluders, line 58: when PoC gets approved, convert it to an issue.

rluders (changes requested):
I had some issues testing it, and gonna give it another try later. But reading the code, in general, it looks good. Just a few comments, mostly 'cause I'm the _someone_ that complains about TODOs. :rofl:

pkg/ocm/clustertransfer/cluster_transfer.go
tremes: Yes thanks. My plan is not to merge it. I will keep it here once there is a new OCM API implemented.

pkg/ocm/clustertransfer/cluster_transfer.go
tremes: What do you mean please? I am returning the error if it is not `nil`

rluders: it is just an idea, 'cause I guess that if there is no error, `err` will be `nil`... right?

tremes: Ahh OK now I understand. Yes it should be possible. I think this is more explicit but I don't have a strong opinion

xJustin:
@tremes is the `clusterTransferInterval: "24h"` user configurable? Also, will this feature be 4.11+ only?

tremes:
@xJustin Answers are yes and yes. Regarding the first one, there is discussion how a user could trigger or force the refresh.

pkg/controller/status/controller.go
```diff
@@ -166,7 +166,7 @@ func (c *Controller) merge(clusterOperator *configv1.ClusterOperator) *configv1.
 }
 
 // calculate the current controller status based on its given sources
-func (c *Controller) currentControllerStatus() (allReady bool, lastTransition time.Time) {
+func (c *Controller) currentControllerStatus() (allReady bool, lastTransition time.Time) { //nolint: gocyclo
```
rluders, line 169: Maybe it is not the goal here... but since we have an cyclomatic complexities alert being ignored here would be a good idea to handle every `if/if-else` into separated functions, and to simplify it with an `switch-case`. What do you think?

pkg/insights/insightsclient/requests.go
```diff
@@ -245,3 +246,49 @@ func (c *Client) RecvGatheringRules(ctx context.Context, endpoint string) ([]byt
...
+		return nil, ocmErrorMessage(resp)
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			klog.Warningf("Failed to close response body: %v", err)
+		}
+	}()
+
+	return io.ReadAll(resp.Body)
+
```
rluders, line 293: remove extra line

pkg/ocm/clustertransfer/cluster_transfer.go
```diff
@@ -0,0 +1,300 @@
...
+		Jitter:   0,
+		Steps:    status.OCMAPIFailureCountThreshold,
+		Cap:      c.configurator.Config().OCMConfig.ClusterTransferInterval,
+	}
+
+	var data []byte
+	err := wait.ExponentialBackoff(bo, func() (bool, error) {
+		var err error
+		data, err = c.client.RecvClusterTransfer(endpoint)
+		if err != nil {
```
rluders, line 218: if you replace if with:
```go
if err == nil {
    return true, nil
}
/// the rest of code here
```
or something similar you can reduce the code indentation and maybe help to reduce cognitive complexity?

pkg/ocm/clustertransfer/cluster_transfer.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	if httpErr != nil {
+		newSummary.Operation.HTTPStatusCode = httpErr.StatusCode
+	}
+	c.UpdateStatus(newSummary)
+}
+
+// getPullSecret gets pull-secret as *v1.Secret
+func (c *Controller) getPullSecret() (*v1.Secret, error) {
+	ps, err := c.coreClient.Secrets("openshift-config").Get(c.ctx, "pull-secret", metav1.GetOptions{})
```
rluders, line 258: can't you just return it directly?

rluders (changes requested):

pkg/controller/status/controller.go
tremes: Yes, but not as a part of this task/PR.

pkg/ocm/clustertransfer/cluster_transfer.go
tremes: Yes. Good point

pkg/ocm/clustertransfer/cluster_transfer.go
tremes: Yes. It seems so

rluders:
I didn't test it locally, but I checked the code several times. It looks good IMO, unit test coverage is good, the code makes sense, and also it was already approved by QE, so... 

/lgtm

PR #574, opened 2022-01-17
Remove "InsightsOperatorPullingSCA" TP feature check

tremes:
This is removing the TechPreview feature set check. IOW it means that the SCA certs will be pulled down automatically. There are also some minor enhancements in the related log messages.

This is not possible to test with cluster-bot cluster so you have to create your own cluster in AWS.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

This requires OCP doc update

## Unit Tests

No test update

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"time"
 
-	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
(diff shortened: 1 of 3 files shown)


rluders (approved):
I can't see any issue.
/lgtm

natiiix (approved):
Tested on a provided testing cluster with an IO image built from this PR. It wasn't possible to test the alert itself due to the necessity to disable CVO to modify the IO image, but the alert part was tested before on a different cluster as a part of a different PR. When setting the `scaEndpoint` to a URL that returns 404, the operator status changed to `SCANotAvailable`, and then back to `AsExpected` after reseting the secret entry to the default endpoint.
/lgtm

PR #576, opened 2022-01-24
feat: pod_definition conditional gather

rluders:
This PR implements a new conditional gather to collect pod definitions when a given alert is configured (currently: `KubePodNotReady`).

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/conditional/namespaces/openshift-monitoring/alertmanager-main-0.json`

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gatherers/conditional/gather_pod_definition_test.go`

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-6558
https://bugzilla.redhat.com/show_bug.cgi?id=2024797


rluders:
The linting is complaining about some that aren't related to this PR. I'm overriding, and the other issues will be handled in a separate PR.

/override ci/prow/lint

pkg/gatherers/conditional/gather_pod_definition.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/openshift/insights-operator/pkg/gatherers"
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// BuildGatherPodDefinition collects pod definition from pods that are firing one of the configured alerts.
+//
+// * Location in archive: conditional/namespaces/<namespace>/pods/<pod>.json
+// * Id in config: pod_definition
+// * Since versions:
+//   * 4.10+
```
tremes, line 21: nit - this will not make it into 4.10. So probably `4.11` pls

pkg/gatherers/conditional/gather_pod_definition.go
```diff
@@ -0,0 +1,91 @@
...
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/insights-operator/pkg/gatherers"
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// BuildGatherPodDefinition collects pod definition from pods that are firing one of the configured alerts.
+//
+// * Location in archive: conditional/namespaces/<namespace>/pods/<pod>.json
+// * Id in config: pod_definition
```
tremes, line 19: Hmm do we use these ids for the conditional gathering at all?

pkg/gatherers/conditional/gather_pod_definition_test.go
```diff
@@ -0,0 +1,86 @@
...
+			g := &Gatherer{firingAlerts: tt.fields.firingAlerts}
+
+			coreClient := tt.args.coreClient(tt.args.ctx)
+			got, gotErr := g.gatherPodDefinition(tt.args.ctx, tt.args.params, coreClient)
+
+			assert.Len(t, got, tt.wantLen)
+			if tt.wantErr {
+				assert.Len(t, gotErr, 1)
+			} else {
+				assert.Len(t, gotErr, 0)
```
tremes, line 82: Wouldn't be simpler to assert something like `assert.NoError(t, gotErr)` ?

pkg/gatherers/conditional/gather_pod_definition.go
rluders: Not that I'm aware. Maybe with conditional service running? We do have it in others CG if I'm not wrong...

pkg/gatherers/conditional/gather_pod_definition_test.go
tremes: Ohh OK it's a slice of errors....I thought it's just `error` so it's probably OK.

pkg/gatherers/conditional/gather_pod_definition.go
tremes: It's used only in the `GatheringFunctionName` (e.g https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/conditional/gathering_functions.go#L19), but I am not sure if it's taken into account in the https://github.com/openshift/insights-operator/blob/master/pkg/gather/gather.go#L203. It seems it is. It would be good to verify and then unify across all the conditional gatherer. And yes it's a good question how it will work with the new service

pkg/gatherers/conditional/gather_pod_definition.go
```diff
@@ -0,0 +1,91 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/insights-operator/pkg/gatherers"
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// BuildGatherPodDefinition collects pod definition from pods that are firing one of the configured alerts.
+//
+// * Location in archive: conditional/namespaces/<namespace>/pods/<pod>.json
```
tremes, line 18: Hmm so we already have e.g `conditional/namespaces/openshift-monitoring/pods/alertmanager-main-0/` dir in our sample archive. Wouldn't be better to move this json into this directory? IOW include directory with the pod name here.

rluders:
The linting is complaining about some that aren't related to this PR. I'm overriding, and the other issues will be handled in a separate PR.

/override ci/prow/lint

tremes:
I reviewed and verified that it works as expected. Thank you!

docs/gathered-data.md
```diff
@@ -660,6 +660,15 @@ Docs for relevant types: https://pkg.go.dev/github.com/openshift/api/operatorcon
   * 4.8+
 
 
+## PodDefinition
+
+collects pod definition from pods that are firing one of the configured alerts.
```
sferich888, line 665: What are the 'configured alerts'? How do I look them up / where do I find the 'configuration'?

sferich888 (commented):
/label px-approved 
Consider updating this to provide more details in the relevant docs. Its not clear what causes this to fire (get collected) because its 'configurable' and/or where that 'configuration' can be seen, reviewed and edited.

docs/gathered-data.md
rluders: The configuration alerts are currently hardcoded in the IO. We are working to get it from a service that we designed [here](https://github.com/RedHatInsights/insights-operator-gathering-conditions-service) and the conditions are [here](https://github.com/RedHatInsights/insights-operator-gathering-conditions).

tremes: I think the point here is to extend/update the documentation. I think it would be good to address it in a new task (to document new config, conditions source etc. once the new service is in prod)

PR #578, opened 2022-02-07
fix: defer in loop

rluders:
This PR removes the use of the `defer` function directly inside the loops, by wrapping it into some functions. It avoids some possible resource leaks. The `defer` functions inside the test files are not causing any issue, in this case, I'm just ignoring them to make the linter happy. And also, this PR fixes some small linting issues related to `fmt.Error`.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No

## Documentation

No

## Unit Tests

No

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-7149

### DIFF
--- a/pkg/authorizer/clusterauthorizer/clusterauthorizer_test.go
+++ b/pkg/authorizer/clusterauthorizer/clusterauthorizer_test.go
@@ -8,8 +8,9 @@ import (
 	"os"
 	"testing"
 
-	"github.com/openshift/insights-operator/pkg/config"
 	"golang.org/x/net/http/httpproxy"
+
+	"github.com/openshift/insights-operator/pkg/config"
 )
 
 // nonCachedProxyFromEnvironment creates Proxier if Proxy is set. It uses always fresh Env
@@ -84,9 +85,9 @@ func Test_Proxy(tt *testing.T) {
 	for _, tcase := range testCases {
 		tc := tcase
 		tt.Run(tc.Name, func(t *testing.T) {
-			// do not use parallel here
 			for k, v := range tc.EnvValues {
-				defer SafeRestoreEnv(k)()
+				// do not use parallel here
+				defer SafeRestoreEnv(k)() // nolint: gocritic
 				// nil will indicate the need to unset Env
 				if v != nil {
 					vv := v.(string)
--- a/pkg/controller/periodic/periodic.go
+++ b/pkg/controller/periodic/periodic.go
@@ -116,30 +116,32 @@ func (c *Controller) Gather() {
 	allFunctionReports := make(map[string]gather.GathererFunctionReport)
 
 	for _, gatherer := range gatherersToProcess {
-		name := gatherer.GetName()
-		start := time.Now()
+		func() {
+			name := gatherer.GetName()
+			start := time.Now()
 
-		ctx, cancel := context.WithTimeout(context.Background(), c.configurator.Config().Interval/2)
-		defer cancel()
+			ctx, cancel := context.WithTimeout(context.Background(), c.configurator.Config().Interval/2)
+			defer cancel()
 
-		klog.V(4).Infof("Running %s gatherer", gatherer.GetName())
-		functionReports, err := gather.CollectAndRecordGatherer(ctx, gatherer, c.recorder, c.configurator)
-		for i := range functionReports {
-			allFunctionReports[functionReports[i].FuncName] = functionReports[i]
-		}
-		if err == nil {
-			klog.V(3).Infof("Periodic gather %s completed in %s", name, time.Since(start).Truncate(time.Millisecond))
-			c.statuses[name].UpdateStatus(controllerstatus.Summary{Healthy: true})
-			continue
-		}
+			klog.V(4).Infof("Running %s gatherer", gatherer.GetName())
+			functionReports, err := gather.CollectAndRecordGatherer(ctx, gatherer, c.recorder, c.configurator)
+			for i := range functionReports {
+				allFunctionReports[functionReports[i].FuncName] = functionReports[i]
+			}
+			if err == nil {
+				klog.V(3).Infof("Periodic gather %s completed in %s", name, time.Since(start).Truncate(time.Millisecond))
+				c.statuses[name].UpdateStatus(controllerstatus.Summary{Healthy: true})
+				return
+			}
 
-		utilruntime.HandleError(fmt.Errorf("%v failed after %s with: %v", name, time.Since(start).Truncate(time.Millisecond), err))
-		c.statuses[name].UpdateStatus(
-			controllerstatus.Summary{
-				Operation: controllerstatus.GatheringReport,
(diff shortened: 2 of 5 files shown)


pkg/controller/periodic/periodic.go
```diff
@@ -116,30 +116,32 @@ func (c *Controller) Gather() {
...
-		name := gatherer.GetName()
-		start := time.Now()
+		func() {
+			name := gatherer.GetName()
+			start := time.Now()
 
-		ctx, cancel := context.WithTimeout(context.Background(), c.configurator.Config().Interval/2)
-		defer cancel()
+			ctx, cancel := context.WithTimeout(context.Background(), c.configurator.Config().Interval/2)
+			defer cancel()
```
tremes, line 124: I am wondering if it wouldn't be better to share one context for all the gatherers.

rluders: I see... That is actually a good point. Let me experiment.

rluders: Oh, no we can't. 'Cause the timeout.

tremes: The timeout is 1h by default. I think all the gatherers should be able to finish within this time.

rluders: Yeh, I know. But it also can be configured, right? Also, would be worth taking the risk?

rluders: BTW, these changes will directly require changes on tests for periodic. 'Cause, it fails.

tremes: Sure it can be configured. The risk is always there. I mean if the gathering interval is too short then it will be very likely canceled. I am just saying that I think a 1h time-out for all the gatherers should be enough. Maybe we should keep the discussion in some other task/PR.

rluders: Yep. I think that we can handle it in some other task, 'cause I'm not sure about it yet.

pkg/gatherers/conditional/validation.go
```diff
@@ -39,7 +39,7 @@ func validateGatheringRules(gatheringRules []GatheringRule) []error {
 	if !result.Valid() {
 		var errs []error
 		for _, err := range result.Errors() {
-			errs = append(errs, fmt.Errorf(err.String()))
+			errs = append(errs, fmt.Errorf("%s", err.String()))
```
natiiix, line 42: I presume this is done to copy the error because the `err` reference changes with each iteration. If that's the case, wouldn't it be better to use a similar approach to what we're currently doing with various K8s resource structs in the gathering code. Something like this:
```go
resultErrors := result.Errors()
for i := range resultErrors {
    errs = append(errs, resultErrors[i])
}
```
If the reason for this conversion from error to string and back to error is done for a different reason, then it's perfectly fine, but it just felt like unnecessary type gymnastics to me.

pkg/gather/gather.go
```diff
@@ -100,7 +100,7 @@ func CollectAndRecordGatherer(
 				gathererName, result.FunctionName, err,
 			)
 
-			errs = append(errs, fmt.Errorf(errStr))
+			errs = append(errs, fmt.Errorf("%s", errStr))
```
natiiix, line 103: Couldn't we just use `fmt.Errorf` instead of `fmt.Sprintf` in the first place? I don't see the string being used for anything other than to create the error.

natiiix (commented):
Just two small things in regard to converting `string`s into `error`s.
Otherwise, I don't see anything wrong with the PR.

pkg/gatherers/conditional/validation.go
rluders: Well... If the `resultErrors[i]` was actually an `error` that would be indeed possible, but it is more the case, the `resultErrors[i]` will return a `ResultError`.

pkg/gather/gather.go
rluders: Yep. Good point. Thanks.

pkg/gatherers/conditional/validation.go
natiiix: @rluders Oh, my bad, sorry about that. I didn't check the types, so I thought it was just regular go `error`s. In that case, we can't really do much about it. I believe there's also a way to directly convert a `string` into `error` somehow without having to do this formatting (should be slightly faster and a bit cleaner imo). Using `errors.New(string)` is one possibility that I have seen recommended, but if you don't feel like changing this, it's fine by me. I can approve the PR either way.

rluders: I prefer to keep it as `fmt.Errorf` 'cause we are using it in several other places. I considered using the `errors.New()` but... TBH, didn't see any reasonable benefit. Lint is happy with both.

natiiix: Okay, sure, it's fine by me.

natiiix (approved):
There is no significant chance in terms of functionality, just a refactor, so I have not tested this locally. I see not further issues with the code.
/lgtm

pkg/gatherers/clusterconfig/install_plans_test.go
```diff
@@ -58,45 +58,47 @@ func Test_InstallPlans_Gather(t *testing.T) {
...
-				installplancontent, err := io.ReadAll(f)
-				if err != nil {
-					t.Fatal("error reading test data file", err)
-				}
+				func() {
+					f, err := os.Open(file)
+					if err != nil {
+						t.Fatal("test failed to read installplan data", err)
+					}
+					defer f.Close()
```
sferich888, line 66: This defer is still in a 'loop' (which I thought this PR was removing)? Or are we only removing it in 'some places'?

sferich888 (changes requested):
I don't understand the PR commit message/comments; and how they accurately describe what this PR is doing.

pkg/authorizer/clusterauthorizer/clusterauthorizer_test.go
```diff
@@ -84,9 +85,9 @@ func Test_Proxy(tt *testing.T) {
 	for _, tcase := range testCases {
 		tc := tcase
 		tt.Run(tc.Name, func(t *testing.T) {
-			// do not use parallel here
 			for k, v := range tc.EnvValues {
-				defer SafeRestoreEnv(k)()
+				// do not use parallel here
+				defer SafeRestoreEnv(k)() // nolint: gocritic
```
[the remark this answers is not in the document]
rluders: Some places. 'Cause, not all places make sense.

sferich888:
/label px-approved

I don't like the commit message/description as it doesn't describe what is being fixed; this will likely lead to debug-ability issues should someone need to review our commit history, related to this code. Something like 'function wrapping key defer statements to avoid resource leaks' is probably more appropriate, to describe the things being fixed.

PR #583, opened 2022-02-17
implemented fetching rules from a remote server for conditional gathering

Sergey1011010:
## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No changes

## Documentation

- `README.md`
- `docs/conditional-gatherer/README.md`

## Unit Tests

- `pkg/gatherers/conditional/parsing_test.go`

## Privacy

No new data was collected

## Changelog

No

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-7163


pkg/config/config.go
```diff
@@ -11,11 +11,12 @@ import (
...
-	Report      bool   `json:"report"`
-	StoragePath string `json:"storagePath"`
-	Interval    string `json:"interval"`
-	Endpoint    string `json:"endpoint"`
-	PullReport  struct {
+	Report                      bool   `json:"report"`
+	StoragePath                 string `json:"storagePath"`
+	Interval                    string `json:"interval"`
+	Endpoint                    string `json:"endpoint"`
+	ConditionalGathererEndpoint string `json:"conditionalGathererEndpoint"`
```
tremes, line 18: Just suggestion. What about `conditionsEndpoint`?

Sergey1011010: the idea was that we may define more in the json later (it's likely to not happen tho) and term `conditions` could be confused with operator conditions

config/local.yaml
```diff
@@ -5,6 +5,7 @@ leaderElection:
 interval: "5m"
 storagePath: /tmp/insights-operator
 endpoint: http://[::1]:8081
+conditionalGathererEndpoint: https://console.redhat.com/api/gathering/gathering_rules
```
tremes, line 8: I am not sure we need it here, but what about [here](https://github.com/openshift/insights-operator/blob/master/pkg/cmd/start/start.go#L27)? Did you test it in a container please?

pkg/controller/const.go
```diff
@@ -1,7 +1,8 @@
 package controller
 
 const (
-	metricCAFile     = "/var/run/configmaps/service-ca-bundle/service-ca.crt"
-	metricHost       = "https://prometheus-k8s.openshift-monitoring.svc:9091"
-	alertManagerHost = "https://alertmanager-main.openshift-monitoring.svc:9094"
+	metricCAFile                       = "/var/run/configmaps/service-ca-bundle/service-ca.crt"
+	metricHost                         = "https://prometheus-k8s.openshift-monitoring.svc:9091"
+	alertManagerHost                   = "https://alertmanager-main.openshift-monitoring.svc:9094"
+	insecurePrometheusTokenEnvVariable = "INSECURE_PROMETHEUS_TOKEN"
```
tremes, line 7: This is only for testing right?

Sergey1011010: Exactly, for convenient local development

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -39,210 +40,29 @@ var gatheringFunctionBuilders = map[GatheringFunctionName]GathererFunctionBuilde
...
-		GatheringFunctions: GatheringFunctions{
-			GatherContainersLogs: GatherContainersLogsParams{
-				AlertName: "ThanosRuleQueueIsDroppingAlerts",
-				Container: "thanos-ruler",
-				TailLines: 50,
-			},
-		},
-	},
-}
-
```
tremes, line 231: Some of these conditions are not currently defined by the new service. Can you please create a task to update it?

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -275,6 +97,14 @@ func (g *Gatherer) GetName() string {
 // GetGatheringFunctions returns gathering functions that should be run considering the conditions
 // + the gathering function producing metadata for the conditional gatherer
 func (g *Gatherer) GetGatheringFunctions(ctx context.Context) (map[string]gatherers.GatheringClosure, error) {
+	newGatheringRules, err := g.fetchGatheringRulesFromServer(ctx)
+	if err != nil {
+		klog.Errorf("unable to fetch gathering rules from the server: %v", err)
+		klog.Info("trying to use cached gathering functions")
```
tremes, line 103: Where are the gathering functions cached please? It seems that `g.gatheringRules` is an empty slice in case of an error and then the `gatheringFunctions` is an empty map or am I wrong?

pkg/gatherers/conditional/parsing.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"k8s.io/klog/v2"
+)
+
+func parseGatheringRules(jsonData string) ([]GatheringRule, error) {
+	var unmarshalledRules struct {
+		Version string          `json:"version"`
```
tremes, line 12: I think it would be good to at least log the version of the conditions IO is running with. The metadata is probably the best place where it would be useful, but let's create a new task for that. WDYT?

Sergey1011010: It definitely will be helpful. I'll create a task for that.

Sergey1011010: https://issues.redhat.com/browse/CCXDEV-7488

pkg/insights/insightsclient/requests.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// dynamically set the proxy environment and authentication
+	c.client.Transport = clientTransport(c.authorizer)
+
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, ocmErrorMessage(resp)
```
tremes, line 233: The `ocmErrorMessage` talks about `OCM API` which can be misleading here.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -39,210 +40,29 @@ var gatheringFunctionBuilders = map[GatheringFunctionName]GathererFunctionBuilde
...
+	firingAlerts                map[string][]AlertLabels
+	gatheringRules              []GatheringRule
+	clusterVersion              string
+	configurator                configobserver.Configurator
+	gatheringRulesServiceClient GatheringRulesServiceClient
+}
+
+type GatheringRulesServiceClient interface {
+	RecvGatheringRules(ctx context.Context, endpoint string) ([]byte, error)
 }
```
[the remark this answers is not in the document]
Sergey1011010: Well, it would be harder to unit test it. Now I'm using a very clean and simple implementation of MockClient https://github.com/openshift/insights-operator/pull/583/files#diff-5819ed4e968701d806c6aac65c11f9e554f0924d6acb9f5de8a09f9e2bedbc8fR248

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: They are stored in `g.gatheringRules`. Yes, if there's no service on the very first run, it's gonna be empty. Cache works if the service disappears later.

pkg/gatherers/conditional/conditional_gatherer.go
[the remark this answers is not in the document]
tremes: Ahh OK, I haven't seen the tests yet. Then it's completely OK. Thanks

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -330,6 +160,39 @@ func (g *Gatherer) GetGatheringFunctions(ctx context.Context) (map[string]gather
...
+	config := g.configurator.Config()
+	if config == nil {
+		return "", fmt.Errorf("config is nil")
+	}
+
+	if g.gatheringRulesServiceClient == nil {
+		return "", fmt.Errorf("gathering rules service client is nil")
+	}
+
+	rulesBytes, err := g.gatheringRulesServiceClient.RecvGatheringRules(ctx, config.ConditionalGathererEndpoint)
```
tremes, line 188: What about adding some info log around here please? Something like fetching or pulling the conditions from endpoint bla bla bla?

pkg/gatherers/conditional/conditional_gatherer.go
tremes: I see. OK

tremes: Probably a detail, but when the `g.gatheringRules` are empty then the metadata is recorded with `null`. It would be better to put the error message there or something like that IMO.

tremes: Because this will never work in disconnected clusters....

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: https://github.com/RedHatInsights/insights-operator-gathering-conditions/pull/10
I'll also update the tag in the service once this one is merged to trigger a new build of the service

pkg/insights/insightsclient/requests.go
Sergey1011010: done

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: done

pkg/gatherers/conditional/conditional_gatherer.go
tremes: Thank you

pkg/controller/controller.go
```diff
@@ -0,0 +1,58 @@
+package controller
```
tremes, line 1: I would rename this file since it has nothing to do with controller right? Perhaps `kubeconfigs.go` or something like that.

pkg/gatherers/conditional/conditional_gatherer.go
tremes: I thought it would be good to put the endpoint URL in the log. Theoretically, it can happen that someone sets wrong endpoint and then it can help when troubleshooting.

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: The service update PR https://github.com/RedHatInsights/insights-operator-gathering-conditions-service/pull/140

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: done

pkg/controller/controller.go
Sergey1011010: done

tremes:
I reviewed this one couple times. Good job. Thank you.

PR #593, opened 2022-03-14
Create an alternate IO deployment manifest excluding the NodeSelector

rluders:
This PR adds the some of the `build-machinery-go` scripts to our makefile in order to allow us to patch the manifests. This is necessary to allow the IO to run on Hypershift Hosted Clusters.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

n/a

## Documentation

n/a

## Unit Tests

n/a

## Privacy

n/a

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-7458


gomod.go
```diff
@@ -0,0 +1,7 @@
+package gomod_required_packages
```
rluders, line 1: :warning: This file is required to include the `build-machinery-go` dependency, and it can be also used to add some "extra" dependencies that are not actually being used in our Go code, but we don't want it to be an indirect dependency or to be removed by the `go mod tidy`.

manifests/07-cluster-operator.yaml
```diff
@@ -6,7 +6,6 @@ metadata:
     exclude.release.openshift.io/internal-openshift-hosted: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
-    include.release.openshift.io/ibm-cloud-managed: "true"
```
tremes, line 9: I think this shouldn't be removed.

tremes, line 9: I think this shouldn't be removed.

Makefile
```diff
@@ -80,9 +93,9 @@ run: ## Executes the insights operator
 		--config=$(CONFIG) \
 		$(RUN_FLAGS)
 
-.PHONY: build
-build: ## Compiles the insights operator
+build: | verify ## Compiles the insights operator
 	go build -o ./bin/insights-operator ./cmd/insights-operator
+.PHONY: build
```
tremes, line 98: We don't need this change. There's a new make target and you can call:
```
make update-profile-manifests
```

gomod.go
tremes: OK, but I would not add it into file called `gomod.go`. I think introducing some new package as in console-operator (see https://github.com/openshift/console-operator/tree/master/pkg/dependencymagnet) is better.

tremes:
Reviewed and verified that the manifest works on the HyperShift hosted cluster and that it's correctly generated with the `make update-profile-manifests` command. Thank you.
/lgtm

PR #603, opened 2022-04-01
Implement fingerprint for records

Sergey1011010:
## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- ``

## Unit Tests

- `pkg/gather/gather_test.go`

## Privacy

No new data was collected

## Changelog

No

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-6209

### DIFF
--- a/pkg/gather/gather.go
+++ b/pkg/gather/gather.go
@@ -89,85 +89,15 @@ func CollectAndRecordGatherer(
 		return nil, err
 	}
 
-	gathererName := gatherer.GetName()
-
 	var allErrors []error
 	var functionReports []GathererFunctionReport
 	totalNumberOfRecords := 0
 
 	for result := range resultsChan {
-		var recordWarnings []error
-		var recordErrs []error
-
-		if result.Panic != nil {
-			recordErrs = append(recordErrs, fmt.Errorf("panic: %v", result.Panic))
-			klog.Error(fmt.Errorf(
-				`gatherer "%v" function "%v" panicked with the error: %v`,
-				gathererName, result.FunctionName, result.Panic,
-			))
-			allErrors = append(allErrors, fmt.Errorf(`function "%v" panicked`, result.FunctionName))
-		}
-
-		for _, err := range result.Errs {
-			if w, isWarning := err.(*types.Warning); isWarning {
-				recordWarnings = append(recordWarnings, w)
-				klog.Warningf(
-					`gatherer "%v" function "%v" produced the warning: %v`, gathererName, result.FunctionName, w,
-				)
-			} else {
-				recordErrs = append(recordErrs, err)
-				klog.Errorf(
-					`gatherer "%v" function "%v" failed with the error: %v`,
-					gathererName, result.FunctionName, err,
-				)
-				allErrors = append(allErrors, fmt.Errorf(`function "%v" failed with an error`, result.FunctionName))
-			}
-		}
-
-		recordedRecs := 0
-		for _, r := range result.Records {
-			wasRecorded := true
-			if errs := rec.Record(r); len(errs) > 0 {
-				for _, err := range errs {
-					if w, isWarning := err.(*types.Warning); isWarning {
-						recordWarnings = append(recordWarnings, w)
-						klog.Warningf(
-							`issue recording gatherer "%v" function "%v" result "%v" because of the warning: %v`,
-							gathererName, result.FunctionName, r.GetFilename(), w,
-						)
-					} else {
-						recordErrs = append(recordErrs, err)
-						klog.Errorf(
-							`error recording gatherer "%v" function "%v" result "%v" because of the error: %v`,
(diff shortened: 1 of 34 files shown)


pkg/record/record.go
```diff
@@ -15,13 +16,29 @@ const (
...
+		if err != nil {
+			return "", err
+		}
+
+		h := sha256.New()
+		h.Write(content)
+		r.fingerprint = hex.EncodeToString(h.Sum(nil))
+	}
+
+	return r.fingerprint, nil
```
tremes, line 37: Maybe this could return `content` as well, because right now we marshal each item twice right? First here and then [in the recorder](https://github.com/openshift/insights-operator/pull/603/files#diff-93cbf5b12a5398c622a4f899c81da8bdca98a02a5cf178e90f1b6b43e6b7005cR76)

pkg/recorder/recorder.go
```diff
@@ -85,12 +93,33 @@ func (r *Recorder) Record(rec record.Record) error {
...
+	if existingPath, found := r.recordedFingerprints[fingerprint]; found {
+		// this doesn't necessarily mean it's an error. There can be a collision after hashing
+		errs = append(errs, &types.Warning{UnderlyingValue: fmt.Errorf(
+			`a record with the same fingerprint "%v" was already recorded at path "%v", `+
+				`recording a new one at path "%v". `, fingerprint, existingPath, recordName,
+		)})
+	}
+
 	r.records[memoryRecord.Name] = memoryRecord
+	r.recordedFingerprints[fingerprint] = recordName
```
tremes, line 119: I was thinking if we need the two maps....Anyway, won't this `recordedFingerprints` map just get bigger and bigger potentially? I think we should clear it same as the `records` map.

Sergey1011010: oops

pkg/record/record.go
Sergey1011010: fixed

natiiix (approved):
Tested on a testing cluster. An archive is generated and looks alright. I can see that the gathered records now have actual fingerprint values in the log. Nothing seems to be broken as a result of the changes. I have no issues with the changes made to the code either.
/lgtm

pkg/gather/gather.go
```diff
@@ -78,57 +82,108 @@ func CollectAndRecordGatherer(
...
+		functionReports = append(functionReports, report)
+		totalNumberOfRecords += report.RecordsCount
+	}
 
-		for _, err := range result.Errs {
-			errs = append(errs, fmt.Errorf(
-				"gatherer %v's function %v failed with error: %v",
+	functionReports = append(functionReports, GathererFunctionReport{
+		FuncName:     gatherer.GetName(),
+		Duration:     time.Since(startTime).Milliseconds(),
```
tremes, line 105: Does it have to be set?

pkg/recorder/recorder.go
```diff
@@ -85,12 +88,40 @@ func (r *Recorder) Record(rec record.Record) error {
...
+
+	if existingPath, found := r.recordedFingerprints[fingerprint]; found {
+		existingRecord, found := r.records[existingPath]
+		if !found {
+			existingRecord = &record.MemoryRecord{Name: "unable to find the record"}
+		}
+		// this doesn't necessarily mean it's an error. There can be a collision after hashing
+		errs = append(errs, &types.Warning{UnderlyingValue: fmt.Errorf(
+			`the record %v with the same fingerprint %v was already recorded, `+
+				`recording another one with a different path %v`,
```
tremes, line 117: It's just a nit, but I got message like:
> "warning: the record MemoryRecord{Name: \"config/configmaps/openshift-config/admin-kubeconfig-client-ca/ca-bundle.crt\", At: \"2022-04-28 14:38:29.852686256 +0200 CEST m=+54.467188189\", len(Data): 65, Fingerprint: \"34cdf5d36d45a9e3ee7c2be31546e3d937ae4ebf144ac07b0210c06a1610a94d\"} with the same fingerprint 34cdf5d36d45a9e3ee7c2be31546e3d937ae4ebf144ac07b0210c06a1610a94d was already recorded, recording another one with a different path MemoryRecord{Name: \"config/configmaps/openshift-config/etcd-ca-bundle/ca-bundle.crt\", At: \"2022-04-28 14:38:29.852757424 +0200 CEST m=+54.467259358\", len(Data): 65, Fingerprint: \"34cdf5d36d45a9e3ee7c2be31546e3d937ae4ebf144ac07b0210c06a1610a94d\"}",

which is pretty verbose IMO. The same fingerprint is mentioned three times. I'm not sure the time and length of the data is of any use.

tremes:
One nitpick, but it looks good to me. I tried to run locally and it seemed good. Thanks.

tremes:
I reviewed it and did basic test. It looks good to me. Thanks!
/lgtm

PR #604, opened 2022-04-04
Bug 2072842: Gather namespace names with overlapping UIDs

tremes:
This adds a new gatherer finding namespaces with overlapping UID ranges. The UID range is obtained from the `openshift.io/sa.scc.uid-range` annotation and examples of the overlapping and non-overlapping ranges are:

- 1000670000/10000 and 1000675000/10000 **are overlapping**
- 1000670000/10000 and 1000680000/10000 **are not overlapping**
For more details see https://issues.redhat.com/browse/CCXDEV-7821

**Steps to reproduce:**

- create some namespaces in your cluster (and make some of them with overlapping UIDs)
- you can use following yaml to create the namespaces:
```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    openshift.io/description: ""
    openshift.io/display-name: ""
    openshift.io/requester: kube:admin
    openshift.io/sa.scc.uid-range: 1000600000/10000
  labels:
    kubernetes.io/metadata.name: test-1
  name: test-1
spec:
```
```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    openshift.io/description: ""
    openshift.io/display-name: ""
    openshift.io/requester: kube:admin
    openshift.io/sa.scc.uid-range: 1000700000/10000
  labels:
    kubernetes.io/metadata.name: test-2
  name: test-2
spec:
```
```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    openshift.io/description: ""
    openshift.io/display-name: ""
    openshift.io/requester: kube:admin
    openshift.io/sa.scc.uid-range: 1000620000/7000
  labels:
    kubernetes.io/metadata.name: test-3
  name: test-3
spec:
```

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/config/namespaces_with_overlapping_uids.json`

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids_test.go`

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids.go
+++ b/pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids.go
@@ -1,0 +1,180 @@
+package clusterconfig
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+type uidRange struct {
+	starting int64
+	length   int64
+}
+
+type namespaceWithRange struct {
+	name string
+	uidRange
+}
+
+// IsOverlappingWith checks if the UIDRange is overlapping with the provided one
+func (u *uidRange) IsOverlappingWith(r uidRange) bool {
+	uSum := u.starting + u.length
+	rSum := r.starting + r.length
+	return (uSum > r.starting && uSum <= rSum) || (rSum > u.starting && rSum <= uSum)
+}
+
+func (u *uidRange) String() string {
+	return fmt.Sprintf("%d/%d", u.starting, u.length)
+}
+
+// GatherNamespacesWithOverlappingUIDs gathers namespaces with overlapping UID ranges
+//
+// The Kubernetes api https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/namespace.go
+// Response is an array of arrays of namespaces with overlapping UIDs. Each namespace is represented by its name and the UID range value
+// from the "openshift.io/sa.scc.uid-range" annotation
+//
+// * Location in archive: config/namespaces_with_overlapping_uids
+// * Id in config: clusterconfig/overlapping_namespace_uids
+// * Since versions:
+//   * 4.11+
+func (g *Gatherer) GatherNamespacesWithOverlappingUIDs(ctx context.Context) ([]record.Record, []error) {
+	gatherKubeClient, err := kubernetes.NewForConfig(g.gatherProtoKubeConfig)
+	if err != nil {
+		return nil, []error{err}
+	}
+	return gatherNamespacesWithOverlappingUIDs(ctx, gatherKubeClient.CoreV1())
+}
+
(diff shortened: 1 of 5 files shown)


pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids.go
```diff
@@ -0,0 +1,169 @@
...
+				if es, ok := resultSet.BothOverlap(n1, n2); ok {
+					es.Insert(n1, n2)
+				} else {
+					s := NewSet()
+					s.Insert(n1, n2)
+					resultSet = append(resultSet, s)
+				}
+			}
+		}
+	}
```
tremes, line 90: Thi is probably not the most performant algorithm....so I am open to suggestions, optimizations,......

pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids.go
```diff
@@ -0,0 +1,180 @@
...
+	length   int64
+}
+
+type namespaceWithRange struct {
+	name string
+	uidRange
+}
+
+// IsOverlappingWith checks if the UIDRange is overlapping with the provided one
+func (u *uidRange) IsOverlappingWith(r uidRange) bool {
```
Sergey1011010, line 28: I think `u` could be passed by value too

pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids.go
```diff
@@ -0,0 +1,180 @@
...
+	var errs []error
+	for i := range nsList.Items {
+		ns := nsList.Items[i]
+		uidRangeString := ns.Annotations["openshift.io/sa.scc.uid-range"]
+		r, err := uidStringToRange(uidRangeString)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("can't read uid range of the %s namespace", ns.Name))
+			continue
+		}
+		namespaces = append(namespaces, namespaceWithRange{ns.Name, *r})
```
Sergey1011010, line 71: Isn't linter gonna complain we don't use field names?

pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids.go
Sergey1011010: I'd rather use some library for looking for overlapping intervals. Probably this one https://github.com/toberndo/go-stree but there can be something better of course

pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids.go
tremes: Looks like it isn't

pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids.go
tremes: Thanks. I will take a look, though I am not sure if it's worth adding a new dependency when it's only in one place.

tremes: Well I was experimenting a little bit with the library - see https://go.dev/play/p/VZOQGGAgLwG (why the array has so many items?) and I don't think it's a good fit for what we need here.

natiiix (approved):
Running the image generated from this PR on a testing cluster and performing the reproduction steps in the PR's description has successfully resulted in an Insights Operator archive containing the mentioned JSON file with two detected overlaps between `openshift` and `test-1`, and `openshift-ingress-canary` and `test-3`, as expected.
I don't believe it's necessary to make any modifications to the code, personally.
/lgtm

PR #607, opened 2022-04-11
Implement Prometheus Collector pattern

natiiix:
This PR changes the Prometheus metrics registry/server used by IO from the `legacyregistry` to an HTTP registry served directly from the IO code. An update of the Prometheus library was included in the changes to make sure the latest version of this dependency was used.

## Manual Testing Steps

0. Start a testing cluster using the OCP image based on this PR (e.g., via cluster-bot).
0. Check in the web console that the `insights-operator` target is displayed as "up" and being properly scraped and that HTTPS is being used.
0. In the Metrics tab, try querying some of the `insightsclient_*` metrics to make sure they are being reported.

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

N/A

## Documentation

N/A

## Unit Tests

N/A

## Privacy

Yes. User data are not involved in the changes.

## Breaking Changes

No

## References

Jira Task: https://issues.redhat.com/browse/CCXDEV-7475

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/openshift/client-go v0.0.0-20220525160904-9e1acff93e4a
 	github.com/openshift/installer v0.9.0-master.0.20191219195746-103098955ced
 	github.com/openshift/library-go v0.0.0-20220525173854-9b950a41acdc
-	github.com/prometheus/common v0.32.1
+	github.com/prometheus/client_golang v1.12.1
 	github.com/spf13/cobra v1.4.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
--- a/pkg/insights/insightsreport/insightsreport.go
+++ b/pkg/insights/insightsreport/insightsreport.go
@@ -15,7 +15,9 @@ import (
 	"github.com/openshift/insights-operator/pkg/authorizer"
 	"github.com/openshift/insights-operator/pkg/config/configobserver"
 	"github.com/openshift/insights-operator/pkg/controllerstatus"
+	"github.com/openshift/insights-operator/pkg/insights"
 	"github.com/openshift/insights-operator/pkg/insights/insightsclient"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 // Controller gathers the report from Smart Proxy
@@ -41,7 +43,7 @@ type InsightsReporter interface {
 var (
 
 	// insightsStatus contains a metric with the latest report information
-	insightsStatus = metrics.NewGaugeVec(&metrics.GaugeOpts{
+	insightsStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "health",
 		Subsystem: "statuses",
 		Name:      "insights",
@@ -290,4 +292,14 @@ func updateInsightsMetrics(report SmartProxyReport) {
 		return
 	}
 	insightsLastGatherTime.Set(float64(t.Unix()))
+}
+
+func init() {
+	insights.RegisterMetricCollectors(insightsStatus)
+
+	insightsStatus.WithLabelValues("low").Set(float64(-1))
+	insightsStatus.WithLabelValues("moderate").Set(float64(-1))
+	insightsStatus.WithLabelValues("important").Set(float64(-1))
+	insightsStatus.WithLabelValues("critical").Set(float64(-1))
+	insightsStatus.WithLabelValues("total").Set(float64(-1))
 }
--- a/pkg/insights/insightsreport/insightsreport.go
+++ b/pkg/insights/insightsreport/insightsreport.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/component-base/metrics"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/insights-operator/pkg/authorizer"
@@ -39,8 +38,11 @@ type InsightsReporter interface {
 	ArchiveUploaded() <-chan struct{}
 }
 
-var (
+const (
+	insightsLastGatherTimeName = "insights_last_gather_time"
+)
 
+var (
 	// insightsStatus contains a metric with the latest report information
 	insightsStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "health",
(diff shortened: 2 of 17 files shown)


go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/openshift/client-go v0.0.0-20211209144617-7385dd6338e3
 	github.com/openshift/installer v0.9.0-master.0.20191219195746-103098955ced
 	github.com/openshift/library-go v0.0.0-20220110124235-d70e915b42bd
+	github.com/prometheus/client_golang v1.11.0
```
tremes, line 16: Why not the latest? :)

pkg/insights/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
+	)
+	startMetricsServer()
+}
+
+// startMetricsServer starts an HTTP server for the Insights metrics registry.
+func startMetricsServer() {
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(insightsMetricsRegistry, promhttp.HandlerOpts{}))
+	go http.ListenAndServe(":8080", mux)
```
tremes, line 28: I think it would be slightly better to move it to anonymous function and handle the error return value.

pkg/insights/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+	insightsMetricsRegistry *prometheus.Registry
+)
+
+func init() {
+	insightsMetricsRegistry = prometheus.NewRegistry()
+	MustRegisterMetricCollectors(
+		collectors.NewGoCollector(),
+		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
+	)
+	startMetricsServer()
```
tremes, line 21: Is it a good practice to start http server within `init` function? I am not sure....I was reading https://go.dev/doc/effective_go#init, but still....

go.mod
natiiix: I didn't specifically choose the version, but my guess would be that the latest is incompatible with the rest of our packages. When I tried to `go get` the latest version now, it also updated a ton of other packages, so I'm not sure if that's something we want necessarily.

pkg/insights/metrics.go
natiiix: That makes sense, but I'm wondering what should we do in case of an error. Kill the entire IO or just restart the server in a loop to not interfere with the rest of IO functionality?

pkg/insights/metrics.go
natiiix: I was wondering what might be a good place for it, and the `init` function kind of works because it's executed before any regular code, which means the metrics will be available regardless of anything else breaking, but at the same time, I think you're right and it should be something in `start.go` perhaps, but I'm not too sure where exactly to make sure the metrics are always available. The thing is, IO has multiple "commands" (executables), so I don't know if only the `insights-operator` one should start the metrics server.

go.mod
tremes: Incompatible? In what sense (any example pls)? Interesting....it seems we will have to update it one day anyway.

pkg/insights/metrics.go
tremes: I am not sure if we need some extra logic. I would just start with logging the error.

pkg/insights/metrics.go
tremes: I think we could add it to https://github.com/openshift/insights-operator/blob/master/pkg/controller/operator.go#L46 method.

go.mod
natiiix: I don't know why it wants to update the other packages as well in this case, but generally speaking, it could simply be using a function/constant/whatever that isn't present in the version we're currently using, so it needs to update the dependency to a version that include that feature. It's not updating anything related to K8s, just some Prometheus side-libraries, Golang libraries that aren't built-in and `xxhash`.

pkg/insights/metrics.go
natiiix: @tremes Okay, but what if it crashes? Will IO continue happily running without the metrics? That seems like it could lead to some problems, so I think having it in an infinite loop with a bit of a sleep in between restarts will be fine. That way, if something bad happens to the server, it will just restart and the metrics will be available again. Having IO run completely without metrics seems like it could cause trouble with debugging and the error message could get lost in the log quite easily. Having IO crash when the server crashes is a possibility, but that could totally break IO in case of some funny networking misconfiguration, possibly, which would then crashloop IO, and that doesn't sound very good to me, either.

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -38,10 +38,13 @@ type InsightsReporter interface {
 	ArchiveUploaded() <-chan struct{}
 }
 
-var (
+const (
+	insightsLastGatherTimeName = "insights_last_gather_time"
+)
```
tremes, line 43: This is renaming the existing metric. Please preserve the original name `insightsclient_last_gather_time` since there's already functionality depending on this metric.

natiiix: Did the name of the metric change, by any chance? Because I'm pretty sure I just copy-pasted the string, so I don't understand why it would be any different from what it originally was, but sure, I'll change that.

tremes: You changed it. See:
- https://github.com/openshift/insights-operator/pull/607/files#diff-d543de2f94800472455800fb3d022ea29cc4d84486bc1ca6a9e2b7d36ae37124L55 
- https://github.com/openshift/insights-operator/pull/607/files#diff-d543de2f94800472455800fb3d022ea29cc4d84486bc1ca6a9e2b7d36ae37124R42

tremes:
I reviewed it and experimented with that. It looks good to me. Thank you!
/lgtm

pkg/insights/insightsreport/insightsreport.go
natiiix: @tremes It wouldn't be visible in the commit history if the name was changed in parallel to my changes because I would have rebased my commits on top of the original name change. It's pretty weird, but whatever.

alvaroaleman:
@natiiix @tremes this needs to also update the IBM cloud managed deployment once it is re-introduced: https://github.com/openshift/insights-operator/pull/636

tremes:
@alvaroaleman Yes. Thanks for heads up. There's https://github.com/openshift/insights-operator/pull/632

PR #618, opened 2022-05-05
Bug 2081997: Fix the clusteroperator conditions values when IO is

tremes:
disabled

Fix for clusteroperator conditions when the operator is disabled

**Steps to reproduce the original problem:**

- make IO degraded - e.g by changing the upload endpoint to some non-existing URL
- make IO disabled - by removing `cloud.openshift.com` token from the`pull-secret` 
- IO is still degraded (and other CO conditions are wrong as well), but it should not be

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Unit Tests

This will require a new integration test

## Privacy

## Breaking Changes

No

## References

https://bugzilla.redhat.com/show_bug.cgi?id=2081997

### DIFF
--- a/pkg/controller/status/controller.go
+++ b/pkg/controller/status/controller.go
@@ -338,21 +338,21 @@ func updateControllerConditions(cs *conditions, ctrlStatus *controllerStatus,
 	}
 
 	// handle when has errors
-	if es := ctrlStatus.getStatus(ErrorStatus); es != nil {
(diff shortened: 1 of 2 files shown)


natiiix (approved):
Tested on a testing cluster. Once the Insights Operator was disabled, as described in the provided steps, the `UploadDegraded` condition disappeared and the `Degraded` condition was in a `False` state, so the operator seemed healthy based on the conditions, even if there was a misconfiguration, which is the correct behaviour.
/lgtm

PR #620, opened 2022-05-08
save conditional gatherer endpoint and firing alerts in the metadata

Sergey1011010:
## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/insights-operator/conditional-gatherer-rules.json`

## Documentation

No changes

## Unit Tests

- `path/to/file_test.go`

## Privacy

No new data was collected

## Changelog

No

## Breaking Changes

No

## References

### DIFF
--- a/docs/insights-archive-sample/insights-operator/conditional-gatherer-rules.json
+++ b/docs/insights-archive-sample/insights-operator/conditional-gatherer-rules.json
@@ -193,10 +193,5 @@
       "was_triggered": false
     }
   ],
-  "endpoint": "https://console.redhat.com/api/gathering/gathering_rules",
-  "firing_alerts": [
-    "AlertmanagerClusterDown",
-    "TargetDown",
-    "Watchdog"
-  ]
+  "endpoint": "https://console.redhat.com/api/gathering/gathering_rules"
 }
--- a/docs/insights-archive-sample/insights-operator/conditional-gatherer-rules.json
+++ b/docs/insights-archive-sample/insights-operator/conditional-gatherer-rules.json
@@ -192,5 +192,11 @@
       "errors": null,
       "was_triggered": false
     }
+  ],
+  "endpoint": "https://console.redhat.com/api/gathering/gathering_rules",
+  "firing_alerts": [
+    "AlertmanagerClusterDown",
+    "TargetDown",
+    "Watchdog"
   ]
 }
--- a/pkg/gatherers/conditional/conditional_gatherer.go
+++ b/pkg/gatherers/conditional/conditional_gatherer.go
@@ -77,10 +77,9 @@ type GatheringRuleMetadata struct {
 
 // GatheringRulesMetadata stores metadata about gathering rules
 type GatheringRulesMetadata struct {
-	Version      string                  `json:"version"`
-	Rules        []GatheringRuleMetadata `json:"rules"`
-	Endpoint     string                  `json:"endpoint"`
-	FiringAlerts []string                `json:"firing_alerts"`
+	Version  string                  `json:"version"`
+	Rules    []GatheringRuleMetadata `json:"rules"`
+	Endpoint string                  `json:"endpoint"`
 }
 
 // GetName returns the name of the gatherer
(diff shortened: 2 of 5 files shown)


docs/insights-archive-sample/insights-operator/conditional-gatherer-rules.json
```diff
@@ -192,5 +192,11 @@
       "errors": null,
       "was_triggered": false
     }
+  ],
+  "endpoint": "https://console.redhat.com/api/gathering/gathering_rules",
+  "firing_alerts": [
+    "AlertmanagerClusterDown",
+    "TargetDown",
+    "Watchdog"
```
tremes, line 200: Isn't that redundant information,please? I mean we have `was_triggered` flag, which implies that an alert was firing right?

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -226,52 +247,46 @@ func (g *Gatherer) updateCache(ctx context.Context) {
 
 func (g *Gatherer) updateAlertsCache(ctx context.Context, metricsClient rest.Interface) error {
 	const logPrefix = "conditional gatherer: "
+	klog.Info(logPrefix + "updating alerts cache for conditional gatherer")
 
 	g.firingAlerts = make(map[string][]AlertLabels)
 
-	data, err := metricsClient.Get().AbsPath("federate").
+	data, err := metricsClient.Get().
+		AbsPath("api/v1/query").
```
tremes, line 255: Is it better to use the common API versus the federate one?

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -226,52 +247,46 @@ func (g *Gatherer) updateCache(ctx context.Context) {
...
 
-	var parser expfmt.TextParser
-	metricFamilies, err := parser.TextToMetricFamilies(bytes.NewReader(data))
+	var response struct {
+		Data struct {
+			Results []struct {
+				Labels map[string]string `json:"metric"`
+			} `json:"result"`
+		} `json:"data"`
+	}
```
tremes, line 269: Maybe time to use the Prometheus client types....

docs/insights-archive-sample/insights-operator/conditional-gatherer-rules.json
Sergey1011010: It can be helpful for debugging and also could be used in the integration tests later

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: I just unified it with the integration tests so that we use the same API everywhere to eliminate strange bugs in the future

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: I'll change it, thanks

tremes: It's not super important for this PR. You can experiment in another one if you want. Not exactly sure, but perhaps https://github.com/prometheus/client_golang/blob/main/api/prometheus/v1/api.go

docs/insights-archive-sample/insights-operator/conditional-gatherer-rules.json
tremes: Having something helpful for debugging and in the integration tests doesn't mean we have to add it to the archive ....at least for me :-P Anyway, it's probably not a big deal (although once something is in the archive then it's pretty difficult to remove it), I am just saying that I think the information is already there.

Sergey1011010: It's not too important tho, we could delete it if you want

tremes: Good point with the conditions potentially being more complicated, but we don't have any at the moment AFAIK so I wouldn't add it now.

Sergey1011010: changed it

tremes: Thanks so we need to change the sample archive as well right.

Sergey1011010: oops

tremes:
I didn't test with any firing alert, but it looks good to me. Thanks.
/lgtm

PR #625, opened 2022-05-17
gather io configuration

Sergey1011010:
## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/config/secrets/openshift-config/support/data.json`
- `docs/insights-archive-sample/insights-operator/config/combined.json`
- `docs/insights-archive-sample/insights-operator/config/from_secret.json`

## Documentation

- `docs/gathered-data.md`

## Privacy

Yes. The data is anonymized.

## Changelog

No

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-7877

### DIFF
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -885,6 +885,17 @@ gathers the alerts that have been silenced.
   * 4.10+
 
 
+## SupportSecret
+
+gathers anonymized support secret if there is any
+
+* Location in archive: config/secrets/openshift-config/support/data.json
+    (can be ommited if the secret doesn't exist)
+* Id in config: support_secret
+* Since version:
+  * 4.12+
+
+
 ## TSDBStatus
 
 gathers the Prometheus TSDB status.
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -890,10 +890,10 @@ gathers the alerts that have been silenced.
 gathers anonymized support secret if there is any
 
 * Location in archive: config/secrets/openshift-config/support/data.json
-    (can be ommited if the secret doesn't exist)
+    (can be omitted if the secret doesn't exist)
 * Id in config: support_secret
 * Since version:
-  * 4.12+
+  * 4.11+
 
 
 ## TSDBStatus
--- a/docs/insights-archive-sample/config/secrets/openshift-config/support/data.json
+++ b/docs/insights-archive-sample/config/secrets/openshift-config/support/data.json
@@ -1,0 +1,3 @@
+{
+  "conditionalGathererEndpoint": "aHR0cDovL2xvY2FsaG9zdDo4MDgw"
+}
--- a/pkg/config/configobserver/configobserver.go
+++ b/pkg/config/configobserver/configobserver.go
@@ -37,6 +37,7 @@ type Controller struct {
 	defaultConfig config.Controller
 	tokenConfig   *config.Controller
 	secretConfig  *config.Controller
+	supportSecret *v1.Secret
 	config        *config.Controller
 	checkPeriod   time.Duration
 	listeners     []chan struct{}
@@ -78,6 +79,12 @@ func (c *Controller) Config() *config.Controller {
 	return c.config
 }
 
+func (c *Controller) SupportSecret() *v1.Secret {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	return c.supportSecret
+}
+
 // ConfigChanged subscribe for config changes
 // 1.Param: A channel where the listener is notified that the config has changed.
 // 2.Param: A func which can be used to unsubscribe from the config changes.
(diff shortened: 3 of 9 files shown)


rluders (approved):
Just looking at the code, it seems good. Very clear and direct implementation.
/lgtm

pkg/gatherers/clusterconfig/io_config.go
```diff
@@ -0,0 +1,84 @@
...
+// * Location in archive: insights-operator/config/combined.json
+// * Location in archive: insights-operator/config/from_secret.json
+// * Location in archive: config/secrets/openshift-config/support/data.json
+// * Id in config: io_config
+// * Since version:
+//   * 4.12+
+func (g *Gatherer) GatherIOConfig(context.Context) ([]record.Record, []error) {
+	combinedConfig := g.configObserver.Config()
+	records := []record.Record{{
+		Name: "insights-operator/config/combined",
```
tremes, line 22: The original goal was to gather just the support secret definition. I wouldn't complicate it with this "combined" data. Shouldn't be the values (when running in container) of the `data.json` and `combined.json` the same?

docs/insights-archive-sample/config/secrets/openshift-config/support/data.json
```diff
@@ -0,0 +1,3 @@
+{
+  "conditionalGathererEndpoint": "aHR0cDovL2xvY2FsaG9zdDo4MDgw"
```
tremes, line 2: I don't understand why this is in a separate file. I think having one file containing the data would be clearer.

pkg/gatherers/clusterconfig/io_config.go
Sergey1011010: I thought having the actual values may be more useful to whoever's gonna use the data for analysis. Should I delete `combined` and `from_secret` collectors?

docs/insights-archive-sample/config/secrets/openshift-config/support/data.json
Sergey1011010: This one contains the values of the support secret as is (with anonymization of course), the other one is the representation of IO's actual config (with anonymization again).

pkg/gatherers/clusterconfig/io_config.go
Sergey1011010: By default `data.json` won't exist because there's no support secret on the cluster, same applies to `from_secret.json` which is parsed configuration of the support secret. `combined.json` always exists and contains the actual state of config IO uses.

tremes: When there's no support secret then the default (known) config is used right? Why to gather something (in each archive) that we know?

tremes: Yeah there can be some differences in the various version, but we have it in Git right? It's basically the https://github.com/openshift/insights-operator/blob/master/config/pod.yaml so I would keep (in the archive) only the data coming from the `support` secret.

Sergey1011010: changed it

pkg/config/configobserver/configobserver.go
```diff
@@ -150,21 +159,23 @@ func (c *Controller) updateToken(ctx context.Context) error {
...
 	secret, err := c.fetchSecret(ctx, "support")
 	if err != nil {
 		return err
 	}
 
-	if secret != nil {
-		nextConfig, err = LoadConfigFromSecret(secret)
+	c.supportSecret = secret
+	if secret == nil {
+		c.setSecretConfig(nil)
```
tremes, line 170: Is this necessary? I think it will be `nil` anyway.

tremes:
One minor comment added. It would be nice to add some test for the new gatherer.

pkg/config/configobserver/configobserver.go
Sergey1011010: I think I needed it to reset the secret config in case the support secret was deleted

tremes: Ohh now I see it. We've done it before as well, but not in such obvious way

docs/gathered-data.md
```diff
@@ -885,6 +885,17 @@ gathers the alerts that have been silenced.
...
 
+## SupportSecret
+
+gathers anonymized support secret if there is any
+
+* Location in archive: config/secrets/openshift-config/support/data.json
+    (can be ommited if the secret doesn't exist)
+* Id in config: support_secret
+* Since version:
+  * 4.12+
```
tremes, line 896: I think we are about to merge it so please change to `4.11+`

Sergey1011010: done

tremes:
Reviewed & tested. Thanks!

PR #629, opened 2022-05-26
bump(k8s v0.24.0)

tremes:

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Documentation

No docs

## Unit Tests

No tests

## Privacy

## Breaking Changes

No

## References


pkg/gatherers/clusterconfig/node_logs_test.go
```diff
@@ -146,14 +147,30 @@ func Test_requestNodeLog(t *testing.T) {
...
+				SetHeader("Accept", "text/plain, */*").
+				SetHeader("Accept-Encoding", "gzip"),
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := requestNodeLog(tt.args.client, tt.args.uri, tt.args.tail, tt.args.unit); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("requestNodeLog() = %v, want %v", got, tt.want)
-			}
+			got := requestNodeLog(tt.args.client, tt.args.uri, tt.args.tail, tt.args.unit)
```
tremes, line 160: >reflect.DeepEqual(got, tt.want)

This is failing with the update to Kubernetes client-go 1.24. Not 100% sure, but I guess that it fails to compare the [`retryFn`](https://github.com/kubernetes/client-go/blob/v0.24.0/rest/request.go#L122) which was previously defined as the [`WithRetry interface`](https://github.com/kubernetes/client-go/blob/v0.23.1/rest/request.go#L114).

Another option is to remove this test (and probably the `requestNodeLog` function as well) since the test doesn't have much value IMO.

natiiix (approved):
Obviously, it is not possible to go through the individual changes here, and aside from some test fixes, there are no significant changes to the IO code.
If it builds and tests pass (i.e., CI doesn't complain), it should be fine.
/lgtm

tremes:
Only version update.
/label px-approved
/label qe-approved
/label docs-approved

PR #633, opened 2022-06-02
make cluster version condition more flexible

Sergey1011010:
Fixes a bug with conditional gatherer integration tests

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No changes

## Documentation

No changes

## Unit Tests

- `pkg/gatherers/conditional/conditions_test.go`

## Privacy

No new data was collected

## Changelog

No

## Breaking Changes

No

## References

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ build: ## Compiles the insights operator
 
 .PHONY: build-debug
 build-debug: ## Compiles the insights operator in debug mode
-	go build $(GO_BUILD_FLAGS) -gcflags="all=-N -l" \
+	go build -gcflags="all=-N -l" \
 		-o ./bin/insights-operator ./cmd/insights-operator
 
 ## --------------------------------------
--- a/pkg/gatherers/conditional/conditional_gatherer.go
+++ b/pkg/gatherers/conditional/conditional_gatherer.go
@@ -286,13 +286,16 @@ func (g *Gatherer) updateAlertsCache(ctx context.Context, metricsClient rest.Int
 }
 
 func (g *Gatherer) updateVersionCache(ctx context.Context, configClient configv1client.ConfigV1Interface) error {
+	const logPrefix = "conditional gatherer: "
+	klog.Info(logPrefix + "updating version cache for conditional gatherer")
+
 	clusterVersion, err := configClient.ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
 
 	g.clusterVersion = clusterVersion.Status.Desired.Version
-
+	klog.Infof(logPrefix+"cluster version is '%v'", g.clusterVersion)
 	return nil
 }
 
--- a/pkg/gatherers/conditional/conditional_gatherer.go
+++ b/pkg/gatherers/conditional/conditional_gatherer.go
@@ -239,8 +239,7 @@ func (g *Gatherer) updateCache(ctx context.Context) {
 }
 
 func (g *Gatherer) updateAlertsCache(ctx context.Context, metricsClient rest.Interface) error {
-	const logPrefix = "conditional gatherer: "
-	klog.Info(logPrefix + "updating alerts cache for conditional gatherer")
+	klog.Info("updating alerts cache for conditional gatherer")
 
 	g.firingAlerts = make(map[string][]AlertLabels)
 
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -99,7 +99,7 @@ build: ## Compiles the insights operator
 
 .PHONY: build-debug
 build-debug: ## Compiles the insights operator in debug mode
-	go build $(GO_BUILD_FLAGS) -gcflags="all=-N -l" \
+	go build -gcflags="all=-N -l" \
```
tremes, line 102: I know it's not used, but are we safe to remove it? Most of the tests are passing so probably OK to remove.....

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -286,13 +286,16 @@ func (g *Gatherer) updateAlertsCache(ctx context.Context, metricsClient rest.Int
 }
 
 func (g *Gatherer) updateVersionCache(ctx context.Context, configClient configv1client.ConfigV1Interface) error {
+	const logPrefix = "conditional gatherer: "
```
tremes, line 289: This repeats in more methods here AFAIK so perhaps it would be better to create global constant.

pkg/gatherers/conditional/conditional_gatherer.go
```diff
@@ -286,13 +286,16 @@ func (g *Gatherer) updateAlertsCache(ctx context.Context, metricsClient rest.Int
...
+	klog.Info(logPrefix + "updating version cache for conditional gatherer")
+
 	clusterVersion, err := configClient.ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
 
 	g.clusterVersion = clusterVersion.Status.Desired.Version
-
+	klog.Infof(logPrefix+"cluster version is '%v'", g.clusterVersion)
```
tremes, line 298: nit: Using `fmt.Sprintf` would be probably slightly nicer :)

pkg/gatherers/conditional/conditions.go
```diff
@@ -100,14 +100,19 @@ func (g *Gatherer) doesClusterVersionMatch(expectedVersionExpression string) (bo
...
-	expectedRange, err := semver.ParseRange(expectedVersionExpression)
+	shortClusterVersion, err := semver.Parse(g.clusterVersion)
 	if err != nil {
 		return false, err
 	}
 
-	// ignore everything after the first three numbers
-	clusterVersion.Pre = nil
-	clusterVersion.Build = nil
+	// ignore everything after the first three numbers for the short version
```
tremes, line 108: I am not sure I fully understand. Does it mean please that we ignore z-stream version (e.g 4.10.5 simply falls under 4.10?) ? Maybe it's OK (but maybe not :)) but we should cover it in the tests IMO.

Makefile
Sergey1011010: Yup, this variable was removed from the Makefile and it's not used in `build`, so it doesn't make sense to have it in `build-debug`

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: Why? Infof does the same thing under the hood

pkg/gatherers/conditional/conditions.go
Sergey1011010: No, in the version `4.10.5-0.nightly-2022-05-25-193227` we ignore `-0.nightly-2022-05-25-193227` in case the version in the condition doesn't use it, for example `>4.10.5`. If the version in the condition does use it, for example the condition is `>4.10.5-0.nightly-2022-05-25-193227`, then it's used.

Sergey1011010: I'll add more tests

pkg/gatherers/conditional/conditional_gatherer.go
tremes: Ahh yeah OK. So then `klog.Infof("%s cluster version is '%v'",logPrefix, g.clusterVersion)` but it doesn't matter much.....

pkg/gatherers/conditional/conditions.go
tremes: OK great. Thank you

Makefile
tremes: OK. Hopefully there's no other environment using it....

pkg/gatherers/conditional/conditional_gatherer.go
Sergey1011010: I realized that it doesn't make much sense. It should be either all conditional gatherer logs are prefix or none at all and in this particular case we already have file name which is `conditional_gatherer.go` added before the logs.

tremes:
I reviewed the changes, but didn't manually test any conditional gatherer. 
/lgtm

tremes:
This is a minor update (neither a feature nor a bug). I am adding the labels myself. 
/label docs-approved
/label px-approved

PR #635, opened 2022-06-03
Revert "Implement Prometheus Collector pattern (#607)"

neisw:
This reverts commit b916a9c3ab4740a2eb39bbfd72beca1c95efded7.

We believe https://github.com/openshift/insights-operator/pull/607/ broke nightly builds due to a missing entry in image-references similar to https://github.com/openshift/cluster-network-operator/blob/master/manifests/image-references#L17-L20

The fix with the reverted code and updated image-references can be tested by adding

/test e2e-metal-ipi-ovn-ipv6

See https://issues.redhat.com/browse/TRT-282 for more details


tremes:
Revert. I am adding the no FF labels.
/label qe-approved
/label docs-approved
/label px-approved

Prashanth684:
this PR is needed to unblock non x86 architecture releases too

PR #646, opened 2022-06-28
chore(manifests): adding insights capability annotations

rluders:
This PR modifies the manifests by adding the capability annotations to it.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

n/a

## Documentation

n/a

## Unit Tests

n/a

## Privacy

n/a

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-8504

### DIFF
--- a/manifests/06-deployment.yaml
+++ b/manifests/06-deployment.yaml
@@ -8,6 +8,7 @@ metadata:
     exclude.release.openshift.io/internal-openshift-hosted: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: insights
 spec:
   strategy:
     type: Recreate
@@ -18,6 +19,7 @@ spec:
     metadata:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        capability.openshift.io/name: insights
       labels:
         app: insights-operator
     spec:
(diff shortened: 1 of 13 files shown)


manifests/06-deployment.yaml
```diff
@@ -18,6 +19,7 @@ spec:
     metadata:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        capability.openshift.io/name: insights
```
tremes, line 22: Is that required here on the template level please (same above for ibm cloud deployment manifest)? I am looking at https://github.com/openshift/cluster-samples-operator/blob/master/manifests/06-operator.yaml and it doesn't seem so.

rluders: Well, now this is a good question...
https://github.com/honza/cluster-baremetal-operator/blob/8f43aca6a89b151af287fa5532749f93c26fc238/manifests/0000_31_cluster-baremetal-operator_06_deployment.yaml#L21

rluders: But I don't think that it is... TBH

tremes: I don't know. @wking is the annotation required on the template level as well please?

wking: The cluster-version operator is just looking at the top-level `metadata.annotations` in each manifest.  It doesn't drill down into `spec` when trying to decide whether a given YAML resource should be reconciled in the current cluster.  So +1 to dropping this line.

tremes: Thank you.

tremes:
Mianfests update. 
/lgtm

rluders:
This PRs  needs [openshift/api](https://github.com/openshift/api/pull/1212) to be merged first, and also, the CVO openshift/api needs to be bumped, otherwise CI is going to fail.

/hold

rluders:
@wking I guess that I'll need to update this one to PascalCase as well, right?

PR #650, opened 2022-07-07
refactor(workloads_info): reduce cognitive complexity

rluders:

## Categories

- [ ] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

### DIFF
--- a/pkg/gatherers/workloads/workloads_info.go
+++ b/pkg/gatherers/workloads/workloads_info.go
@@ -63,37 +63,63 @@ func (g *Gatherer) GatherWorkloadInfo(ctx context.Context) ([]record.Record, []e
 	return gatherWorkloadInfo(ctx, gatherKubeClient.CoreV1(), gatherOpenShiftClient)
 }
 
-// nolint: funlen, gocyclo, gocritic
 func gatherWorkloadInfo(
 	ctx context.Context,
 	coreClient corev1client.CoreV1Interface,
 	imageClient imageclient.ImageV1Interface,
 ) ([]record.Record, []error) {
-	// load images as we find them
-	imageCh := make(chan string, workloadGatherPageSize)
-	imagesDoneCh := gatherWorkloadImageInfo(ctx, imageClient.Images(), imageCh)
+	imageCh, imagesDoneCh := gatherWorkloadImageInfo(ctx, imageClient.Images())
 
-	// load pods in order
 	start := time.Now()
+	limitReached, info, err := workloadInfo(ctx, coreClient, imageCh)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	workloadImageResize(info.PodCount)
+
+	records := []record.Record{
+		{
+			Name: "config/workload_info",
+			Item: record.JSONMarshaller{Object: &info},
+		},
+	}
+
+	// wait for as many images as we can find to load
+	handleWorkloadImageInfo(ctx, &info, start, imagesDoneCh)
+
+	if limitReached {
+		return records, []error{fmt.Errorf("the %d limit for number of pods gathered was reached", podsLimit)}
+	}
+	return records, nil
+}
+
+// nolint: funlen, gocritic, gocyclo
+func workloadInfo(
+	ctx context.Context,
+	coreClient corev1client.CoreV1Interface,
+	imageCh chan string,
+) (bool, workloadPods, error) {
+	defer close(imageCh)
 	limitReached := false
 
 	var info workloadPods
-	var namespace string
-	var namespaceHash string
+	var namespace, namespaceHash, continueValue string
 	var namespacePods workloadNamespacePods
 	h := sha256.New()
 
 	// Use the Limit and Continue fields to request the pod information in chunks.
-	var continueValue string
 	for {
 		pods, err := coreClient.Pods("").List(ctx, metav1.ListOptions{
 			Limit:    workloadGatherPageSize,
 			Continue: continueValue,
 		})
 		if err != nil {
-			return nil, []error{err}
+			return false, workloadPods{}, err
 		}
-		for _, pod := range pods.Items {
+
+		for podIdx := range pods.Items {
+			pod := pods.Items[podIdx]
 			// initialize the running state, including the namespace hash
 			if pod.Namespace != namespace {
 				if len(namespace) != 0 {
@@ -117,45 +143,35 @@ func gatherWorkloadInfo(
 			namespacePods.Count++
 
 			switch {
-			case pod.Status.Phase == corev1.PodSucceeded, pod.Status.Phase == corev1.PodFailed:
+			case isPodTerminated(&pod):
 				// track terminal pods but do not report their data
 				namespacePods.TerminalCount++
 				continue
-			case pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodPending,
-				len(pod.Status.InitContainerStatuses) != len(pod.Spec.InitContainers),
-				len(pod.Status.ContainerStatuses) != len(pod.Spec.Containers):
+			case podCanBeIgnored(&pod):
 				// consider pods that are in a known state
 				// or pods without filled out status are invalid
 				namespacePods.IgnoredCount++
 				continue
 			}
 
-			var podShape workloadPodShape
-			var ok bool
-			podShape.InitContainers, ok = calculateWorkloadContainerShapes(h, pod.Spec.InitContainers, pod.Status.InitContainerStatuses)
+			podShape, ok := calculatePodShape(h, &pod)
 			if !ok {
 				namespacePods.InvalidCount++
 				continue
 			}
-			podShape.Containers, ok = calculateWorkloadContainerShapes(h, pod.Spec.Containers, pod.Status.ContainerStatuses)
-			if !ok {
-				namespacePods.InvalidCount++
-				continue
-			}
-
-			podShape.RestartsAlways = pod.Spec.RestartPolicy == corev1.RestartPolicyAlways
 
 			if index := workloadPodShapeIndex(namespacePods.Shapes, podShape); index != -1 {
 				namespacePods.Shapes[index].Duplicates++
-			} else {
-				namespacePods.Shapes = append(namespacePods.Shapes, podShape)
+				continue
+			}
+			namespacePods.Shapes = append(namespacePods.Shapes, podShape)
 
-				for _, container := range podShape.InitContainers {
-					imageCh <- container.ImageID
-				}
-				for _, container := range podShape.Containers {
-					imageCh <- container.ImageID
-				}
+			for _, container := range podShape.InitContainers {
+				imageCh <- container.ImageID
+			}
+
+			for _, container := range podShape.Containers {
+				imageCh <- container.ImageID
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/gatherers/workloads/workloads_info.go
```diff
@@ -299,6 +320,44 @@ func gatherWorkloadImageInfo(
 	return imagesDoneCh
 }
 
+// imagesFromChannel drains the channel of any image IDs
+func imagesFromChannel(pendingIDs map[string]struct{}, images map[string]workloadImage, id string, imageCh <-chan string) {
```
tremes, line 324: Perhaps `readmageNames` or `readImageSHAs` .

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -299,6 +320,44 @@ func gatherWorkloadImageInfo(
...
+	for k := range pendingIDs {
+		delete(pendingIDs, k)
+	}
```
tremes, line 327: This looks a little bit weird at the first glance. Do we need to pass the map as the parameter when it's completely deleted then?

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -174,24 +191,56 @@ func gatherWorkloadInfo(
 		info.Namespaces[namespaceHash] = namespacePods
 		info.PodCount += namespacePods.Count
 	}
+	return limitReached, info, imageCh, nil
+}
 
-	workloadImageResize(info.PodCount)
+func podIsTerminal(pod *corev1.Pod) bool {
```
tremes, line 197: Perhaps `podIsRunning` or `isPodRunning` or maybe only `isRunning`.

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -174,24 +191,56 @@ func gatherWorkloadInfo(
...
+	return pod.Status.Phase == corev1.PodSucceeded ||
+		pod.Status.Phase == corev1.PodFailed
+}
 
-	records := []record.Record{
-		{
-			Name: "config/workload_info",
-			Item: record.JSONMarshaller{Object: &info},
-		},
+func podCanBeIgnored(pod *corev1.Pod) bool {
```
tremes, line 202: Some similar name suggestions as above and this would definitely deserve some explaining comment, because I don't know when a pod can be ignored.

pkg/gatherers/workloads/workloads_info.go
rluders: Nice catch. I overlooked that.

rluders:
I would like to add some unit tests for it, but I'll address it in a separate task. The current unit tests already test the expected result.

tremes:
This breaks the gatherer at the moment. Simple test is to run it before the update and after the update and you should notice that after the update there's inconsistency in `imageCount` data as well as in `images` data.

pkg/gatherers/workloads/workloads_info.go
tremes: This conversation doesn't seem to be resolved.

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -175,23 +195,56 @@ func gatherWorkloadInfo(
...
+		pod.Status.Phase == corev1.PodFailed
+}
+
+func podCanBeIgnored(pod *corev1.Pod) bool {
+	return pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodPending ||
+		len(pod.Status.InitContainerStatuses) != len(pod.Spec.InitContainers) ||
+		len(pod.Status.ContainerStatuses) != len(pod.Spec.Containers)
+}
+
+func handlePodShape(h hash.Hash, pod *corev1.Pod) (workloadPodShape, bool) {
```
tremes, line 212: Just name suggestion - would be more suitable to call it `calculatePodShape`? Since we have `calculateWorkloadContainerShapes`, `calculateWorkloadInfo`

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -75,25 +75,54 @@ func gatherWorkloadInfo(
...
+	// wait for as many images as we can find to load
+	handleWorkloadImageInfo(ctx, &info, start, imageCh, imagesDoneCh)
+
+	if limitReached {
+		return records, []error{fmt.Errorf("the %d limit for number of pods gathered was reached", podsLimit)}
+	}
+	return records, nil
+}
+
+// nolint: funlen, gocritic, gocyclo
```
tremes, line 101: I thought the goal was to get rid of the gocyclo - i.e reduce the complexity :)

tremes:
It seems we don't need to ignore the:

- https://github.com/openshift/insights-operator/pull/650/files#diff-5c6d4b139523747033ba70556af76c97a5dd864579dd8b3a04778a9cdb3a282cR66
- https://github.com/openshift/insights-operator/pull/650/files#diff-5c6d4b139523747033ba70556af76c97a5dd864579dd8b3a04778a9cdb3a282cL217

anymore.

pkg/gatherers/workloads/workloads_info.go
rluders: That was the main goal... I reduced it a little bit, the score is better now but... not perfect. Trust me it is better than before and it is almost impossible to remove the remaining complexity without rewrite it completely.

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -175,23 +194,56 @@ func gatherWorkloadInfo(
...
-	workloadImageResize(info.PodCount)
+	return limitReached, info, imageCh, nil
+}
 
-	records := []record.Record{
-		{
-			Name: "config/workload_info",
-			Item: record.JSONMarshaller{Object: &info},
-		},
+func isPodRunning(pod *corev1.Pod) bool {
```
tremes, line 200: I just realized that this name doesn't make sense. It's actually the opposite right? When a pod is in the `PodSucceeded` or `PodFailed` state then it's definitely not running. Sorry for the confusion. I would call the function `podTerminated` or `isPodTerminated`.

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -75,25 +74,54 @@ func gatherWorkloadInfo(
 
 	// load pods in order
 	start := time.Now()
+	limitReached, info, imageCh, err := workloadInfo(ctx, coreClient, imageCh)
```
tremes, line 77: There's no need to return the `imageCh` here.

pkg/gatherers/workloads/workloads_info.go
rluders: `isPodTerminated` seems good.

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -175,23 +194,56 @@ func gatherWorkloadInfo(
...
+	podShape.RestartsAlways = pod.Spec.RestartPolicy == corev1.RestartPolicyAlways
+
+	return podShape, true
+}
+
+func handleWorkloadImageInfo(
+	ctx context.Context,
+	info *workloadPods,
+	start time.Time,
+	imageCh chan string,
```
tremes, line 233: This `imageCh` parameter is sent here only to be closed. It looks a little bit weird IMO. I guess we could close the channel before the function call.

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -75,25 +74,56 @@ func gatherWorkloadInfo(
 
 	// load pods in order
 	start := time.Now()
+	limitReached, info, err := workloadInfo(ctx, coreClient, imageCh)
+	if err != nil {
+		return []record.Record{}, []error{err}
```
tremes, line 79: Just nit: I think it should be OK to return `nil, []error{err}` as previously

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -75,25 +74,56 @@ func gatherWorkloadInfo(
...
+	workloadImageResize(info.PodCount)
+
+	records := []record.Record{
+		{
+			Name: "config/workload_info",
+			Item: record.JSONMarshaller{Object: &info},
+		},
+	}
+
+	close(imageCh)
```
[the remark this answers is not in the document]
ncaak: I agree, there are a function call before the close statement. It seems good catching expected errors but using defer costs almost nothing and it could safe some future errors.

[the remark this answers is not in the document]
rluders: Well, if I try to add `defer` to close the `imageCh` it just doesn't give me the `imageCount` or the `images` at the report.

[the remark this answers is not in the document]
tremes: Ohh yes it seems too late, because then there's no time (or a way) to read from the `imagesDoneCh` channel. Perhaps we could close this channel somewhere else.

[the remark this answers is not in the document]
rluders: Let me see what I can do.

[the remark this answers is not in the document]
rluders: @tremes @ncaak check it now... I am still not happy with it. I tried to defer and create both channels in several places but it was causing panic or it was not generating the correct report data.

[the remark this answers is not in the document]
tremes: Why are you not happy with that? I think it's OK

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -63,37 +63,65 @@ func (g *Gatherer) GatherWorkloadInfo(ctx context.Context) ([]record.Record, []e
...
 	imageClient imageclient.ImageV1Interface,
 ) ([]record.Record, []error) {
 	// load images as we find them
-	imageCh := make(chan string, workloadGatherPageSize)
-	imagesDoneCh := gatherWorkloadImageInfo(ctx, imageClient.Images(), imageCh)
+	imageCh, imagesDoneCh := gatherWorkloadImageInfo(ctx, imageClient.Images())
 
 	// load pods in order
 	start := time.Now()
+	limitReached, info, err := workloadInfo(ctx, coreClient, imageCh)
```
tremes, line 76: The two comments above:
- `// load pods in order`
- `// load images as we find them`
seems to be at wrong place (as they were probably previously) so I would either move them or remove them.

pkg/gatherers/workloads/workloads_info.go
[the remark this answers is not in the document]
rluders: I would like to close the channel where it is created, but it doesn't seem possible. I mean, I could be wrong here but, maybe adding some `WaitGroup` would help, but it will require some even bigger refactoring.

tremes:
I found one more thing missing (when comparing to original code) and that is https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/workloads/workloads_info.go#L254-L256. It seems like minor but useful "optimization", because without this "draining" it seems we would [iterate](https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/workloads/workloads_info.go#L269) over some IDs more than 1.

tremes:
I tested and reviewed several times. Hopefully I didn't miss anything.
/lgtm

tremes:
This is basically refactoring.
/label docs-approved
/label px-approved

PR #657, opened 2022-08-16
helm upgrade and uninstall metric gathering

Kartikey-star:
Signed-off-by: Kartikey Mamgain <mamgainkartikey@gmail.com>
We would like to add "console_helm_uninstalls_total" and "console_helm_upgrades_total"metric created at the time of upgrading and uninstallling a chart from Openshift Developer Console gathered by Insights Operator so that we can take decisions from that data. In order to achieve this i have modified gatherMostRecentMetrics method and updated the docs.

In order to test and check the metrics is getting gathered by the operator you can follow the given steps:

1. Connect to a cluster remotely
2. Install a Chart using Openshift Developer Console.For example let's say redis-sed from openshift-helm-charts.
3. Once the chart is installed you can upgrade the chart, update the hostname to any arbitory value.(Performing this action actually increases the metric value for console_helm_upgrades_total)
4. After doing this you can uninstall the chart.(Performing this action actually increases the metric value for console_helm_uninstalls_total)
5. You can run the operator locally and should be able to check the data gathered in config/metrics.

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting


pkg/gatherers/clusterconfig/recent_metrics.go
```diff
@@ -46,6 +48,8 @@ const (
 //   - "virt_platform metric": 4.6.34+, 4.7.16+, 4.8+
 //   - "vsphere_node_hw_version_total": 4.7.11+, 4.8+
 //   - "console_helm_installs_total": 4.11+
+//   - "console_helm_upgrades_total": 4.11+
+//   - "console_helm_uninstalls_total": 4.11+
```
tremes, line 52: 4.11 is out already.

Kartikey-star: yes, made the change.

tremes:
Thanks @Kartikey-star. It looks good. Can you please add a new sample of these metrics (we don't have the `console_helm_installs_total` either) to the https://github.com/openshift/insights-operator/blob/master/docs/insights-archive-sample/config/metrics?

tremes:
This is simple extension. Thank you.
/lgtm

PR #658, opened 2022-08-16
openshift-machine-api warning events gatherer

Megonasayit:
This PR implements new gatherer for gathering "non-standard" or "abnormal" events from the "openshift-machine-api" namespace.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/events/openshift-machine-api.json`

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gatherers/clusterconfig/openshift_machine_api_events_test.go`
- `pkg/gatherers/clusterconfig/events_filtering_test.go`

## Privacy

There are no sensitive data in the newly collected information.

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-8389


pkg/gatherers/clusterconfig/openshift_machine_api_events.go
```diff
@@ -0,0 +1,78 @@
...
+	records, err := gatherOpenshiftMachineApiEvents(ctx, gatherKubeClient.CoreV1(), namespace, g.interval)
+	if err != nil {
+		return nil, []error{err}
+	}
+	return records, nil
+}
+
+func gatherOpenshiftMachineApiEvents(ctx context.Context,
+	coreClient corev1client.CoreV1Interface,
+	namespace string,
```
tremes, line 34: You called the function `gatherOpenshiftMachineApiEvents` so no need to pass the `namespace` name as parameter. I think you can simply hardcoded here `coreClient.Events("openshift-machine-api").List(ctx, metav1.ListOptions{})`

pkg/gatherers/clusterconfig/openshift_machine_api_events.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+		if events.Items[index].LastTimestamp.Time.IsZero() {
+			compactedEvents.Items[i].LastTimestamp = events.Items[index].Series.LastObservedTime.Time
+		}
+	}
+	sort.Slice(compactedEvents.Items, func(i, j int) bool {
+		return compactedEvents.Items[i].LastTimestamp.Before(compactedEvents.Items[j].LastTimestamp)
+	})
+
+	return []record.Record{{Name: fmt.Sprintf("config/events/%s", namespace), Item: record.JSONMarshaller{Object: &compactedEvents}}}, nil
```
tremes, line 77: As mentioned on the Slacl - this should go into `events/%s` please.

pkg/gatherers/clusterconfig/openshift_machine_api_events.go
Megonasayit: Done.

pkg/gatherers/clusterconfig/openshift_machine_api_events.go
Megonasayit: Done.

pkg/gatherers/clusterconfig/events_filtering.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+
+	return filteredEvents
+}
+
+//eventListToCompactedEventList() coverts EventList() into CompactedEventList()
+func eventListToCompactedEventList(events v1.EventList) CompactedEventList {
+	compactedEvents := CompactedEventList{Items: make([]CompactedEvent, len(events.Items))}
+	for i := 0; i < len(events.Items); i++ {
```
tremes, line 56: I think we can do `for i := range events.Items` right?

pkg/gatherers/clusterconfig/openshift_machine_api_events.go
```diff
@@ -0,0 +1,44 @@
...
+	interval time.Duration) ([]record.Record, error) {
+	events, err := coreClient.Events("openshift-machine-api").List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	// filter the event list to only recent events with type different than "Normal"
+	filteredEvents := filterEvents(interval, events, "Warning")
+	compactedEvents := eventListToCompactedEventList(filteredEvents)
+
+	return []record.Record{{Name: "events/openshift-machine-api", Item: record.JSONMarshaller{Object: &compactedEvents}}}, nil
```
tremes, line 43: I think this is not correct. This will produce only one record with the name `openshift-machine-api`.

pkg/gatherers/clusterconfig/events_filtering_test.go
```diff
@@ -0,0 +1,77 @@
...
+	normalEvent := v1.Event{
+		ObjectMeta:    metav1.ObjectMeta{Name: "normalEvent"},
+		LastTimestamp: metav1.Now(),
+		Type:          "Normal",
+	}
+	warningEvent := v1.Event{
+		ObjectMeta:    metav1.ObjectMeta{Name: "warningEvent"},
+		LastTimestamp: metav1.Now(),
+		Type:          "Warning",
+	}
```
tremes, line 44: This test and the one above could be merged into one (See e.g [this](https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/clusterconfig/namespaces_with_overlapping_uids_test.go#L15) for how you can define multiple similar tests.
I would also recommend to test more than one event with the "abnormal" event type :)

pkg/gatherers/clusterconfig/openshift_machine_api_events.go
Megonasayit: It will produce one file with name `openshift-machine-api` but all gathered events will be written on it.

pkg/gatherers/clusterconfig/events_filtering.go
```diff
@@ -0,0 +1,72 @@
...
+	"sort"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+//filterEvents() returns events that occoured since last interval
+func filterEvents(interval time.Duration, events *v1.EventList, Type string) v1.EventList {
+	oldestEventTime := time.Now().Add(-interval)
+	var filteredEvents = v1.EventList{}
```
tremes, line 13: You can do `var filteredEvents v1.EventList`

pkg/gatherers/clusterconfig/events_filtering.go
```diff
@@ -0,0 +1,72 @@
...
+			if isEventAbnormal(&events.Items[i]) && isEventNew(&events.Items[i], oldestEventTime) {
+				filteredEvents.Items = append(filteredEvents.Items, events.Items[i])
+			}
+		}
+	default:
+		for i := range events.Items {
+			if isEventNew(&events.Items[i], oldestEventTime) {
+				filteredEvents.Items = append(filteredEvents.Items, events.Items[i])
+			}
+		}
```
tremes, line 26: I still don't like this switch much to be honest :) What if someone passes "Error" type for example? Perhaps I would split this function into two:
- `getEventsForInterval` (or from interval? Naming is difficult:-)) - this would handle just the time "filtering"
- `filterAbnormalEvents` or something like that and this function (or logic) can stay in the new gatherer for now

pkg/gatherers/clusterconfig/events_filtering.go
```diff
@@ -0,0 +1,72 @@
...
+
+func isEventAbnormal(event *v1.Event) bool {
+	return event.Type != "Normal"
+}
+
+//eventListToCompactedEventList() coverts EventList() into CompactedEventList()
+func eventListToCompactedEventList(events *v1.EventList) CompactedEventList {
+	compactedEvents := CompactedEventList{Items: make([]CompactedEvent, len(events.Items))}
+	for i := range events.Items {
+		compactedEvents.Items[i] = CompactedEvent{
```
tremes, line 54: Please extract the `compactedEvents.Items[i]` and the `events.Items[i]` into new variable here. I believe this will make the code more readable.

pkg/gatherers/clusterconfig/events_filtering_test.go
```diff
@@ -0,0 +1,187 @@
...
+				ObjectMeta:    metav1.ObjectMeta{Name: "newEvent"},
+				LastTimestamp: metav1.Now(),
+				Type:          "Normal",
+			},
+			expected: true,
+		},
+		{
+			event: v1.Event{
+				ObjectMeta:    metav1.ObjectMeta{Name: "oldEvent"},
+				LastTimestamp: metav1.NewTime(time.Now().Add(-6 * time.Minute)),
```
tremes, line 152: You should be able to do `metav1.NewTime(time.Now().Sub(6 * time.Minute))`

tremes: Hmm maybe not....since it returns duration

pkg/gatherers/clusterconfig/events_filtering.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"sort"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// filterEvents() returns events that occoured since last interval
+func getEventsForInterval(interval time.Duration, events *v1.EventList) v1.EventList {
```
tremes, line 11: nit: Please fix the comment

pkg/gatherers/clusterconfig/events_filtering.go
```diff
@@ -0,0 +1,76 @@
...
+	for i := range events.Items {
+		if isEventNew(&events.Items[i], oldestEventTime) {
+			filteredEvents.Items = append(filteredEvents.Items, events.Items[i])
+		}
+	}
+	return filteredEvents
+}
+
+// if LastTimestamp is zero then try to check the event series
+func isEventNew(event *v1.Event, oldestEventTime time.Time) bool {
```
tremes, line 23: The comment here would deserve an update as well :) This comment is more suitable to the inner code.

pkg/gatherers/clusterconfig/events_filtering.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+	return filteredEvents
+}
+
+func isEventAbnormal(event *v1.Event) bool {
+	return event.Type != "Normal"
+}
+
+// eventListToCompactedEventList() coverts EventList() into CompactedEventList()
```
tremes, line 51: nit: Typo in the comment `coverts`

tremes: Maybe better to say `converts v1.EventList into CompactedEventList` - the brackets seems not necessary when we talk about types

pkg/gatherers/clusterconfig/openshift_machine_api_events_test.go
```diff
@@ -0,0 +1,87 @@
...
+		ctx        context.Context
+		coreClient corev1client.CoreV1Interface
+	}
+	test := struct {
+		name    string
+		args    args
+		wantErr bool
+		want    []record.Record
+	}{
+		name: "empty openshift-machine-api events",
```
tremes, line 67: It's not an empty event test anymore right :)

pkg/gatherers/clusterconfig/openshift_machine_api_events_test.go
```diff
@@ -0,0 +1,87 @@
...
+		args: args{
+			ctx: context.TODO(),
+			coreClient: kubefake.NewSimpleClientset(warningEvent1.DeepCopy(), normalEvent1.DeepCopy(),
```
tremes, line 70: this is interesting usage of the client (not sure if I already saw it). No need to call `DeepCopy` on every object. You can pass pointer.

pkg/gatherers/clusterconfig/openshift_machine_api_events_test.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+		wantErr: false,
+		want:    []record.Record{{Name: "events/openshift-machine-api", Item: record.JSONMarshaller{Object: &compactedEvents}}},
+	}
+
+	t.Run(test.name, func(t *testing.T) {
+		t.Parallel()
+		got, err := gatherOpenshiftMachineAPIEvents(test.args.ctx, test.args.coreClient, 1*time.Minute)
+		if (err != nil) != test.wantErr {
+			t.Errorf("gatherOpenshiftMachineApiEvents() error = %v, wantErr %v", err, test.wantErr)
```
tremes, line 82: you can use `assert` for this as well (something like `assert.NoErrorr`)

pkg/gatherers/clusterconfig/events_filtering_test.go
```diff
@@ -0,0 +1,170 @@
...
+					ObjectMeta:    metav1.ObjectMeta{Name: "newEvent3"},
+					LastTimestamp: metav1.NewTime(timeNow),
+				},
+			},
+		},
+	}
+
+	filteredEvents := getEventsForInterval(1*time.Minute, &test.events)
+	if !reflect.DeepEqual(filteredEvents, test.expected) {
+		t.Errorf("filterEvents() = %v, want %v", filteredEvents, test.expected)
```
tremes, line 62: Same with the `assert` here. See below

tremes:
I reviewed it several times. God job! Thank you.
/lgtm

PR #664, opened 2022-08-23
OCPBUGS-439: update the DVO metrics gatherer

tremes:
The DVO metrics gatherer relied on the `deployment-validation-operator` namespace name, which is not very convenient, because it can be installed elsewhere.

Steps to verify:
1. Install the `deployment-validation-operator` through the OperatorHub (it will be installed in the `openshift-operators` namespace).
2. Create manually the metrics service for the DVO. You can use [this definition](https://github.com/app-sre/deployment-validation-operator/blob/330a67b6c64462673343c8a0d4f30fe91c5ecac9/deploy/openshift/service.yaml). This should be fixed with the https://issues.redhat.com/browse/DVO-60
3. Create some new namespace and deploy something (e.g the default httpd example) without the resource limits specified. This is required to introduce some DVO checks to be reported. 
4. Run IO gathering and check that the DVO metrics are gathered.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

No new data

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/OCPBUGS-439

### DIFF
--- a/docs/conditional-gatherer/README.md
+++ b/docs/conditional-gatherer/README.md
@@ -94,7 +94,7 @@ echo '{
 4. Wait for the alert to fire:
 ```bash
 export PROMETHEUS_HOST=$(oc get route -n openshift-monitoring prometheus-k8s -o jsonpath='{@.spec.host}')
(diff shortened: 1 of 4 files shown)


rluders (approved):
Reviewed the code and tested it.
/lgtm

PR #675, opened 2022-09-09
OCPBUGS-13930: feat(containers_logs): allows filter by specific container

rluders:
This PR enables `containers_logs` conditional gather to collect logs only from the given container and specific pods.

## Categories

- [ ] Bugfix
- [X] Enhancement
- [ ] Backporting

## Sample Archive

n/a

## Documentation

n/a

## Unit Tests

n/a

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-8768

### DIFF
--- a/pkg/gatherers/common/gather_logs.go
+++ b/pkg/gatherers/common/gather_logs.go
@@ -79,14 +79,22 @@ func CollectLogsFromContainers( //nolint:gocyclo
 
 	var skippedContainers int
 	var records []record.Record
+	var regexErrorMessage = "fails to compile regex '%s' to filter '%s'"
+
+	podNameRegex, err := regexp.Compile(resourceFilter.PodNameRegexFilter)
+	if err != nil {
+		klog.Warningf(regexErrorMessage, resourceFilter.PodNameRegexFilter, "pod name")
+		return nil, err
+	}
+	containerNameRegex, err := regexp.Compile(resourceFilter.ContainerNameRegexFilter)
+	if err != nil {
+		klog.Warningf(regexErrorMessage, resourceFilter.ContainerNameRegexFilter, "container")
+		return nil, err
+	}
 
 	for i := range pods.Items {
 		if len(resourceFilter.PodNameRegexFilter) > 0 {
-			match, err := regexp.MatchString(resourceFilter.PodNameRegexFilter, pods.Items[i].Name)
-			if err != nil {
-				return nil, err
-			}
-			if !match {
+			if !podNameRegex.MatchString(pods.Items[i].Name) {
 				continue
 			}
 		}
@@ -109,11 +117,7 @@ func CollectLogsFromContainers( //nolint:gocyclo
 
 		for _, containerName := range containerNames {
 			if len(resourceFilter.ContainerNameRegexFilter) > 0 {
-				match, err := regexp.MatchString(resourceFilter.ContainerNameRegexFilter, containerName)
-				if err != nil {
-					return nil, err
-				}
-				if !match {
+				if !containerNameRegex.MatchString(containerName) {
 					continue
 				}
 			}
--- a/pkg/gatherers/common/gather_logs.go
+++ b/pkg/gatherers/common/gather_logs.go
@@ -59,8 +59,8 @@ type LogMessagesFilter struct {
 func CollectLogsFromContainers( //nolint:gocyclo
 	ctx context.Context,
 	coreClient v1.CoreV1Interface,
-	containersFilter LogContainersFilter,
-	messagesFilter LogMessagesFilter,
+	containersFilter *LogContainersFilter,
+	messagesFilter *LogMessagesFilter,
 	buildLogFileName func(namespace string, podName string, containerName string) string,
 ) ([]record.Record, error) {
 	if buildLogFileName == nil {
@@ -212,7 +212,7 @@ func FilterLogFromScanner(scanner *bufio.Scanner, messagesToSearch []string, reg
 	return strings.Join(result, "\n"), nil
 }
 
-func podLogOptions(containerName string, messagesFilter LogMessagesFilter) *corev1.PodLogOptions {
+func podLogOptions(containerName string, messagesFilter *LogMessagesFilter) *corev1.PodLogOptions {
 	sinceSeconds := &messagesFilter.SinceSeconds
 	if messagesFilter.SinceSeconds == 0 {
 		sinceSeconds = nil
--- a/pkg/gatherers/common/gather_logs.go
+++ b/pkg/gatherers/common/gather_logs.go
@@ -17,8 +17,8 @@ import (
 	"github.com/openshift/insights-operator/pkg/utils/marshal"
 )
 
-// LogContainersFilter allows you to filter containers
-type LogContainersFilter struct {
+// LogResourceFilter allows you to filter containers
+type LogResourceFilter struct {
 	Namespace                string
 	LabelSelector            string
 	FieldSelector            string
@@ -59,7 +59,7 @@ type LogMessagesFilter struct {
 func CollectLogsFromContainers( //nolint:gocyclo
 	ctx context.Context,
 	coreClient v1.CoreV1Interface,
-	containersFilter *LogContainersFilter,
+	resourceFilter *LogResourceFilter,
 	messagesFilter *LogMessagesFilter,
 	buildLogFileName func(namespace string, podName string, containerName string) string,
 ) ([]record.Record, error) {
@@ -69,9 +69,9 @@ func CollectLogsFromContainers( //nolint:gocyclo
 		}
 	}
 
-	pods, err := coreClient.Pods(containersFilter.Namespace).List(ctx, metav1.ListOptions{
-		LabelSelector: containersFilter.LabelSelector,
-		FieldSelector: containersFilter.FieldSelector,
+	pods, err := coreClient.Pods(resourceFilter.Namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: resourceFilter.LabelSelector,
+		FieldSelector: resourceFilter.FieldSelector,
 	})
 	if err != nil {
 		return nil, err
@@ -81,8 +81,8 @@ func CollectLogsFromContainers( //nolint:gocyclo
 	var records []record.Record
 
 	for i := range pods.Items {
-		if len(containersFilter.PodNameRegexFilter) > 0 {
-			match, err := regexp.MatchString(containersFilter.PodNameRegexFilter, pods.Items[i].Name)
+		if len(resourceFilter.PodNameRegexFilter) > 0 {
+			match, err := regexp.MatchString(resourceFilter.PodNameRegexFilter, pods.Items[i].Name)
 			if err != nil {
 				return nil, err
 			}
@@ -99,7 +99,7 @@ func CollectLogsFromContainers( //nolint:gocyclo
 			containerNames = append(containerNames, pods.Items[i].Spec.InitContainers[j].Name)
 		}
 
-		containersLimited := containersFilter.MaxNamespaceContainers > 0 && len(records) >= containersFilter.MaxNamespaceContainers
+		containersLimited := resourceFilter.MaxNamespaceContainers > 0 && len(records) >= resourceFilter.MaxNamespaceContainers
 		if containersLimited {
 			skippedContainers += len(containerNames)
 			continue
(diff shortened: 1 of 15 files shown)


pkg/gatherers/conditional/gathering_functions.go
```diff
@@ -78,6 +78,8 @@ type GatherLogsOfNamespaceParams struct {
 	Namespace string `json:"namespace"`
 	// A number of log lines to keep for each container
 	TailLines int64 `json:"tail_lines"`
+	// Container name to specify the containers to get log from
+	Container string `json:"container,omitempty"`
```
tremes, line 82: Is it one container name or would it make sense to have more?

rluders: It supports Regex. So, we are good with just one.

pkg/gatherers/conditional/gather_containers_logs.go
```diff
@@ -57,37 +63,29 @@ func (g *Gatherer) gatherContainersLogs(
...
-			FieldSelector: fmt.Sprintf("metadata.name=%s", podName),
+			Namespace: info.namespace,
 		}
 
-		// The container label may not be present for all alerts (e.g., KubePodNotReady).
-		if len(podContainer) > 0 {
-			logContainersFilter.ContainerNameRegexFilter = fmt.Sprintf("^%s$", podContainer)
+		if len(params.Container) > 0 {
+			logContainersFilter.ContainerNameRegexFilter = fmt.Sprintf("^%s$", params.Container)
+		} else if len(info.container) > 0 {
```
rluders, line 81: `KubePod*` conditions does not have the container.

pkg/gatherers/conditional/gather_containers_logs.go
```diff
@@ -57,37 +63,29 @@ func (g *Gatherer) gatherContainersLogs(
...
+			logContainersFilter.ContainerNameRegexFilter = fmt.Sprintf("^%s$", info.container)
+		}
+
+		if len(params.PodName) > 0 {
+			logContainersFilter.PodNameRegexFilter = fmt.Sprintf("^%s$", params.PodName)
+		} else {
+			logContainersFilter.FieldSelector = fmt.Sprintf("metadata.name=%s", info.name)
```
tremes, line 88: I would set the "defaults" in the `logContainersFilter` and remove the else here and above.

rluders: We actually can't do this. There are somethings to be considered.

* For pod name:

If we make `FieldSelector` default, we will have to remove it inside the if, 'cause otherwise we are going to request to the API just the pod that is firing the alert, or saying that we want `metadata.name=`.

* For container:

Check this: https://github.com/openshift/insights-operator/pull/675#discussion_r970870885

tremes: OK I see. Good point.

pkg/gatherers/conditional/gather_containers_logs.go
tremes: +1 for adding some comment or explanation for this condition, because it looks quite cryptic at the first glance

pkg/gatherers/common/gather_logs.go
```diff
@@ -22,6 +22,7 @@ type LogContainersFilter struct {
 	Namespace                string
 	LabelSelector            string
 	FieldSelector            string
+	PodNameRegexFilter       string
```
tremes, line 25: Now when I am looking at this again, it seems weird to define `PodNameRegexFilter` in type called `LogContainersFilter`. Pod is something different than a container. Maybe we should re-think.

pkg/gatherers/common/gather_logs.go
```diff
@@ -80,6 +81,16 @@ func CollectLogsFromContainers( //nolint:gocyclo
 	var records []record.Record
 
 	for i := range pods.Items {
+		if len(containersFilter.PodNameRegexFilter) > 0 {
+			match, err := regexp.MatchString(containersFilter.PodNameRegexFilter, pods.Items[i].Name)
```
tremes, line 85: It would be better to compile the regex outside of the loop.

pkg/gatherers/common/gather_logs.go
tremes: @rluders can you remind me why we can't use the `LabelSelector` here please?

rluders: The rule:

```json
"containers_logs": {
      "alert_name": "CephClusterWarningState",
      "pod_name": "rook-ceph-osd-.*",
      "container": "osd",
      "tail_lines": 30
 }
 ```
 
If I remember correctly, the pod that triggers the alert isn't the one that we want to grab logs from, instead, we want to collect the logs from `osd` containers specifically from `rook-ceph-osd-*` pods.

Comment (rluders) [reply pkg/gatherers/common/gather_logs.go:25]: BTW, here: https://github.com/openshift/insights-operator/pull/675#discussion_r971634618

File: pkg/gatherers/common/gather_logs.go (hunk 5 above)
Comment (rluders) [reply pkg/gatherers/common/gather_logs.go:85]: I would love to, but I can't, since it depends on `pods.Items[i].Name`.

File: pkg/gatherers/common/gather_logs.go (hunk 4 above)
Comment (tremes) [reply pkg/gatherers/common/gather_logs.go:25]: Thanks. Maybe I don't fully remember it, but this seems little suspicious to me. It would be good to share an example of the alert with all the labels and also to reproduce it in some cluster.

Comment (rluders) [reply pkg/gatherers/common/gather_logs.go:25]:
```json
{
  "metric": {
    "__name__": "ALERTS",
    "alertname": "CephClusterWarningState",
    "alertstate": "firing",
    "container": "mgr",
    "endpoint": "http-metrics",
    "instance": "0.0.0.0:9283",
    "job": "rook-ceph-mgr",
    "managedBy": "ocs-storagecluster",
    "namespace": "openshift-storage",
    "pod": "rook-ceph-mgr-a-57dc547555-kvsc7",
    "service": "rook-ceph-mgr",
    "severity": "warning"
  },
  "value": [
    1674040588.66,
    "1"
  ]
}
```

pkg/gatherers/conditional/gather_containers_logs.go
```diff
@@ -57,44 +63,38 @@ func (g *Gatherer) gatherContainersLogs(
...
 
-		logContainersFilter := common.LogContainersFilter{
-			Namespace:     podNamespace,
-			FieldSelector: fmt.Sprintf("metadata.name=%s", podName),
+		if len(params.Container) > 0 {
+			logContainersFilter.ContainerNameRegexFilter = fmt.Sprintf("^%s$", params.Container)
+		} else if len(info.container) > 0 {
+			// KubePod* conditions does not have the container so, we need to verify it first,
+			// and then we can apply it to the creation of the regex rule.
+			logContainersFilter.ContainerNameRegexFilter = fmt.Sprintf("^%s$", info.container)
```
tremes, line 84: I am afraid that this will lead to gathering of logs from the whole namespace for `KubePod*` alerts, because the `ContainerNameRegexFilter` as well as `PodNameRegexFilter` will not be set in this case. Can you please verify it?

rluders: Just to keep history from our discussion:

This will not happen because of the alert. KubePod* will always have the PodName on its Alert given the nature of the alert itself.

```json
{
  "metric": {
    "__name__": "ALERTS",
    "alertname": "KubePodNotReady",
    "alertstate": "firing",
    "namespace": "openshift-storage",
    "pod": "noobaa-core-0",
    "severity": "warning"
  },
  "value": [
    1674040588.66,
    "1"
  ]
}
{
  "metric": {
    "__name__": "ALERTS",
    "alertname": "KubePodNotReady",
    "alertstate": "firing",
    "namespace": "openshift-storage",
    "pod": "noobaa-db-pg-0",
    "severity": "warning"
  },
  "value": [
    1674040588.66,
    "1"
  ]
}
```

So, this is not a problem... also, if I check for PodName exists, it will remove the possibility of getting logs from containers inside a namespace. And If remember correctly, and I can test it, it will impact the `AlertmanagerFailedToSendAlerts`

pkg/gatherers/conditional/gather_containers_logs.go
```diff
@@ -40,6 +40,12 @@ func (g *Gatherer) BuildGatherContainersLogs(paramsInterface interface{}) (gathe
 	}, nil
 }
 
+type podInfo struct {
+	name      string
+	namespace string
+	container string
+}
+
 func (g *Gatherer) gatherContainersLogs(
```
tremes, line 49: I would add some comment here as well please.

pkg/gatherers/conditional/common_test.go
```diff
@@ -71,3 +73,62 @@ func Test_getAlertPodNamespace(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			container, err := getAlertPodContainer(tc.labels)
+			if container != tc.expectedResult {
+				t.Errorf("Expected result to be %s but got %s", tc.expectedResult, container)
+			}
+			if err != tc.expectedErr {
+				t.Errorf("Expected error to be %v but got %v", tc.expectedErr, err)
+			}
```
tremes, line 131: nit: we can use the `assert` library when it's already in our deps.

pkg/gatherers/conditional/common.go
```diff
@@ -4,29 +4,32 @@ import (
 	"fmt"
 )
 
+var (
+	ErrAlertPodNameMissing      = fmt.Errorf("alert is missing 'pod' label")
+	ErrAlertPodNamespaceMissing = fmt.Errorf("alert is missing 'namespace' label")
+	ErrAlertPodContainerMissing = fmt.Errorf("alert is missing 'container' label")
```
tremes, line 10: Perhaps it would be good to include the alert name in the errors. WDYT?

pkg/gatherers/conditional/common_test.go
rluders: I guess that I tried to use it, but CI was complaining about the dependency... But, I guess that I can check the dependency and try again.

pkg/gatherers/conditional/common.go
rluders: Well, it is already logging it with the AlertName at `pkg/gatherers/conditional/gather_containers_logs.go#L68`, and I don't want to inject the AlertName into this function, seems a little bit weird...

So, maybe, I can append the error to the `errs` slice at `pkg/gatherers/conditional/gather_containers_logs.go`, with (or something similar):

```go
errs = append(errs, errors.Wrap(err, params.AlertName))
```

WDYT?

tremes: OK if it's logged with the alert name then it's OK. Thanks

pkg/gatherers/conditional/common_test.go
```diff
@@ -28,13 +29,8 @@ func Test_getAlertPodName(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := getAlertPodName(tt.labels)
-			if (err != nil) != tt.wantErr {
-				t.Errorf("getAlertPodName() error = %v, wantErr %v", err, tt.wantErr)
-				return
-			}
-			if got != tt.want {
-				t.Errorf("getAlertPodName() got = %v, want %v", got, tt.want)
-			}
+			assert.Equal(t, tt.wantErr, err != nil)
```
tremes, line 33: There's `assert.NoError` or something like that.

rluders: Well, this is true but then... I will have to use an if statement to cover `wantErr: true` and `wantErr: false`. Using `assert.Equal` one-line solution.

```go
if tt.wantErr {
    assert.Equal(t, tt.wantErr, true)
} else {
    !assert.NoError(t, tt.wantErr)
}
```
vs
```go
assert.Equal(t, tt.wantErr, err != nil)
```

pkg/gatherers/common/gather_logs.go
tremes: Well you can do `r,_ := regexp.Compile(resourceFilter.PodNameRegexFilter)` out side the for loop and then use the `r`

pkg/gatherers/common/gather_logs.go
```diff
@@ -79,14 +79,22 @@ func CollectLogsFromContainers( //nolint:gocyclo
 
 	var skippedContainers int
 	var records []record.Record
+	var regexErrorMessage = "fails to compile regex '%s' to filter '%s'"
+
+	podNameRegex, err := regexp.Compile(resourceFilter.PodNameRegexFilter)
+	if err != nil {
+		klog.Warningf(regexErrorMessage, resourceFilter.ContainerNameRegexFilter, "pod name")
```
tremes, line 86: This should be `resourceFilter.PodNameRegexFilter`

tremes:
I reviewed and do manual basic test. 
/lgtm

rluders:
Not a feature release, just an enhancement to the existing one.
/label px-approved
/label docs-approved

tremes:
I have already reviewed this. I quickly checked again. 
/lgtm

rluders:
Closing. This data isn't required anymore.

rluders:
Dropping.

PR #683, opened 2022-10-02
OCPBUGS-2915: Updated info link in insights recommendations

rhrmo:

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Unit Tests

- `pkg/insights/insightsreport/insightsreport_test.go`

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-8942

### DIFF
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -162,11 +162,6 @@ func userAgent(releaseVersionEnv string, v apimachineryversion.Info, cv *configv
 	return fmt.Sprintf("insights-operator/%s cluster/%s", gitVersion, cv.Spec.ClusterID)
 }
 
-func (c *Client) SetClusterVersion(cv *configv1.ClusterVersion) *Client {
-	c.clusterVersion = cv
-	return c
-}
-
 func (c *Client) GetClusterVersion() (*configv1.ClusterVersion, error) {
 	if c.clusterVersion != nil {
 		return c.clusterVersion, nil
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -162,7 +162,7 @@ func userAgent(releaseVersionEnv string, v apimachineryversion.Info, cv *configv
 	return fmt.Sprintf("insights-operator/%s cluster/%s", gitVersion, cv.Spec.ClusterID)
 }
 
-func (c *Client) getClusterVersion() (*configv1.ClusterVersion, error) {
+func (c *Client) GetClusterVersion() (*configv1.ClusterVersion, error) {
 	if c.clusterVersion != nil {
 		return c.clusterVersion, nil
 	}
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -27,7 +27,6 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/openshift/insights-operator/pkg/insights"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachineryversion "k8s.io/apimachinery/pkg/version"
 )
@@ -174,9 +173,6 @@ func (c *Client) GetClusterVersion() (*configv1.ClusterVersion, error) {
 	}
 
 	cv, err := gatherConfigClient.ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
-		return nil, nil
-	}
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/insights/insightsclient/insightsclient.go
+++ b/pkg/insights/insightsclient/insightsclient.go
@@ -162,6 +162,11 @@ func userAgent(releaseVersionEnv string, v apimachineryversion.Info, cv *configv
 	return fmt.Sprintf("insights-operator/%s cluster/%s", gitVersion, cv.Spec.ClusterID)
 }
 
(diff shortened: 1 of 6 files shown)


pkg/insights/insightsreport/insightsreport.go
```diff
@@ -292,11 +292,18 @@ func (c *Controller) readInsightsReport(report types.SmartProxyReport) ([]types.
 			continue
 		}
 
+		clusterVersion, err := c.client.GetClusterVersion()
+		if err != nil {
+			klog.Errorf("Unable to extract cluster version")
+			continue
+		}
+
```
tremes, line 300: This doesn't have to be in the for cycle, right?

pkg/insights/metrics.go
```diff
@@ -57,7 +57,7 @@ func (c *InsightsRecommendationCollector) Collect(ch chan<- prometheus.Metric) {
 			prometheus.NewDesc(c.metricName, "", []string{}, prometheus.Labels{
 				"description": rec.Description,
 				"total_risk":  totalRiskToStr(rec.TotalRisk),
-				"info_link":   fmt.Sprintf("https://console.redhat.com/openshift/insights/advisor/recommendations/%s%%7C%s", ruleIDStr, rec.ErrorKey),
+				"info_link":   fmt.Sprintf("https://console.redhat.com/openshift/insights/advisor/clusters/%s?first=%s%%|%s", rec.ClusterID, ruleIDStr, rec.ErrorKey),
```
tremes, line 60: Hmm I am just wondering what the `first` query parameter is for

rhrmo: I think it has something to do with order of recommendations in the hybrid cloud console, but I am not 100% sure. If you put for example `?second=` it works as well. It works even if you leave just `?` in the url and delete the `first=` part. But default url contains `?first=` if you get to it through hybrid cloud console site so I would just let it be.

tremes: OK thanks for checking. I would like to keep it as simple as possible so having no param would be the best IMO but we can ask someone to tell us more about this.

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -162,7 +162,12 @@ func userAgent(releaseVersionEnv string, v apimachineryversion.Info, cv *configv
 	return fmt.Sprintf("insights-operator/%s cluster/%s", gitVersion, cv.Spec.ClusterID)
 }
 
-func (c *Client) getClusterVersion() (*configv1.ClusterVersion, error) {
+func (c *Client) SetClusterVersion(cv *configv1.ClusterVersion) *Client {
+	c.clusterVersion = cv
+	return c
+}
```
tremes, line 168: This function is used only in the tests, right? If so then we should be able to mock it and remove this function IMO.

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -261,11 +265,17 @@ type healthStatusCounts struct {
...
+func (c *Controller) readInsightsReport(report types.SmartProxyReport) ([]types.InsightsRecommendation, healthStatusCounts, time.Time, error) {
 	healthStatus := healthStatusCounts{}
 	healthStatus.total = report.Meta.Count
 	activeRecommendations := []types.InsightsRecommendation{}
 
+	clusterVersion, err := c.client.GetClusterVersion()
+	if err != nil {
+		klog.Errorf("Unable to extract cluster version: %v", err)
+		return nil, healthStatus, time.Time{}, err
+	}
```
tremes, line 277: I think this call has nothing to do with reading the Insights report :) ....but I am thinking how to do it better...

pkg/insights/types/types.go
```diff
@@ -56,4 +58,5 @@ type InsightsRecommendation struct {
 	ErrorKey    string
 	Description string
 	TotalRisk   int
+	ClusterID   v1.ClusterID
```
tremes, line 61: This is not a bad idea at all. I am just thinking if for example setting the cluster version once in the `RecommendationCollector` would be an option to consider.

rhrmo: I will look into it.

pkg/insights/insightsclient/insightsclient.go
tremes: Perhaps we could introduce some similar interface as in https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/conditional/conditional_gatherer.go#L43 and then mock it?

pkg/insights/metrics.go
```diff
@@ -37,6 +38,15 @@ func init() {
...
 	activeRecommendations []types.InsightsRecommendation
 	metricName            string
+	clusterID             v1.ClusterID
+}
+
+func (c *InsightsRecommendationCollector) SetClusterID(clusterID v1.ClusterID) {
+	c.clusterID = clusterID
+}
+
+func (c *InsightsRecommendationCollector) GetClusterID() v1.ClusterID {
```
tremes, line 48: nit: See what says [Effective Go](https://go.dev/doc/effective_go#Getters) and yes I know we have these getters with `Get` all over the place,

rhrmo: Thank you, good point.

pkg/insights/insightsreport/insightsreport.go
```diff
@@ -243,7 +243,8 @@ func (c *Controller) Run(ctx context.Context) {
 	c.StatusController.UpdateStatus(controllerstatus.Summary{Healthy: true})
 	klog.V(2).Info("Starting report retriever")
 	klog.V(2).Infof("Initial config: %v", c.configurator.Config())
-
+	clusterVersion, _ := c.client.GetClusterVersion()
+	insights.RecommendationCollector.SetClusterID(clusterVersion.Spec.ClusterID)
```
tremes, line 247: We definitely need to use and handle the error, because otherwise this line can fail with nil panic.

tremes:
I reviewed and tested this one. Thank you!
/lgtm

PR #691, opened 2022-10-24
fix: storage/ceph path structure

rluders:
This PR fixes CephCluster and StorageCluster resources that uses the same paths.

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gatherers/clusterconfig/openshift_storage_test.go`

## Privacy

## Breaking Changes

Yes.

The path to `ceph_cluster.go` report and `openshift_storage.go` changed:
-  OLD  `config/storage/<namespace>/<name>.json`
-  NEW `config/storage/<namespace>/<resource>/<name>.json`.

## References

https://issues.redhat.com/browse/CCXDEV-9400


docs/gathered-data.md
```diff
@@ -59,7 +59,7 @@ API Reference:
 
 	https://github.com/rook/rook/blob/master/pkg/apis/ceph.rook.io/v1/types.go
 
-* Location in archive: config/storage/<namespace>/<name>.json
+* Location in archive: config/storage/<namespace>/<resource>/<name>.json
```
jholecek-rh, line 62: Will the CephCluster gatherer gather any other CRDs than CephCluster? If not, I suggest to make the path specific and use `cephcluster(s)` instead of `<resource>`.

docs/gathered-data.md
```diff
@@ -797,7 +797,7 @@ API Reference:
 
 	https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/storagecluster_types.go
 
-* Location in archive: config/storage/<namespace>/<name>.json
+* Location in archive: config/storage/<namespace>/<resource>/<name>.json
```
jholecek-rh, line 800: Will the OpenshiftStorage gatherer gather any other CRs than StorageCluster? If not, I suggest to make the path more specific and use `storagecluster(s)` instead of `<resource>`.

This might also be a good opportunity to rename the gatherer from OpenshiftStorage to StorageCluster. I don't care so much about names in the code but I do care about the section title in this document, which happens to be generated from the code.

docs/gathered-data.md
rluders: Yeah, I can change it. Basically what I mean by that was exactly it, that it will be `cephclusters`, but I think that it is better to let it explicit.

docs/gathered-data.md
rluders: Same thing here. I can change it. Also, good idea to rename the Gather.

docs/gathered-data.md
```diff
@@ -979,6 +964,23 @@ gathers the alerts that have been silenced.
...
+## StorageCluster
+
+collects `storageclusters.ocs.openshift.io` resources
+from Openshift Data Foundation Stack.
+
+API Reference:
+
+	https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/storagecluster_types.go
+
+* Location in archive:
```
rluders, line 976: @tremes @jholecek-rh I'm trying to find a nice way to document the path changes. Suggestions?

docs/gathered-data.md
```diff
@@ -979,6 +964,23 @@ gathers the alerts that have been silenced.
...
+	 - 4.11.z : config/storage/<namespace>/<name>.json
+	 - 4.12+  : config/storage/<namespace>/storageclusters/<name>.json
+* Id in config: clusterconfig/openshift_storage
```
rluders, line 979: Also here, how it is going to look when whe change the config key, 'cause we change the Gather name?

tremes: changing config key is probably OK in my opinion.

docs/gathered-data.md
```diff
@@ -979,6 +964,23 @@ gathers the alerts that have been silenced.
   - 4.10+
 
 
+## StorageCluster
```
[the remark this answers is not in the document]
tremes: probably no

docs/gathered-data.md
```diff
@@ -59,7 +59,7 @@ API Reference:
 
 	https://github.com/rook/rook/blob/master/pkg/apis/ceph.rook.io/v1/types.go
 
-* Location in archive: config/storage/<namespace>/<name>.json
+* Location in archive: config/storage/<namespace>/cephclusters/[name].json
```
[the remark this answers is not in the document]
tremes: OK we can create a new bug/task for that

docs/gathered-data.md
tremes: The suggestion below looks good to me. Question is how it's rendered in the markdown.

docs/gathered-data.md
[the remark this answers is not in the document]
jholecek-rh: Good question. I think we should.

Reason: Rules should make conclusions based on data that are present in an archive (as opposed to data that are NOT present in the archive). If there are no data (for a particular resource), it can mean any of

1. No such resources in the cluster
2. Not gathered in the cluster version
3. An error occurred during gathering

To tell the three cases apart, rules use the gathers info. They need to know the gatherer name for that.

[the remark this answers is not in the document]
jholecek-rh: Correction: My previous comment applies to config ID rather than gather name itself. We probably don't need to document the original gatherer name, but we should document the original config ID.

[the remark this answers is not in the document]
tremes: So it's this discussion right https://github.com/openshift/insights-operator/pull/691#discussion_r1004214654

docs/gathered-data.md
jholecek-rh: Looks good assuming we will not backport the change?

docs/gathered-data.md
jholecek-rh: I'd do the same as for the file path.

docs/gathered-data.md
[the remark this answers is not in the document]
jholecek-rh: yep

docs/gathered-data.md
rluders: @jholecek-rh Yes. If we backport it, we are going just to update the documentation the path to the backported version.

docs/gathered-data.md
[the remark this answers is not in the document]
rluders: Fixed. The issue is that we can't use `<%s>` to the placeholders, we should use `{%s}` instead.

rluders:
I still need to generate and add some sample data to the archive.

jholecek-rh:
@rluders Looks good except the availability per OCP version. Are you going to update that separately or it is correct now? The info doesn't match data from the queries that I ran...

docs/gathered-data.md
```diff
@@ -52,17 +52,29 @@ The CRD sizes above are in the raw (uncompressed) state.
...
+| <= 4.11 | config/storage/{namespace}/cephclusters/{name}.json 	 |
+| 4.12+   | config/storage/{namespace}/cephclusters/{name}.json 	 |
+
+### Config ID:
+clusterconfig/ceph_cluster
+
+### Since versions:
+* 4.8.24
+* 4.9.23
+* 4.10.17
```
tremes, line 76: Are we really sure that this data is collected in "such old OCP versions" ?

docs/gathered-data.md
```diff
@@ -52,17 +52,29 @@ The CRD sizes above are in the raw (uncompressed) state.
...
 
-* Location in archive: config/storage/<namespace>/<name>.json
-* Id in config: clusterconfig/ceph_cluster
-* Since versions:
-  - 4.12+
+### Location in archive:
+| Version | Path													 |
+| ------- | ----------------------------------------------------- |
+| <= 4.11 | config/storage/{namespace}/cephclusters/{name}.json 	 |
+| 4.12+   | config/storage/{namespace}/cephclusters/{name}.json 	 |
```
tremes, line 68: well this should be 4.13+ right? And then we will need to update with some 4.12.z version, I guess.

docs/gathered-data.md
```diff
@@ -979,6 +976,32 @@ gathers the alerts that have been silenced.
...
+| ------- | --------------------------------------------------------	|
+| <= 4.11 | config/storage/{namespace}/{name}.json 					|
+| 4.12+   | config/storage/{namespace}/storageclusters/{name}.json 	|
+
+### Config ID:
+clusterconfig/storage_cluster
+
+### Since versions:
+* 4.8.49+
+* 4.10.30+
```
tremes, line 1001: 4.9 is missing? Again I would double check the versions which the data should be available in.

tremes: See the https://github.com/openshift/insights-operator/blob/master/CHANGELOG.md and the `Gather ODF config data` commit message

tremes: The bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=2079315 talks about backporting to 4.10, but I actually can't see any cloned or related BZ for such backport.

docs/gathered-data.md
tremes: OK so this seems to go down to 4.9 see https://issues.redhat.com/browse/OCPBUGS-251. Can we verify the 4.8 then?

docs/gathered-data.md
rluders: Not 100% yet, I was double checking in our channel if we will bp to already push it with the correct version.

docs/gathered-data.md
rluders: https://github.com/openshift/insights-operator/pulls?q=is%3Apr+is%3Aclosed+ceph

docs/gathered-data.md
rluders: Well, I was actually taking as reference the comment from @jholecek-rh - so, maybe his data isn't correct?

docs/gathered-data.md
tremes: Yes, but first of all I would start with correct versions in the master branch

docs/gathered-data.md
tremes: OK thanks :)

docs/gathered-data.md
tremes: I would say: _trust but verify_ :smiley:

jholecek-rh: No, my [data](https://github.com/openshift/insights-operator/pull/691#issuecomment-1289265414) is not authoritative!

* The data is based on archives that we receive. If there were no clusters with the CRs at a particular version, the version would not show up.
* The data may be skewed because of clusters that have not finished upgrading.
* The data I shared here is not even the complete output of the query that I run.

The purpose of the data was to demonstrate the discrepancy between the documentation and real data.

@tremes Mentioned he would update the script that can update the documentation with backports. I'm OK to wait for that (if it happens this year or so). :)

rluders: OK. Can we agree that I'll adjust the data for the current release, and we can handle the documentation update/fix in other task? Otherwise this will take a very long time to get merged, and the verification doesn't seems to be easy. I already created the task for updating/validating all the gatherers documentation.

rluders: Maybe we need to create a spreadsheet with all gathers to track down this to update the documentation.

tremes: It's up to you. I think it would be better to have it in one PR. We don't hurry, do we? We are speaking mainly about two gatherers right now so I don't see any reason for a spreadsheet with all of them.

rluders: Well, since we don't hurry about it, I can verify and update at least the ceph and storage documentation.

About the spreadsheet, we are going to verify and update all the gatherers information. At this point, I'm not sure if all the information that we have in our documentation is correct, and a spreadsheet where we can colaborate, comment, discuss and edit together makes IMHO a lot of diference to track down all this before we update the docs.

pkg/gatherers/clusterconfig/ceph_cluster.go
```diff
@@ -5,24 +5,40 @@ import (
...
+// | Version | Path													 |
+// | ------- | ----------------------------------------------------- |
+// | <= 4.12 | config/storage/{namespace}/{name}.json 				 |
+// | 4.13+   | config/storage/{namespace}/cephclusters/{name}.json 	 |
+//
+// ### Config ID:
+// clusterconfig/ceph_cluster
 //
-//	https://github.com/rook/rook/blob/master/pkg/apis/ceph.rook.io/v1/types.go
+// ### Released version:
```
jholecek-rh, line 34: Other gatherers are using a single list "Since versions" instead of "Release version" and "Backport versions". Is this a new norm? If so:

* Do you plan to update documentation of all gatherers? 
* Could you always include both sections, possibly saying "None" in "Backport versions" to prevent any doubts?

tremes: I would keep `Since versions`

docs/gathered-data.md
```diff
@@ -979,6 +979,30 @@ gathers the alerts that have been silenced.
...
+## StorageCluster
+
+Collects `storageclusters.ocs.openshift.io` resources
+from Openshift Data Foundation Stack.
+
+### API Reference:
+https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/storagecluster_types.go
+
+### Sample data
+docs/insights-archive-sample/config/storage/openshift-storage/storageclusters/ocs-storagecluster.json
```
tremes, line 991: Again this is changing the original format. We can change it, but then it should be changed everywhere,

tremes: Ohh sorry this is new right? I confused myself with `Location in archive:` ...so it's probably OK and you can ignore me

pkg/gatherers/clusterconfig/ceph_cluster.go
rluders: Keeping only the `Since versions` is confusing. I would prefer to know what version this was released in and the versions that it was backported. It is clear, with no space for interpretation.

- Yes. I plan to update all the documentation. Separated PR soon. (we need to validate it, after all - I found a lot of inconsistencies)
- Yes. I can add `None` for `Backport versions`.

jholecek-rh: I don't see how we could use that distinction. We care about what OCP versions have the data, not how the change got to those versions.

Anyway, feel free to go ahead with the change if it helps you somehow.

docs/gathered-data.md
```diff
@@ -52,17 +52,32 @@ The CRD sizes above are in the raw (uncompressed) state.
...
+clusterconfig/ceph_cluster
+
+### Released version:
+* 4.12
+
+### Backport versions:
+* 4.8.49
+* 4.9.48
+* 4.10.31
+* 4.11.2
```
tremes, line 80: So why can't we keep the original `Since versions` format? We will have two gatherers docs with different format from the rest. I personally can't see any difference between released version and backport versions.

JoaoFula:
@rluders it is being taken into account as [CCXDEV-8803](https://issues.redhat.com/browse/CCXDEV-8803) is being developed. Regarding the failing tests, I will deactivate test_ocm_response (IO is not being degraded in time).

JoaoFula:
@rluders it is being taken into account as [CCXDEV-8803](https://issues.redhat.com/browse/CCXDEV-8803) is being developed. Regarding the failing tests, I will deactivate test_ocm_response (IO is not being degraded in time).

tremes:
I reviewed but I didn't run the code since it should be just renaming and sample archive update.
/lgtm

PR #694, opened 2022-11-02
Change of kube-system namespace configmap location according to docs. 

rhrmo:

## Categories

- [X] Bugfix
- [ ] Enhancement
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/config/configmaps/kube-system/cluster-config-v1/install-config`

## Unit Tests

- `pkg/gatherers/clusterconfig/cluster_config_v1_config_map_test.go`

## Privacy

There are no sensitive data in the newly collected information.

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-8900


pkg/gatherers/clusterconfig/config_maps.go
```diff
@@ -55,10 +55,6 @@ func (g *Gatherer) GatherConfigMaps(ctx context.Context) ([]record.Record, []err
 	records = append(records, monitoringRec...)
 	errs = append(errs, monitoringErrs...)
 
-	clusterConfigV1Rec, clusterConfigV1Errs := gatherClusterConfigV1(ctx, coreClient)
-	records = append(records, clusterConfigV1Rec...)
-	errs = append(errs, clusterConfigV1Errs...)
```
tremes, line 60: I believe we don't want to do this. The task was about fixing the archive path (to be aligned with other config maps) and the corresponding sample archive sample AFAIK.

rhrmo: Task notes:
"Since the ConfigMap from the kube-system namespace is not used in any rule at the moment, I actually prefer to change the way it is stored to match the documented structure.

However, I don't know why the kube-system ConfigMap was added in the first place and who might be using it. It might have been added because of a rule."

I can update the structure instead, I'm not sure which one is better.

docs/insights-archive-sample/config/configmaps/kube-system/cluster-config-v1/install-config.json
```diff
@@ -0,0 +1 @@
+"apiVersion: v1\nbaseDomain: xxxxxxxxxxxxxxxxxxx\ncompute:\n- hyperthreading: Enabled\n  name: worker\n  platform: {}\n  replicas: 3\ncontrolPlane:\n  hyperthreading: Enabled\n  name: master\n  platform: {}\n  replicas: 3\nmetadata:\n  creationTimestamp: null\n  name: dev.rhrmo-dev.7.11.2022\nnetworking:\n  clusterNetwork:\n  - cidr: 10.128.0.0/14\n    hostPrefix: 23\n  machineNetwork:\n  - cidr: 10.0.0.0/16\n  networkType: OVNKubernetes\n  serviceNetwork:\n  - 172.30.0.0/16\nplatform:\n  aws:\n    region: xxxxxxxxx\npublish: External\npullSecret: \"\"\n"
```
tremes, line 1: This doesn't look like valid json file (and from what I've seen locally, it's not a valid one)

pkg/gatherers/clusterconfig/cluster_config_v1_config_map.go
```diff
@@ -41,11 +41,13 @@ func gatherClusterConfigV1(ctx context.Context, coreClient corev1client.CoreV1In
...
-		Name: fmt.Sprintf("config/configmaps/%s/%s", configMap.Namespace, configMap.Name),
-		Item: record.JSONMarshaller{Object: configMap},
-	}}, nil
+	var name string
+	var item record.Marshalable
+	for dk, dv := range configMap.Data {
+		name = fmt.Sprintf("config/configmaps/%s/%s/%s", configMap.Namespace, configMap.Name, dk)
+		item = record.JSONMarshaller{Object: dv}
+	}
+	return []record.Record{{Name: name, Item: item}}, nil
```
tremes, line 50: This looks weird TBH. It seems to rely on the fact that there's only one key/value pair in the config map. Hmm it's probably related to the line 43 here.

docs/insights-archive-sample/config/configmaps/kube-system/cluster-config-v1/install-config.json
rhrmo: Will re-naming it to just install-config without the .json at the end be enough? :thinking:

tremes:
I reviewed and verified in my cluster. Thank you!
/lgtm

tremes:
Minor fix of one configmap path in the archive. 
/label px-approved
/label qe-approved
/label docs-approved

PR #704, opened 2022-12-09
chore(docs): update gathered documentation

rluders:
This PR updates the in-code documentation about gathering data.

> ⚠️ Sorry in advance for the enormous number of files changed, but this needs to be handled in one single PR.

The scope:

- Replace the old template with the new one
- All data verified (version, backports, etc)
- Regenerate the documentation

Out of scope:

- Add missing sample data
- Backport this documentation

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

N/A

## Documentation

- `docs/gathered-data.md`

## Unit Tests

N/A

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-9566

### DIFF
--- a/cmd/gendoc/main.go
+++ b/cmd/gendoc/main.go
@@ -24,7 +24,7 @@ var (
 	randSource      = rand.NewSource(time.Now().UnixNano())
 	reGather        = regexp.MustCompile(`^((Build)?Gather)(.*)`)
 	reExample       = regexp.MustCompile(`^(Example)(.*)`)
-	reSampleArchive = regexp.MustCompile(`docs\/(insights-archive-sample\/.*)`)
+	reSampleArchive = regexp.MustCompile(`docs/(insights-archive-sample/.*)`)
 )
 
 type DocBlock struct {
--- a/cmd/gendoc/main.go
+++ b/cmd/gendoc/main.go
@@ -58,7 +58,6 @@ func main() {
 	}
 	keys := make([]string, 0, len(md))
 	for k := range md {
-
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -1779,7 +1779,7 @@ None
 ### Changes
 - Renamed from `OpenshiftStorage` to `StorageCluster` in version `4.12.0+`
 - Config ID changed from `clusterconfig/openshift_storage` to `clusterconfig/storage_cluster` in version `4.12.0+`
-
+- In OCP 4.11 and OCP 4.12, the location of gathered data collides with data gathered by the [GatherCephCluster](#cephcluster) gatherer. It is practically impossible to tell the two resources apart. Use with caution.
 
 ## SupportSecret
 
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -1516,8 +1516,8 @@ Collects the cluster's `PodDisruptionBudgets`.
 - 4.6.0
 
 ### Backported versions
-- 4.5.15+
 - 4.4.30+
+- 4.5.15+
 
 ### Changes
 None
--- a/docs/gathered-data.md
+++ b/docs/gathered-data.md
@@ -13,15 +13,34 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
 
 ## APIRequestCounts
 
-BuildGatherApiRequestCounts creates a gathering closure which collects API requests counts for the
-resources mentioned in the alert provided as a string parameter
-Params is of type AlertIsFiringConditionParams:
-  - alert_name string - name of the firing alert
+Creates a gathering closure which collects API requests counts for the
+resources mentioned in the alert provided as a string parameter.
 
-* Location in archive: conditional/alerts/{alert_name}/api_request_counts.json
-* Id in config: conditional/api_request_counts_of_resource_from_alert
-* Since versions:
-  - 4.10+
+Params is of type AlertIsFiringConditionParams:
+- alert_name string - name of the firing alert
+
+### API Reference
+None
+
+### Sample data
+- [docs/insights-archive-sample/conditional/alerts/APIRemovedInNextEUSReleaseInUse/api_request_counts.json](./insights-archive-sample/conditional/alerts/APIRemovedInNextEUSReleaseInUse/api_request_counts.json)
+
+### Location in archive
+| Version   | Path												 	 |
+| --------- | ------------------------------------------------------- |
+| >= 4.10.0 | conditional/alerts/{alert_name}/api_request_counts.json |
+
+### Config ID
+`conditional/api_request_counts_of_resource_from_alert`
+
+### Released version
+- 4.10.0
+
+### Backported versions
+- 4.9.6+
+
+### Notes
+None
 
 
 ## ActiveAlerts
@@ -38,13 +57,13 @@ None
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.12   | config/alerts.json				                        	|
+| >= 4.12.0 | config/alerts.json				                        	|
 
 ### Config ID
 `clusterconfig/active_alerts`
 
 ### Released version
-- 4.12
+- 4.12.0
 
 ### Backported versions
 None
@@ -68,13 +87,13 @@ None
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.6    | config/crd/{name}.json 					                |
+| >= 4.6.0  | config/crd/{name}.json 					                |
 
 ### Config ID
 `clusterconfig/crds`
 
 ### Released version
-- 4.6
+- 4.6.0
 
 ### Backported versions
 None
@@ -97,16 +116,16 @@ from Openshift Data Foundation Stack.
 - [docs/insights-archive-sample/config/storage/openshift-storage/cephclusters/ocs-storagecluster-cephcluster.json](./insights-archive-sample/config/storage/openshift-storage/cephclusters/ocs-storagecluster-cephcluster.json)
 
 ### Location in archive
-| Version | Path													 |
-| ------- | ----------------------------------------------------- |
-| <= 4.12 | config/storage/{namespace}/{name}.json 				 |
-| 4.13+   | config/storage/{namespace}/cephclusters/{name}.json 	 |
+| Version   | Path													 |
+| --------- | ----------------------------------------------------- |
+| <= 4.12.0 | config/storage/{namespace}/{name}.json 				 |
+| 4.13.0+   | config/storage/{namespace}/cephclusters/{name}.json 	 |
 
 ### Config ID
 `clusterconfig/ceph_cluster`
 
 ### Released version
-- 4.12
+- 4.12.0
 
 ### Backported versions
 - 4.8.49+
@@ -133,13 +152,13 @@ when Now < ValidBefore or Now > ValidAfter
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.5    | config/certificatesigningrequests 					        |
+| >= 4.5.0  | config/certificatesigningrequests 					        |
 
 ### Config ID
 `clusterconfig/certificate_signing_requests`
 
 ### Released version
-- 4.5
+- 4.5.0
 
 ### Backported versions
 - 4.3.25+
@@ -163,13 +182,13 @@ Collects the cluster Authentication - the Authentication with name cluster.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.2    | config/authentication.json 								|
+| >= 4.2.0  | config/authentication.json 								|
 
 ### Config ID
 `clusterconfig/authentication`
 
 ### Released version
-- 4.2
+- 4.2.0
 
 ### Backported versions
 None
@@ -192,13 +211,13 @@ Collects the cluster FeatureGate - the FeatureGate with name cluster.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.2    | config/featuregate.json 					                |
+| >= 4.2.0  | config/featuregate.json 					                |
 
 ### Config ID
 `clusterconfig/feature_gates`
 
 ### Released version
-- 4.2
+- 4.2.0
 
 ### Backported versions
 None
@@ -221,13 +240,13 @@ Collects cluster `images.config.openshift.io` resource definition.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| > 4.11    | config/image.json					                        |
+| > 4.11.0  | config/image.json					                        |
 
 ### Config ID
 `clusterconfig/image`
 
 ### Released version
-- 4.11
+- 4.11.0
 
 ### Backported versions
 - 4.10.8+
@@ -249,14 +268,14 @@ None
 ### Location in archive
 | Version   | Path																			     |
 | --------- | ----------------------------------------------------------------------------------- |
-| <= 4.5    | config/imagepruner.json														 	 |
-| >= 4.6    | config/clusteroperator/{group}/{kind}/{name}.json |
+| <= 4.5.0  | config/imagepruner.json														 	 |
+| >= 4.6.0  | config/clusteroperator/{group}/{kind}/{name}.json |
 
 ### Config ID
 `clusterconfig/image_pruners`
 
 ### Released version
-- 4.5
+- 4.5.0
 
 ### Backported versions
 None
@@ -267,22 +286,36 @@ None
 
 ## ClusterImageRegistry
 
-fetches the cluster Image Registry configuration
+Collects the cluster Image Registry configuration
 
-**Conditional data**: If the Image Registry configuration uses any PersistentVolumeClaim for the storage, the corresponding
-PersistentVolume definition is gathered
+### API Reference
+None
 
-* Location in archive: config/clusteroperator/imageregistry.operator.openshift.io/config/cluster.json
-* Location in older versions: config/imageregistry.json
-* Id in config: clusterconfig/image_registries
-* Since versions:
-  - 4.3.40+
-  - 4.4.12+
-  - 4.5+
+### Sample data
+- [docs/insights-archive-sample/config/clusteroperator/imageregistry.operator.openshift.io/config/cluster.json](./insights-archive-sample/config/clusteroperator/imageregistry.operator.openshift.io/config/cluster.json)
 
-* PV definition since versions:
-  - 4.6.20+
-  - 4.7+
+### Location in archive
+| Version   | Path																			|
+| --------- | ------------------------------------------------------------------------------ |
+| < 4.7.0   | config/imageregistry.json 					                        			|
+| >= 4.7.0  | config/clusteroperator/imageregistry.operator.openshift.io/config/cluster.json |
+| >= 4.6.20 | config/persistentvolumes/{name}.json 											|
+
+### Config ID
+`clusterconfig/image_registries`
+
+### Released version
+- 4.5.0
+
+### Backported versions
+- 4.3.40+
+- 4.4.12+
+
+### Notes
+PersistentVolumes was included at `4.7.0` and backported to `4.6.20+`.
+
+**Conditional data**: If the Image Registry configuration uses any PersistentVolumeClaim for the storage,
+the corresponding PersistentVolume definition is gathered
 
 
 ## ClusterInfrastructure
@@ -299,13 +332,13 @@ Collects the cluster Infrastructure - the Infrastructure with name cluster.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.2    | config/infrastructure.json 					            |
+| >= 4.2.0  | config/infrastructure.json 					            |
 
 ### Config ID
 `clusterconfig/infrastructures`
 
 ### Released version
-- 4.2
+- 4.2.0
 
 ### Backported versions
 None
@@ -328,13 +361,13 @@ Collects the cluster Ingress - the Ingress with name cluster.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.2    | config/ingress.json 					                    |
+| >= 4.2.0  | config/ingress.json 					                    |
 
 ### Config ID
 `clusterconfig/ingress`
 
 ### Released version
-- 4.2
+- 4.2.0
 
 ### Backported versions
 None
@@ -352,18 +385,18 @@ Collects the cluster Network - the Network with name cluster.
 - https://docs.openshift.com/container-platform/4.3/rest_api/index.html#network-v1-config-openshift-io
 
 ### Sample data
-- [docs/insights-archive-sample/config/.json](./insights-archive-sample/config/.json)
+- [docs/insights-archive-sample/config/network.json](./insights-archive-sample/config/network.json)
 
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.2    | config/network.json 					                    |
+| >= 4.2.0  | config/network.json 					                    |
 
 ### Config ID
 `clusterconfig/networks`
 
 ### Released version
-- 4.2
+- 4.2.0
 
 ### Backported versions
 None
@@ -386,13 +419,13 @@ Collects the cluster OAuth - the OAuth with name cluster.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.2    | config/oauth.json 					                        |
+| >= 4.2.0  | config/oauth.json 					                        |
 
 ### Config ID
 `clusterconfig/oauths`
 
 ### Released version
-- 4.2
+- 4.2.0
 
 ### Backported versions
 None
@@ -421,7 +454,7 @@ None
 ### Location in archive
 | Version   | Path																    |
 | --------- | ---------------------------------------------------------------------- |
-| >= 4.3    | config/pod/{namespace}/{pod}.json										|
+| >= 4.3.0  | config/pod/{namespace}/{pod}.json										|
 | >= 4.8.2  | config/pod/{namespace}/logs/{pod}/{container}_{current|previous}.log   |
 | >= 4.8.2  | events/																|
 
@@ -432,7 +465,7 @@ None
 `clusterconfig/operators_pods_and_events`
 
 ### Released version
-- 4.3
+- 4.3.0
 
 ### Backported versions
 - 4.6.35+
@@ -445,20 +478,36 @@ It used to be Specs for `ClusterOperators`, and it was initially included at `4.
 
 ## ClusterOperators
 
-collects all the ClusterOperators definitions and their related resources
-from the `operator.openshift.io` group.
+Collects all the ClusterOperators definitions and their resources.
 
-The Kubernetes api https://github.com/openshift/client-go/blob/master/config/clientset/versioned/typed/config/v1/clusteroperator.go#L62
-Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#clusteroperatorlist-v1config-openshift-io
+### API Reference
+- https://github.com/openshift/client-go/blob/master/config/clientset/versioned/typed/config/v1/clusteroperator.go#L62
+- https://docs.openshift.com/container-platform/4.3/rest_api/index.html#clusteroperatorlist-v1config-openshift-io
 
-* Location of operators related resources: config/clusteroperator/{group}/{kind}/{name}
-* Location of operators in archive: config/clusteroperator/
-* Location of operators related resources in older versions: config/clusteroperator/{kind}-{name}
-* See: [docs/insights-archive-sample/config/clusteroperator](./insights-archive-sample/config/clusteroperator)
-* Id in config: clusterconfig/operators
-* Spec config for CO resources since versions:
-  - 4.6.16+
-  - 4.7+
+### Sample data
+- [docs/insights-archive-sample/config/clusteroperator](./insights-archive-sample/config/clusteroperator)
+
+### Location in archive
+| Version   | Path														|
+| --------- | --------------------------------------------------------	|
+| >= 4.3.0  | config/clusteroperator/ 									|
+| < 4.7.0   | config/clusteroperator/{kind}-{name}.json 					|
+| >= 4.7.0  | config/clusteroperator/{group}/{kind}/{name}.json 			|
+
+### Config ID
+`clusterconfig/`
+
+### Released version
+- 4.2.0
+
+### Backported versions
+None
+
+### Notes
+The `ClusterOperators` were used to also collect `pods` and `events`, it changed at the `4.8.2` release. The `pods`
+and `events` gathering were introduced in `4.3.0` and backported to `4.2.10`.
+
+* Spec config for CO resources was introduced at `4.7.0` and backported to`4.6.16+`
 
 
 ## ClusterProxy
@@ -475,13 +524,13 @@ Collects the cluster Proxy - the Proxy with name cluster.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.3    | config/proxy.json 					                        |
+| >= 4.3.0  | config/proxy.json 					                        |
 
 ### Config ID
 `clusterconfig/proxies`
 
 ### Released version
-- 4.3
+- 4.3.0
 
 ### Backported versions
 None
@@ -492,47 +541,80 @@ None
 
 ## ClusterVersion
 
-fetches the ClusterVersion (including the cluster ID) with the name 'version' and its resources.
+Collects the ClusterVersion (including the cluster ID) with the name 'version' and its resources.
 
-The Kubernetes api https://github.com/openshift/client-go/blob/master/config/clientset/versioned/typed/config/v1/clusterversion.go#L50
-Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#clusterversion-v1config-openshift-io
+### API Reference
+- https://github.com/openshift/client-go/blob/master/config/clientset/versioned/typed/config/v1/clusterversion.go#L50
+- https://docs.openshift.com/container-platform/4.3/rest_api/index.html#clusterversion-v1config-openshift-io
 
-* Location in archive: config/version/
-* See: [docs/insights-archive-sample/config/version](./insights-archive-sample/config/version)
-* Location of pods in archive: config/pod/
-* Location of events in archive: events/
-* Location of cluster ID: config/id
-* See: [docs/insights-archive-sample/config/id](./insights-archive-sample/config/id)
-* Id in config: clusterconfig/version
+### Sample data
+- [docs/insights-archive-sample/config/version.json](./insights-archive-sample/config/version.json)
+- [docs/insights-archive-sample/config/pod](./insights-archive-sample/config/pod)
+- [docs/insights-archive-sample/events/](./insights-archive-sample/events/)
+- [docs/insights-archive-sample/config/id](./insights-archive-sample/config/id)
+
+### Location in archive
+| Version   | Path														|
+| --------- | ---------------------------------------------------------- |
+| >= 4.2.0  | config/version.json										|
+| >- 4.2.0  | config/id													|
+| >= 4.8.2  | config/pod/openshift-cluster-version/{name}.json			|
+| >= 4.8.2  | events/{namespace}.json									|
+
+### Config ID
+`clusterconfig/version`
+
+### Released version
+- 4.2.0
+
+### Backported versions
+None
+
+### Notes
+None
 
 
 ## ConfigMaps
 
-fetches the ConfigMaps from namespace openshift-config
+Collects the ConfigMaps from namespace openshift-config
 and tries to fetch "cluster-monitoring-config" ConfigMap from openshift-monitoring namespace.
 
-Anonymization: If the content of ConfigMap contains a parseable PEM structure (like certificate) it removes the inside of PEM blocks.
-For ConfigMap of type BinaryData it is encoded as standard base64.
+### API Reference
+- https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/configmap.go#L80
+- https://docs.openshift.com/container-platform/4.3/rest_api/index.html#configmaplist-v1core
+
+### Sample data
+- [docs/insights-archive-sample/config/configmaps](./insights-archive-sample/config/configmaps)
+
+### Location in archive
+| Version   | Path														|
+| --------- | ---------------------------------------------------------- |
+| < 4.7.0   | config/configmaps/{configmap}								|
+| >= 4.7.0  | config/configmaps/{namespace}/{name}/{configmap}         	|
+
+### Config ID
+`clusterconfig/config_maps`
+
+### Released version
+- 4.5.0
+
+### Backported versions
+- 4.3.23+
+- 4.4.6+
+
+### Notes
+**Anonymization**:
+If the content of ConfigMap contains a parseable PEM structure (like certificate) it removes
+the inside of PEM blocks. For ConfigMap of type BinaryData it is encoded as standard base64.
 In the archive under configmaps we store name of the namespace, name of the ConfigMap and then each ConfigMap Key.
 For example config/configmaps/NAMESPACENAME/CONFIGMAPNAME/CONFIGMAPKEY1
 
-The Kubernetes api https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/configmap.go#L80
-Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#configmaplist-v1core
-
-* Location in archive: config/configmaps/{namespace-name}/{configmap-name}/
-* Location in older versions: config/configmaps/{configmap-name}/
-* See: [docs/insights-archive-sample/config/configmaps](./insights-archive-sample/config/configmaps)
-* Id in config: clusterconfig/config_maps
-* Since versions:
-  - 4.3.25+
-  - 4.4.6+
-  - 4.5+
-
-* "cluster-monitoring-config" ConfigMap data since versions:
+**Additional data**:
+- "cluster-monitoring-config" ConfigMap data since versions:
   - 4.6.22+
   - 4.7+
 
-* "cluster-config-v1" ConfigMap since versions:
+- "cluster-config-v1" ConfigMap since versions:
   - 4.9+
 
 
@@ -550,13 +632,13 @@ None
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.6    | config/running_containers.json 							|
+| >= 4.6.0  | config/running_containers.json 							|
 
 ### Config ID
 `clusterconfig/container_images`
 
 ### Released version
-- 4.6
+- 4.6.0
 
 ### Backported versions
 None
@@ -579,13 +661,13 @@ Collects ContainerRuntimeConfig information.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.7    | config/containerruntimeconfigs/{name}.json	    			|
+| >= 4.7.0  | config/containerruntimeconfigs/{name}.json	    			|
 
 ### Config ID
 `clusterconfig/container_runtime_configs`
 
 ### Released version
-- 4.7
+- 4.7.0
 
 ### Backported versions
 - 4.6.18+
@@ -596,12 +678,31 @@ None
 
 ## ContainersLogs
 
-collects either current or previous containers logs for pods firing one of the configured alerts.
+Collects either current or previous containers logs for pods firing one of the
+configured alerts.
 
-* Location in archive: conditional/namespaces/{namespace}/pods/{pod}/containers/{container}/{logs|logs-previous}/last-{tail-length}-lines.log
-* Id in config: conditional/containers_logs
-* Since versions:
-  - 4.10+
+### API Reference
+-
+
+### Sample data
+- [docs/insights-archive-sample/conditional/namespaces/openshift-cluster-samples-operator/pods/cluster-samples-operator-8ffb9b45f-49mjr/containers/cluster-samples-operator-watch/logs/last-100-lines.log](./insights-archive-sample/conditional/namespaces/openshift-cluster-samples-operator/pods/cluster-samples-operator-8ffb9b45f-49mjr/containers/cluster-samples-operator-watch/logs/last-100-lines.log)
+
+### Location in archive
+| Version   | Path																										   			|
+| --------- | ---------------------------------------------------------------------------------------------------------------------- |
+| >= 4.10.0 | conditional/namespaces/{namespace}/pods/{pod}/containers/{container}/{logs|logs-previous}/last-{tail-length}-lines.log |
+
+### Config ID
+`conditional/containers_logs`
+
+### Released version
+- 4.10.0
+
+### Backported versions
+None
+
+### Notes
+None
 
 
 ## CostManagementMetricsConfigs
@@ -617,13 +718,13 @@ None
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.10   | config/cost_management_metrics_configs/{name}.json 		|
+| >= 4.10.0 | config/cost_management_metrics_configs/{name}.json 		|
 
 ### Config ID
 `clusterconfig/cost_management_metrics_configs`
 
 ### Released version
-- 4.10
+- 4.10.0
 
 ### Backported versions
 - 4.8.27+
@@ -648,13 +749,13 @@ None
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.10   | config/dvo_metrics 					                    |
+| >= 4.10.0 | config/dvo_metrics 					                    |
 
 ### Config ID
 `clusterconfig/dvo_metrics`
 
 ### Released version
-- 4.10
+- 4.10.0
 
 ### Backported versions
 None
@@ -677,13 +778,13 @@ Collects HostSubnet information.
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.6    | config/hostsubnet/{name}.json				                |
+| >= 4.6.0  | config/hostsubnet/{name}.json				                |
 
 ### Config ID
 `clusterconfig/host_subnets`
 
 ### Released version
-- 4.6
+- 4.6.0
 
 ### Backported versions
 - 4.4.29+
@@ -695,18 +796,33 @@ None
 
 ## ImageStreamsOfNamespace
 
-creates a gathering closure which collects image streams from the provided namespace
+Creates a gathering closure which collects image streams from the provided namespace
+
 Params is of type GatherImageStreamsOfNamespaceParams:
-  - namespace string - namespace from which to collect image streams
+- namespace string - namespace from which to collect image streams
 
-API reference:
+### API Reference
+- https://docs.openshift.com/container-platform/4.7/rest_api/image_apis/imagestream-image-openshift-io-v1.html#apisimage-openshift-iov1namespacesnamespaceimagestreams
 
-	https://docs.openshift.com/container-platform/4.7/rest_api/image_apis/imagestream-image-openshift-io-v1.html#apisimage-openshift-iov1namespacesnamespaceimagestreams
+### Sample data
+- [docs/insights-archive-sample/conditional/namespaces/openshift-cluster-samples-operator/imagestreams/example.json](./insights-archive-sample/conditional/namespaces/openshift-cluster-samples-operator/imagestreams/example.json)
 
-* Location in archive: conditional/namespaces/{namespace}/imagestreams/{name}
-* Id in config: conditional/image_streams_of_namespace
-* Since versions:
-  - 4.9+
+### Location in archive
+| Version   | Path														|
+| --------- | --------------------------------------------------------	|
+| >= 4.9.0  | conditional/namespaces/{namespace}/imagestreams/{name}.json |
+
+### Config ID
+`conditional/image_streams_of_namespace`
+
+### Released version
+- 4.9.0
+
+### Backported versions
+None
+
+### Notes
+None
 
 
 ## InstallPlans
@@ -724,13 +840,13 @@ It also collects Total number of all installplans and all non-unique installplan
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.7    | config/instalplans.json									|
+| >= 4.7.0  | config/instalplans.json									|
 
 ### Config ID
 `clusterconfig/install_plans`
 
 ### Released version
-- 4.7
+- 4.7.0
 
 ### Backported versions
 - 4.5.33+
@@ -753,13 +869,13 @@ None
 ### Location in archive
 | Version   | Path														|
 | --------- | --------------------------------------------------------	|
-| >= 4.10   | config/.json 					                        	|
+| >= 4.10.0 | config/.json 					                        	|
 
 ### Config ID
 `clusterconfig/jaegers`
 
 ### Released version
-- 4.10
+- 4.10.0
 
 ### Backported versions
 None
@@ -785,13 +901,13 @@ Collects ogs from kube-controller-manager pods in the
 ### Location in archive
 | Version   | Path																	 |
 | --------- | ----------------------------------------------------------------------- |
-|  >= 4.11  | config/pod/openshift-kube-controller-manager/logs/{pod-name}/errors.log |
+| >= 4.11.0 | config/pod/openshift-kube-controller-manager/logs/{pod-name}/errors.log |
 
 ### Config ID
 `clusterconfig/kube_controller_manager_logs`
 
 ### Released version
-- 4.11
+- 4.11.0
 
 ### Backported versions
 - 4.10.6+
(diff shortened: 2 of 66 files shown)


pkg/gatherers/clusterconfig/const.go
```diff
@@ -58,6 +58,14 @@ var (
 	costManagementMetricsConfigResource = schema.GroupVersionResource{
 		Group: "costmanagement-metrics-cfg.openshift.io", Version: "v1beta1", Resource: "costmanagementmetricsconfigs",
 	}
+	operatorGVR = schema.GroupVersionResource{
```
rluders, line 61: I moved the GVR from `pkg/gatherers/clusterconfig/olm_operators.go` to this file to keep some code consistency once all GVRs seem to be declared here.

cmd/gendoc/main.go
```diff
@@ -337,7 +339,8 @@ func parseDoc(method, doc string) *DocBlock {
 		doc = strings.TrimLeft(doc, method)
 	}
 	doc = strings.TrimLeft(doc, " ")
-
+	// generates the link to the sample archive
+	doc = reSampleArchive.ReplaceAllString(doc, "[$0](./$1)")
```
rluders, line 343: This is a very simple new feature that looks for the sample data provided in the documentation and converts it into a link, making it easier to access the sample file from the documentation.

pkg/gatherers/clusterconfig/active_alerts.go
```diff
@@ -51,17 +65,20 @@ func gatherActiveAlerts(ctx context.Context, alertsClient rest.Interface) ([]rec
 		return nil, []error{err}
 	}
 
+	// the maximal number of recorded alerts
+	var limit = 1000
```
rluders, line 69: This used to be a constant declared at the top of the file. IMHO, it seems a little bit problematic to have it declared in the package scoped and is used only here, so, I converted it to a local variable inside the method.

pkg/gatherers/clusterconfig/certificate_signing_requests.go
```diff
@@ -53,7 +59,9 @@ func (g *Gatherer) GatherCertificateSigningRequests(ctx context.Context) ([]reco
 func gatherCertificateSigningRequests(ctx context.Context,
 	certClient certificatesv1.CertificateSigningRequestsGetter) ([]record.Record, []error) {
 	requests, err := certClient.CertificateSigningRequests().List(ctx, metav1.ListOptions{
-		Limit: csrGatherLimit,
+		// limit the maximum number of crs that
+		// will be listed in a single request to reduce memory usage.
+		Limit: 5000,
```
rluders, line 64: Another one that used to be a constant declared at the package global scoped and was only being used here.

pkg/gatherers/clusterconfig/install_plans.go
```diff
@@ -19,36 +19,47 @@ import (
...
 	Namespace string
 	Name      string
 	CSV       string
 	Count     int
 }
 
-// InstallPlanAnonymizer implements serialization of top x installplans
-type InstallPlanAnonymizer struct {
+// installPlanAnonymizer implements serialization of top x installplans
+type installPlanAnonymizer struct {
```
rluders, line 30: It didn't need to be exported.

pkg/gatherers/clusterconfig/jaeger_cr.go
```diff
@@ -38,6 +52,8 @@ func gatherJaegerCR(ctx context.Context, dynamicClient dynamic.Interface) ([]rec
 		return nil, []error{err}
 	}
 	var errs []error
+	// Limit the number of gathered jaegers.jaegertracing.io resources
+	var limit = 5
```
rluders, line 56: This `limit` variable was declared at the package global scope and it could be changed anywhere, shadowing some variable in other scopes. Since it was just used here, I moved it to the right scope.

pkg/gatherers/workloads/workloads_info.go
```diff
@@ -38,13 +38,31 @@ const (
 	podsLimit = 8000
```
rluders, line 38: I just want to add a note here to say that, this gatherer has some constants and global variables that I would LOVE to remove, but it seems a little bit complicated to include on this PR. If I decide to change it, I'll submit it as a separate PR.

ncaak (approved):
Mostly changes in the documentation (kudos, as this is tedious work).

I agree with the changes on scopes: package consts -> function vars. In extent packages these constants should be on a centralized file. But in this case, as they only apply to one function/file, this approach is correct IMHO.

pkg/gatherers/clusterconfig/image_registries.go
```diff
@@ -157,6 +155,7 @@ func anonymizeImageRegistry(config *registryv1.Config) *registryv1.Config {
 
 	// kubectl.kubernetes.io/last-applied-configuration annotation contains complete previous resource definition
 	// including the sensitive information as bucket, keyIDs, etc.
+	var lacAnnotation = "kubectl.kubernetes.io/last-applied-configuration"
```
rluders, line 158: Removing variable from package scope.

pkg/gatherers/clusterconfig/olm_operators.go
```diff
@@ -7,21 +7,12 @@ import (
...
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 
 	"github.com/openshift/insights-operator/pkg/record"
 	"github.com/openshift/insights-operator/pkg/utils"
 )
 
-var (
```
rluders, line 17: moved to const.go

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -40,27 +40,43 @@ type CompactedEventList struct {
 	Items []CompactedEvent `json:"items"`
 }
 
-// Used to detect the possible stack trace on logs
-var stackTraceRegex = regexp.MustCompile(`\.go:\d+\s\+0x`)
```
rluders, line 44: I moved this to the local scope.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -40,27 +40,43 @@ type CompactedEventList struct {
...
-//   - 4.7+
+// ### API Reference
+// None
+//
+// ### Sample data
+// - docs/insights-archive-sample/pod
+// - docs/insights-archive-sample/events
+//
+// ### Location in archive
+// | Version   | Path																    |
```
rluders, line 58: This one generates multiple files, I tried to give it some sense, but we should split it into separate gatherers in the future.

pkg/gatherers/clusterconfig/sap_pods.go
```diff
@@ -16,23 +16,35 @@ import (
...
+// - 4.8.2
+//
+// ### Backported versions
+// - 4.7.5+
+// - 4.6.25+
+//
+// ### Notes
+// - Only pods with a failing status are collected.
+// - Failed pods belonging to a job that has later succeeded are ignored.
+// **Conditional data**: This data is collected only if the `installers.datahub.sap.com` resource is found in the cluster.
```
rluders, line 47: This is a little bit confusing. Any resource is only included if it exists in the cluster. Isn't?

pkg/gatherers/clusterconfig/sap_vsystem_iptables_logs.go
```diff
@@ -15,26 +15,34 @@ import (
...
-// * Since versions:
-//   - 4.6.25+
-//   - 4.7.5+
-//   - 4.8+
+// ### Backported versions
+// - 4.7.5+
+// - 4.6.25+
+//
+// ### Notes
+// **Conditional data**: This data is collected only if the `installers.datahub.sap.com` resource is found in the cluster.
```
rluders, line 45: Same thing here. It isn't true for all resources? Do we really need to have it here?

pkg/gatherers/clusterconfig/sap_pods.go
tremes: Yes. `installers.datahub.sap.com` does exist in a cluster when SAP is installed AFAIK.

pkg/gatherers/clusterconfig/ceph_cluster.go
```diff
@@ -16,27 +16,32 @@ import (
...
+// `clusterconfig/ceph_cluster`
 //
-// ### Since versions:
-// * 4.8.49
-// * 4.9.48
-// * 4.10.31
-// * 4.11.2
-// * 4.12
+// ### Released version
+// - 4.12
```
tremes, line 35: I already mentioned that distinguishing released version and backported version doesn't bring any value IMHO. Anyway I think we should be explicit here and say `4.12.0` or use the previous format of `4.12+`

pkg/gatherers/clusterconfig/host_subnets.go
```diff
@@ -11,17 +11,32 @@ import (
 	"github.com/openshift/insights-operator/pkg/record"
 )
 
-// GatherHostSubnet collects HostSubnet information
+// GatherHostSubnet Collects HostSubnet information.
```
tremes, line 14: I don't understand this change TBH. Why capital letter? I agree it's a sentence so I would say:
`GatherHostSubnet collects HostSubnet information.`

pkg/gatherers/clusterconfig/sap_vsystem_iptables_logs.go
tremes: I think it's an important information or not?

pkg/gatherers/clusterconfig/operators_pods_and_events.go
tremes: Hm so now the regex will be compiled multiple times instead and not just once.

pkg/gatherers/clusterconfig/host_subnets.go
rluders: Check how it looks at the document: https://github.com/openshift/insights-operator/blob/8af0f63eb6941ced61b925a59b43ab7038bcf0cf/docs/gathered-data.md

rluders: Basically, when generating the documentation the name of the gather is a title and the description paragraph starts with `Collect...`

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: Wow! Wow! Wow! This is true. Thanks! I'll revert this change.

pkg/gatherers/clusterconfig/sap_vsystem_iptables_logs.go
rluders: Well... It is... I guess, but... why don't we have it in any other gatherer? Don't we have any resources that could exist only if the resource is installed in the cluster? 🤔

pkg/gatherers/clusterconfig/ceph_cluster.go
rluders: And I do prefer to have the released version explicit. Already pointed out the reasons... However, I do agree that we could change it to `4.12+` or `4.12.0`.

pkg/gatherers/clusterconfig/sap_vsystem_iptables_logs.go
tremes: Ahh now I see what you mean. This would mean to introduce a new type of condition (something like ResourceAvailable), right? Not sure it's worth investigating....we can create a spike for that.

pkg/gatherers/clusterconfig/host_subnets.go
tremes: Ahh I see. OK thanks.

pkg/gatherers/clusterconfig/sap_vsystem_iptables_logs.go
jholecek-rh: I don't think this is useful for rules as long as "no gatherer errors in gathers.json" and "no data in the archive" means "no resources on the cluster". Rules that need some resources will just get skipped if an archive wouldn't contain any.

jholecek-rh: Yes, I think so.

docs/gathered-data.md
```diff
@@ -314,9 +314,6 @@ None
 ### Notes
 PersistentVolumes was included at `4.7.0` and backported to `4.6.20+`.
 
-**Conditional data**: If the Image Registry configuration uses any PersistentVolumeClaim for the storage,
-the corresponding PersistentVolume definition is gathered
```
tremes: I still think this is important information about the gatherer so I don't know why we are removing?

tremes: Thanks

pkg/gatherers/clusterconfig/certificate_signing_requests.go
```diff
@@ -20,27 +20,33 @@ import (
...
+// - https://docs.openshift.com/container-platform/4.3/rest_api/index.html#certificatesigningrequestlist-v1beta1certificates
+//
+// ### Sample data
+// - docs/insights-archive-sample/config/storage/openshift-storage/ocs-storagecluster-cephcluster.json
 //
-// The Kubernetes api:
+// ### Location in archive
+// | Version   | Path														|
+// | --------- | --------------------------------------------------------	|
+// | >= 4.5.0  | config/certificatesigningrequests 					        |
```
jholecek-rh, line 36: This is slightly confusing. Was the gatherer available in 4.3.26 or not? This table says it wasn't but the "backports" section says it was. Which is true?

It took me some time to realize that you included the version here to cover file path changes, e.g. as in `pkg/gatherers/clusterconfig/config_maps.go`. The meaning is kind of clear there when there are multiple file paths with different versions. The simple greater than/smaller than ranges are probably good enough for that purpose because file path changes are unlikely to be backported.

@shzhou12 What do you think? Is this good enough or we want to try to find something better? 

I don't have a really better solution. Some other options could be:

* Use the table only when there have been some file path changes
* Use "all applicable" or similar if the file path hasn't changed
* Use the version of the farthest backport, i.e. `>=4.3.25` in this case

docs/gathered-data.md
```diff
@@ -13,1039 +13,1984 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+None
 
-* Location in archive: config/clusteroperator/imageregistry.operator.openshift.io/config/cluster.json
-* Location in older versions: config/imageregistry.json
-* Id in config: clusterconfig/image_registries
-* Since versions:
-  - 4.3.40+
-  - 4.4.12+
-  - 4.5+
+### Sample data
```
jholecek-rh, line 294: @rluders Having links to sample data is a great improvement. Thank you!

docs/gathered-data.md
```diff
@@ -13,1039 +13,1984 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+### Released version
+- 4.3.0
+
+### Backported versions
+- 4.6.35+
+- 4.7.11+
+
+### Notes
+This used to be part of `ClusterOperators`, it was split generating this gather at `4.8.2`.
+It used to be Specs for `ClusterOperators`, and it was initially included at `4.3` and backported to `4.2.10`.
```
jholecek-rh, line 473: Am I supposed to understand this sentence? Beacuse I don't :)

pkg/gatherers/clusterconfig/ceph_cluster.go
```diff
@@ -16,27 +16,32 @@ import (
...
-// ### Location in archive:
-// | Version | Path													 |
-// | ------- | ----------------------------------------------------- |
-// | <= 4.12 | config/storage/{namespace}/{name}.json 				 |
-// | 4.13+   | config/storage/{namespace}/cephclusters/{name}.json 	 |
+// ### Location in archive
+// | Version   | Path													 |
+// | --------- | ----------------------------------------------------- |
+// | <= 4.12.0 | config/storage/{namespace}/{name}.json 				 |
+// | 4.13.0+   | config/storage/{namespace}/cephclusters/{name}.json 	 |
```
tremes, line 29: So what would be the path for e.g 4.12.4 ?

pkg/gatherers/clusterconfig/container_images.go
```diff
@@ -17,28 +17,31 @@ import (
...
+// | >= 4.6.0  | config/running_containers.json 							|
+//
+// ### Config ID
+// `clusterconfig/container_images`
+//
+// ### Released version
+// - 4.6.0
+//
+// ### Backported versions
+// None
```
tremes, line 41: This seems to be wrong, right? There was:
```
// * Since versions:
//   - 4.5.33+
//   - 4.6.16+
//   - 4.7+
```

docs/gathered-data.md
```diff
@@ -13,1039 +13,1984 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+- 4.2.0
+
+### Backported versions
+None
+
+### Notes
+The `ClusterOperators` were used to also collect `pods` and `events`, it changed at the `4.8.2` release. The `pods`
+and `events` gathering were introduced in `4.3.0` and backported to `4.2.10`.
+
+* Spec config for CO resources was introduced at `4.7.0` and backported to`4.6.16+`
```
jholecek-rh, line 507: This is becoming a little chaotic. How about structuring this as a changelog with one entry per change? Something like

```
### Changes

* The gathering of `pods` and `events` was moved to a separate gatherer, `ClusterOperatorPodsAndEvents`, in `4.8.2`.
* Spec config for CO resources was introduced at `4.7.0` and backported to `4.6.16+`
* The `pods`
and `events` gathering were introduced in `4.3.0` and backported to `4.2.10`.
```

pkg/gatherers/clusterconfig/image_pruners.go
```diff
@@ -14,11 +14,31 @@ import (
...
+// None
+//
+// ### Sample data
+// - docs/insights-archive-sample/clusteroperator/imageregistry.operator.openshift.io/imagepruner/cluster.json
+//
+// ### Location in archive
+// | Version   | Path																			     |
+// | --------- | ----------------------------------------------------------------------------------- |
+// | <= 4.5.0  | config/imagepruner.json														 	 |
+// | >= 4.6.0  | config/clusteroperator/{group}/{kind}/{name}.json |
```
tremes, line 29: Again what would be the location for e.g 4.5.1 ?

docs/gathered-data.md
rluders: LOL... I'll try to put it in better words.

pkg/gatherers/clusterconfig/version.go
```diff
@@ -20,18 +20,37 @@ import (
...
+// - docs/insights-archive-sample/config/version.json
+// - docs/insights-archive-sample/config/pod
+// - docs/insights-archive-sample/events/
+// - docs/insights-archive-sample/config/id
+//
+// ### Location in archive
+// | Version   | Path														|
+// | --------- | ---------------------------------------------------------- |
+// | >= 4.2.0  | config/version.json										|
+// | >- 4.2.0  | config/id													|
```
tremes, line 39: probably typo `>-`

pkg/gatherers/clusterconfig/certificate_signing_requests.go
rluders: Correct. I'm using it to track the path changes. I also questioned myself if the version should be the lowest backported version, but then I found out that we could have some weird cases. For example:

Some feature was released to `4.7.0` and backported to `4.6.14` and `4.5.22`. How should I mention it in the table? 'Cause if I use `>= 4.5.22` it means that `4.6.0` will also have the file, which isn't true, since for `4.6` it was only backported to `4.6.15+`.

I couldn't come up with a better solution for this. I'm not sure if `all applicable` or not using the table will help as well. I mean, it could help till a certain point, but it will not solve the problem, IMHO, until we include some path change to the gatherer and backport it then we are back to the problem.

Maybe by adding a `note` before the table saying that, if not specified in the table the backport uses the same path as the release version? 🤔

jholecek-rh: When the table uses the "released version", there is a **gap**. The documentation doesn't say what the location is in backported versions. The reader is left to realize that the table states the "released version" and assume that backports used the same locations.

When we used the lowest backport, you are correct that there might be a **perceived contradiction** between the table and info about availability. The reader needs to know/realize that the table describes only file locations and not availability. Once they realize that, they don't have to make any further assumptions.

For me personally, the second case is easier to understand, but I acknowledge that this might be a matter of personal taste. :shrug:

rluders: I don't have any strong opinion about it. I think that both work... but I'll not discard the need for the note to explain how to read the table data. I prefer to not let the user make assumptions.

@tremes WDYT?

pkg/gatherers/clusterconfig/certificate_signing_requests.go
```diff
@@ -20,27 +20,33 @@ import (
...
-// Collects CSRs which werent Verified, or when Now < ValidBefore or Now > ValidAfter
+// GatherCertificateSigningRequests Collects anonymized CertificateSigningRequests which weren't Verified, or
+// when Now < ValidBefore or Now > ValidAfter
+//
+// ### API Reference
+// - https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/certificates/v1beta1/certificatesigningrequest.go#L78
+// - https://docs.openshift.com/container-platform/4.3/rest_api/index.html#certificatesigningrequestlist-v1beta1certificates
+//
+// ### Sample data
+// - docs/insights-archive-sample/config/storage/openshift-storage/ocs-storagecluster-cephcluster.json
```
shzhou12, line 31: A tiny thing here, the sample data is not correct.

pkg/gatherers/clusterconfig/certificate_signing_requests.go
shzhou12: @jholecek-rh I think rule developers could easily understand it even not adding backported version here. It is better to add some explanation somewhere.

pkg/gatherers/clusterconfig/ceph_cluster.go
shzhou12: It is also helpful if rule developers could know the backport version of the path change.

pkg/gatherers/clusterconfig/certificate_signing_requests.go
tremes: I agree that user/reader should not guess and I think I agree with what Jan mentioned here. Perhaps we could provide the location with versions only in the case there have been any updates in the location.

pkg/gatherers/clusterconfig/ceph_cluster.go
rluders: @shzhou12 not sure if I understood it. What do you mean by that?

Are you saying that, if we backport the changes from `4.13.0` that changes to `4.12.40`, and introduce the change of path by doing it we should also add it as a note? If yes, in this case, we just need to change the version in the table that points to a new path (line 29) to be `>= 4.12.40`

pkg/gatherers/clusterconfig/container_images.go
rluders: Yeah. You are correct. The only thing is that it seems that it was backported to `4.6.1+` and not to `4.6.16+`.

-  4.5.33: [RN](https://access.redhat.com/errata/RHSA-2021:0428) [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1914255)
- 4.6.1: [RN](https://access.redhat.com/errata/RHBA-2020:4196) [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1860032)

pkg/gatherers/clusterconfig/certificate_signing_requests.go
rluders: Some suggestions:

### Location in archive ( 👍 )
| Version   | Path											|
| --------- | --------------------------------------------------------	|
| any          | config/running_containers.json 					        |

### Location in archive ( 👎 )
| Version   | Path											|
| --------- | --------------------------------------------------------	|
| *               | config/running_containers.json 					        |

### Location in archive ( 😄 )
`config/running_containers.json`

@tremes @jholecek-rh @shzhou12 vote (by reacting) or suggest something else.

shzhou12: I prefer the third one. It is simple and has no ambiguity when understanding the version range.

pkg/gatherers/clusterconfig/ceph_cluster.go
shzhou12: I mean this. The pattern will be more complicated if the file path changes backported to multiple versions. If we do not backport file path changes, please ignore my comment.

rluders: Well, I can't assure you that we will not backport path changes... So, I guess that now I can see the problem here.
  
Let's imagine that we have a feature that is currently changing the path, and at first glance we considered:

| Version   | Path  |
| --------- | ------- |
| < 4.13.0 | config/old/path/file.json 	|
| >= 4.13.0 | config/new/path/file.json	 |

Later then, we decide to backport this change to `4.12.4` and `4.11.10`, and we update the table:

| Version   | Path  |
| --------- | ------- |
| < 4.11.10 | config/old/path/file.json 	|
| >= 4.11.10 | config/new/path/file.json	 |

In this case, we do have a problem. Mostly 'cause we don't have the change for some `>= 4.12.0 < 4.12.4`. So, basically, it will not work.

Even my idea to add a note, will not help... So, we will probably have to specify, in this case, every single backported version that is affected by the change in the table.

| Version   | Path  |
| --------- | ------- |
| < 4.11.10 | config/old/path/file.json 	|
| 4.11.10+ | config/new/path/file.json	 |
| >= 4.12.4 | config/new/path/file.json	 |

This probably overlaps with the discussion that we are having [here](https://github.com/openshift/insights-operator/pull/704#discussion_r1067072076)

rluders: OK, seems that mixing `<= < > >=` with `+` isn't something nice. So, in the best-case scenario, we will have to specify in the table any change of path that could happen in backports, to leave it as clear as possible. I'll review all and update the docs again.

pkg/gatherers/clusterconfig/certificate_signing_requests.go
rluders: Well, folks, I updated the information. Hopefully, it is better now.

docs/gathered-data.md
rluders: I tried to improve this statement in the current docs.

ncaak (approved):

pkg/gatherers/clusterconfig/config_maps.go
```diff
@@ -15,32 +15,42 @@ import (
...
+// ### Config ID
+// `clusterconfig/config_maps`
 //
-// * "cluster-config-v1" ConfigMap since versions:
-//   - 4.9+
+// ### Released version
+// - 4.5.0
+//
+// ### Backported versions
+// - 4.3.23+
```
tremes, line 41: This should be `4.3.25+` right? When looking at the original versions.

pkg/gatherers/clusterconfig/container_images.go
```diff
@@ -17,28 +17,30 @@ import (
...
+//
+// ### Config ID
+// `clusterconfig/container_images`
+//
+// ### Released version
+// - 4.7.0
+//
+// ### Backported versions
+// - 4.5.33+
+// - 4.6.1+
```
tremes, line 40: This should be `4.6.16+` version, right? When looking at the original versions.

pkg/gatherers/clusterconfig/machine_autoscaler.go
```diff
@@ -11,20 +11,29 @@ import (
...
-//	https://docs.openshift.com/container-platform/4.7/rest_api/autoscale_apis/machineautoscaler-autoscaling-openshift-io-v1beta1.html#machineautoscaler-autoscaling-openshift-io-v1beta1
+// ### Config ID
+// `clusterconfig/machine_autoscalers`
 //
-// * Location in archive: config/machineautoscalers/{namespace}/{machineautoscaler-name}.json
-// * Id in config: clusterconfig/machine_autoscalers
-// * Since versions:
-//   - 4.8+
+// ### Released version
+// - 4.8.2
```
tremes, line 30: This should be probably `4.8.0`

pkg/gatherers/clusterconfig/container_images.go
rluders: 4.5.33: [RN](https://access.redhat.com/errata/RHSA-2021:0428) [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1914255)
4.6.1: [RN](https://access.redhat.com/errata/RHBA-2020:4196) [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1860032)

pkg/gatherers/clusterconfig/machine_config_pools.go
```diff
@@ -13,21 +13,30 @@ import (
...
+// ### Config ID
+// `clusterconfig/machine_config_pools`
 //
-// * Location in archive: config/machineconfigpools/
-// * Id in config: clusterconfig/machine_config_pools
-// * Since versions:
-//   - 4.5.33+
-//   - 4.6+
+// ### Released version
+// - 4.7.0
```
tremes, line 32: I guess this should be 4.6.0 and the below `4.6.16+` should be removed. Based on the original data.

pkg/gatherers/clusterconfig/machine_configs.go
```diff
@@ -13,17 +13,30 @@ import (
...
+// - `config/machineconfigs/{resource}.json`
+//
+// ### Config ID
+// `clusterconfig/machine_configs`
+//
+// ### Released version
+// - 4.8.2
+//
+// ### Backported versions
+// None
```
tremes, line 36: It's missing `4.8.5`

pkg/gatherers/clusterconfig/mutatingwebhookconfigurations.go
```diff
@@ -14,17 +14,30 @@ import (
...
+// - docs/insights-archive-sample/config/mutatingwebhookconfigurations
+//
+// ### Location in archive
+// - `config/mutatingwebhookconfigurations/{resource}.json`
+//
+// ### Config ID
+// `clusterconfig/mutating_webhook_configurations`
+//
+// ### Released version
+// - 4.10.13
```
tremes, line 32: Again wrong released version.

pkg/gatherers/clusterconfig/olm_operators.go
```diff
@@ -36,17 +27,32 @@ type csvRef struct {
...
+// - `config/olm_operators`
+//
+// ### Config ID
+// `clusterconfig/olm_operators`
+//
+// ### Released version
+// - 4.7.0
+//
+// ### Backported versions
+// - 4.6.26+
```
tremes, line 52: It seems this hasn't been backported.

pkg/gatherers/clusterconfig/container_images.go
tremes: OK thanks....this one is confusing though

pkg/gatherers/clusterconfig/openshift_apiserver_operator_logs.go
```diff
@@ -9,20 +9,31 @@ import (
...
+// - `config/pod/{namespace-name}/logs/{pod-name}/errors.log`
 //
-//	https://docs.openshift.com/container-platform/4.6/rest_api/workloads_apis/pod-core-v1.html#apiv1namespacesnamespacepodsnamelog
+// ### Config ID
+// `clusterconfig/openshift_apiserver_operator_logs`
 //
-// * Location in archive: config/pod/{namespace-name}/logs/{pod-name}/errors.log
-// * Id in config: clusterconfig/openshift_apiserver_operator_logs
+// ### Released version
+// - 4.7.0
```
tremes, line 30: This looks like an educated guess :)) hopefully a good one :)

pkg/gatherers/clusterconfig/openshift_sdn_logs.go
```diff
@@ -10,25 +10,33 @@ import (
...
-// * Location in archive: config/pod/openshift-sdn/logs/{pod-name}/errors.log
-// * Id in config: clusterconfig/openshift_sdn_logs
-// * Since versions:
-//   - 4.6.19+
-//   - 4.7+
+// ### Released version
+// - 4.7.0
+//
+// ### Backported versions
+// - 4.6.21+
```
tremes, line 36: Shouldn't it be `4.6.19+` ?

pkg/gatherers/clusterconfig/config_maps.go
rluders: Yes. You are right, I'll fix it. Here are the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1825758) and [RN](https://access.redhat.com/errata/RHBA-2020:2436).

pkg/gatherers/clusterconfig/machine_autoscaler.go
rluders: So, I had the same question here. But it seems that the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1966523) and the [RN](https://access.redhat.com/errata/RHSA-2021:2438) points to `4.8.2`.

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -43,24 +43,40 @@ type CompactedEventList struct {
...
+//
+// ### Released version
+// - 4.8.2
+//
+// ### Backported versions
+// - 4.6.35+
+// - 4.7.11+
+//
+// ### Changes
+// - This gather used to be part of [ClusterOperators](#ClusterOperators), it was split generating this gather at `4.8.2`.
```
tremes, line 77: Nit: I am not sure I really understand the following:

> ...,it was split generating this gather at 4.8.2.

pkg/gatherers/clusterconfig/pod_disruption_budgets.go
```diff
@@ -14,18 +14,30 @@ const (
...
+//
+// ### Config ID
+// `clusterconfig/pdbs`
+//
+// ### Released version
+// - 4.6.0
+//
+// ### Backported versions
+// - 4.5.15+
+// - 4.4.30+
```
tremes, line 37: Isn't there an unwritten rule that this is ordered by starting the oldest version? And the `4.5.15+` doesn't correspond to the original `4.5.34+`

pkg/gatherers/clusterconfig/pod_network_connectivity_checks.go
```diff
@@ -11,17 +11,31 @@ import (
...
+// - docs/insights-archive-sample/config/podnetworkconnectivitychecks.json
+//
+// ### Location in archive
+// - `config/podnetworkconnectivitychecks.json`
+//
+// ### Config ID
+// `clusterconfig/pod_network_connectivity_checks`
+//
+// ### Released version
+// - 4.8.2
```
tremes, line 32: I've seen this a couple of times here and it's probably not a big deal, but I am wondering about the `4.8.0` vs `4.8.2`

pkg/gatherers/clusterconfig/sap_pods.go
```diff
@@ -16,23 +16,34 @@ import (
...
+//
+// ### Config ID
+// `clusterconfig/sap_pods`
+//
+// ### Released version
+// - 4.8.2
+//
+// ### Backported versions
+// - 4.7.5+
+// - 4.6.25+
```
tremes, line 43: Is this wrong or was it originally wrong? There was `4.6.24+`

pkg/gatherers/clusterconfig/validatingwebhookconfigurations.go
```diff
@@ -14,17 +14,30 @@ import (
...
+// ### Config ID
+// `clusterconfig/validating_webhook_configurations`
+//
+// ### Released version
+// - 4.10.13
+//
+// ### Backported versions
+// - 4.9.11+
+// - 4.8.24+
+// - 4.7.40+
```
tremes, line 37: Different ordering.

pkg/gatherers/clusterconfig/validatingwebhookconfigurations.go
```diff
@@ -14,17 +14,30 @@ import (
...
+// - docs/insights-archive-sample/config/validatingwebhookconfigurations/
+//
+// ### Location in archive
+// - `config/validatingwebhookconfigurations/{name}.json`
+//
+// ### Config ID
+// `clusterconfig/validating_webhook_configurations`
+//
+// ### Released version
+// - 4.10.13
```
tremes, line 32: This should be `4.10.3` according to [errate](https://errata.devel.redhat.com/advisory/86559)

pkg/gatherers/clusterconfig/machine_config_pools.go
tremes: OK then. THanks

pkg/gatherers/clusterconfig/machine_configs.go
```diff
@@ -13,17 +13,30 @@ import (
...
+// - docs/insights-archive-sample/config/machineconfigs/75-worker-sap-data-intelligence.json
+//
+// ### Location in archive
+// - `config/machineconfigs/{resource}.json`
+//
+// ### Config ID
+// `clusterconfig/machine_configs`
+//
+// ### Released version
+// - 4.8.2
```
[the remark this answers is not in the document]
rluders: Same case that the one for `machine_autoscaler`.

pkg/gatherers/clusterconfig/machine_configs.go
rluders: Do you have the PR, BZ, or RN that confirms it?

pkg/gatherers/clusterconfig/mutatingwebhookconfigurations.go
rluders: Fixing. It was `4.10.3` [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2027629) [RN](https://access.redhat.com/errata/RHSA-2022:0056)

pkg/gatherers/clusterconfig/machine_configs.go
tremes: https://bugzilla.redhat.com/show_bug.cgi?id=1977754

tremes: It seems the release version was 4.9.0 in fact https://errata.devel.redhat.com/advisory/81213

pkg/gatherers/clusterconfig/olm_operators.go
rluders: I have a different information based on [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1935574) and [RN](https://access.redhat.com/errata/RHBA-2021:1232)

tremes: I was probably wrong here. You can ignore.

pkg/gatherers/clusterconfig/openshift_apiserver_operator_logs.go
rluders: [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1913356) and [RN](https://access.redhat.com/errata/RHSA-2020:5633) and we don't have it on [release-4.6](https://github.com/openshift/insights-operator/tree/release-4.6/pkg/gather/clusterconfig).

pkg/gatherers/clusterconfig/openshift_sdn_logs.go
rluders: Based on? Check: [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1916843] and [RN](https://access.redhat.com/errata/RHSA-2020:5633)

pkg/gatherers/clusterconfig/operators_pods_and_events.go
rluders: I can rephrase it, but basically, `GatherClusterOperatorPodsAndEvents` used to be inside the `GatherClusterOperatorPodsAndEvents`, and it actually moved from it to its own gather at `4.8.2`.

pkg/gatherers/clusterconfig/openshift_sdn_logs.go
tremes: I think it's https://bugzilla.redhat.com/show_bug.cgi?id=1921554. Do not mix `sdn logs` and `sdn controller logs`

pkg/gatherers/clusterconfig/machine_configs.go
rluders: This is a very good piece of information. Thank you. I'll adjust.

pkg/gatherers/clusterconfig/openshift_sdn_logs.go
rluders: Thanks!

pkg/gatherers/clusterconfig/pod_disruption_budgets.go
rluders: Yes! Nice catch, thanks.

pkg/gatherers/clusterconfig/pod_network_connectivity_checks.go
rluders: Yeah, I found a lot of places where the RN points to `4.8.2` and not to `4.8.0`. If you are curious [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1949907)

pkg/gatherers/clusterconfig/sap_pods.go
rluders: Based on the information that I found, it was wrong. You can confirm: [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1935775) [RN](https://access.redhat.com/errata/RHBA-2021:1153)

pkg/gatherers/clusterconfig/service_accounts.go
```diff
@@ -16,22 +16,31 @@ import (
...
+//
+// ### Config ID
+// `clusterconfig/service_accounts`
+//
+// ### Released version
+// - 4.7.0
+//
+// ### Backported versions
+// - 4.6.20+
+// - 4.5.34+
```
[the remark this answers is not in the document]
rluders: Nice. Thank you.

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
 
 ## APIRequestCounts
 
-BuildGatherApiRequestCounts creates a gathering closure which collects API requests counts for the
-resources mentioned in the alert provided as a string parameter
+Creates a gathering closure which collects API requests counts for the
+resources mentioned in the alert provided as a string parameter.
```
jholecek-rh, line 17: We might want to look at the documentation for conditional gathering next. While this text probably makes sense as a developer documentation for the conditional gatherer, it doesn't really tell users what is actually gathered (how to find out/what to do to change what is gathered).

rluders: This is actually a really good point. I think that depending on the changes since we do have a small number of conditionals gathered I can update it with a better description of this PR.

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+- 4.12.0
+
+### Backported versions
+- 4.8.49+
+- 4.9.48+
+- 4.10.31+
+- 4.11.2+
+
+### Changes
+None
```
jholecek-rh, line 135: I would like to include include a warning about the file collisions with OpenshiftStorage/StorageCluster in OCP 4.11 and OCP 4.12. Not sure if the "Changes" section is the best place. It works for me.

Something like this would do (please check specifically that the links are correct):

```suggestion
* In OCP 4.11 and OCP 4.12, the location of gathered data collides with data gathered by the [GatherStorageCluster](#storagecluster) gatherer. It is practically impossible to tell the two resources apart. Use with caution.
```

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
-* 4.11
+### Released version
+- 4.11.0
+
+### Backported versions
+None
+
+### Changes
+- Renamed from `OpenshiftStorage` to `StorageCluster` in version `4.12.0+`
+- Config ID changed from `clusterconfig/openshift_storage` to `clusterconfig/storage_cluster` in version `4.12.0+`
```
jholecek-rh, line 1782: I would like to add a counterpart to the disclaimer for CephCluster (suggested in another comment):

```suggestion
- In OCP 4.11 and OCP 4.12, the location of gathered data collides with data gathered by the [GatherCephCluster](#cephcluster) gatherer. It is practically impossible to tell the two resources apart. Use with caution.
```

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
-  - 4.5+
+### Backported versions
+- 4.3.40+
+- 4.4.12+
 
-* PV definition since versions:
-  - 4.6.20+
-  - 4.7+
+### Changes
+- "PersistentVolumes" were introduced in version 4.7.0 and backported to 4.6.20+.
```
jholecek-rh, line 305: It hurts to see (at least) four different ways to denote gathered API objects throughout the document:

1. Name only, no extra formatting: Authentication
2. Name only, double-quoted: "PersistentVolumes"
3. Name only, mono-spaced: `PersistentVolumes`
4. Name and API group, mono-spaced: `cephclusters.ceph.rook.io`

I don't mean to bloat the scope of this PR - is it easier to unify it within this PR or create a separate one once this gets merged?

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
-* Spec config for CO resources since versions:
-  - 4.6.16+
-  - 4.7+
+### Sample data
+- [docs/insights-archive-sample/config/clusteroperator](./insights-archive-sample/config/clusteroperator)
+
+### Location in archive
+| Version   | Path														|
+| --------- | --------------------------------------------------------	|
+| >= 4.2.0  | config/clusteroperator/ 									|
```
jholecek-rh, line 471: This doesn't tell what the structure of the folder was. It is either redundant (the remaining lines cover all the locations across versions) or it is incomplete. I guess it is the former.

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+- 4.2.0
+
+### Backported versions
+None
+
+### Changes
+- `config/pod/{namespace}/{pod}.json` and `events/` were moved to
+[ClusterOperatorPodsAndEvents](#ClusterOperatorPodsAndEvents) since `4.8.2`,
+both were introduced at `4.3.0` as part of this gatherer and backported to `4.2.10+`.
+- ClusterOperator resources was included at `4.7.0` and backported to `4.6.16+`.
```
jholecek-rh, line 490: I don't understand what was added/backported. The "Locations" table suggests that the location of the ClusterOperator resources changed in 4.7.0. If the location change was backported to 4.6.16, the "Locations" table is incomplete/inaccurate.

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+- [docs/insights-archive-sample/events/](./insights-archive-sample/events/)
+- [docs/insights-archive-sample/config/id](./insights-archive-sample/config/id)
+
+### Location in archive
+| Version   | Path														|
+| --------- | ---------------------------------------------------------- |
+| >= 4.2.0  | config/version.json										|
+| >= 4.2.0  | config/id													|
+| >= 4.8.2  | config/pod/openshift-cluster-version/{name}.json			|
+| >= 4.8.2  | events/{namespace}.json									|
```
jholecek-rh, line 540: I would assume that:

* the `name` is always `version` and 
* the `namespace` is always `openshift-cluster-version`

If I am correct, it would be nice to have the specific paths here.

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+
+### Changes
+None
 
 
 ## ContainersLogs
 
-collects either current or previous containers logs for pods firing one of the configured alerts.
+Collects either current or previous containers logs for pods firing one of the
+configured alerts.
```
jholecek-rh, line 653: Are the "configured alerts" hardcoded or this is a conditional gatherer?

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+### Changes
+None
 
 
 ## InstallPlans
 
-collects Top x InstallPlans from all openshift namespaces.
-Because InstallPlans have unique generated names, it groups them by namespace and the "template"
-for name generation from field generateName.
+Collects top X InstallPlans from all openshift namespaces. Because InstallPlans have
```
jholecek-rh, line 793: Begs the question what is "X"...

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+### Changes
+None
 
 
 ## KubeControllerManagerLogs
 
-collects logs from kube-controller-manager pods in the openshift-kube-controller-manager namespace with following substrings:
-  - "Internal error occurred: error resolving resource",
-  - "syncing garbage collector with updated resources from discovery",
+Collects ogs from kube-controller-manager pods in the
```
jholecek-rh, line 848: Little typo here. I would also suggest to use the same formatting for the pod name as for the namespace.

```suggestion
Collects logs from `kube-controller-manager` pods in the
```

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+### API Reference
+- https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/apis/autoscaling/v1beta1/machineautoscaler_types.go
+- https://docs.openshift.com/container-platform/4.7/rest_api/autoscale_apis/machineautoscaler-autoscaling-openshift-io-v1beta1.html#machineautoscaler-autoscaling-openshift-io-v1beta1
+
+### Sample data
+- [docs/insights-archive-sample/config/machineautoscalers/openshift-machine-api/worker-us-east-1a.json](./insights-archive-sample/config/machineautoscalers/openshift-machine-api/worker-us-east-1a.json)
 
-	https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/apis/autoscaling/v1beta1/machineautoscaler_types.go
+### Location in archive
+- `config/machineautoscalers/{namespace}/{resource}.json`
```
jholecek-rh, line 920: Does `{resource}` mean the resource kind, resource name, resource UUID or something else?

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+None
+
+### Changes
+None
 
 
 ## Nodes
 
-collects all Nodes.
+Collects all cluster nodes.
```
jholecek-rh, line 1205: `Node` or `Nodes` is a resource kind. Better to denote it so, imho. This document uses at least four different ways to denote resource kinds (see another comment of mine), so I struggle to recommend one.

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+### API Reference
+None
+
+### Sample data
+- [docs/insights-archive-sample/pod](./insights-archive-sample/pod)
+- [docs/insights-archive-sample/events](./insights-archive-sample/events)
+
+### Location in archive
+- `config/pod/{namespace}/{pod}.json`
+- `events/`
```
jholecek-rh, line 437: What is the structure of the folder? There are other gatherers that put files to the folder. What makes sure the gatherers do not collide?

pkg/gatherers/clusterconfig/operators_pods_and_events.go
```diff
@@ -43,24 +43,40 @@ type CompactedEventList struct {
...
+// ### API Reference
+// None
+//
+// ### Sample data
+// - docs/insights-archive-sample/pod
+// - docs/insights-archive-sample/events
+//
+// ### Location in archive
+// - `config/pod/{namespace}/{pod}.json`
+// - `events/`
```
jholecek-rh, line 63: I suppose the structure of the folder is like this?

```suggestion
// - `events/{namespace}.json`
```

docs/gathered-data.md
jholecek-rh: Actually, judging by other gatherers, I assume it should look like this:

```suggestion
- `events/{namespace}.json`
```

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
-  - "Node %s may be offline... retrying": An egress node has failed the egress IP health check once,
-    so it has big chances to be marked as offline soon or, at the very least, there has been a connectivity glitch.
-  - "Node %s is offline": An egress node has failed enough probes to have been marked offline for egress IPs.
-    If it has egress CIDRs assigned, its egress IPs have been moved to other nodes.
-    Indicates issues at either the node or the network between the master and the node.
-  - "Node %s is back online": This indicates that a node has recovered from the condition described
-    at the previous message, by starting succeeding the egress IP health checks.
-    Useful just in case that previous “Node %s is offline” messages are lost,
-    so that we have a clue that there was failure previously.
+Collects logs from sdn-controller pod in openshift-sdn namespace with
```
jholecek-rh, line 1374: Pod and namespace names are often denoted using monospace font in other gatherers:

```suggestion
Collects logs from `sdn-controller` pod in `openshift-sdn` namespace with
```

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+- [docs/insights-archive-sample/config/securitycontextconstraint](./insights-archive-sample/config/securitycontextconstraint)
+- [docs/insights-archive-sample/config/clusterrolebinding](./insights-archive-sample/config/clusterrolebinding)
 
-* Location in archive: config/securitycontentconstraint/, config/clusterrolebinding/
-* Id in config: clusterconfig/sap_config
-* Since versions:
-  - 4.6.20+
-  - 4.7+
+### Location in archive
+- `config/{securitycontentconstraint|clusterrolebinding}/{name}.json`
```
jholecek-rh, line 1540: Could we avoid unnecessary complexity in path templates. It is hard enough to figure out what the paths are if there have been any changes.

```suggestion
- `config/clusterrolebinding/{name}.json`
- `config/securitycontentconstraint/{name}.json`
```

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
+- https://pkg.go.dev/k8s.io/client-go/kubernetes/typed/batch/v1
+- https://pkg.go.dev/k8s.io/client-go/dynamic
+
+### Sample data
+None
+
+### Location in archive
+- `config/pod/{namespace}/{name}.json`
 
-**Conditional data**: This data is collected only if the "installers.datahub.sap.com" resource is found in the cluster.
```
jholecek-rh, line 881: I thought we eventually agreed on keeping this line. It is important because "the gatherer completed successfully" and "there are no data in the archive" **does not imply** "there were no resources on the cluster".

docs/gathered-data.md
```diff
@@ -13,1040 +13,1878 @@ archives are packaged as `.tar.gz` files in `/var/lib/insights-operator`.
...
 
 
 ## SAPVsystemIptablesLogs
 
-collects logs from SAP vsystem-iptables containers
+Collects logs from SAP `vsystem-iptables` containers
 including one from license management pods with the following substring:
   - "can't initialize iptables table",
 
-**Conditional data**: This data is collected only if the "installers.datahub.sap.com" resource is found in the cluster.
```
jholecek-rh, line 902: Same as above. I thought we eventually agreed on keeping this line.

docs/gathered-data.md
tremes: And it should be `OpenShift` when I am seeing it....

rluders: I guess that the original idea for the person that wrote it was to use `X`, since the value of the limit of items could change in code, but... it isn't helpful when we are reading.

For the `Openshift` I guess that is better if we use `openshift-* namespace`, like:

https://github.com/openshift/insights-operator/blob/1f10ef9673a311c232d51aa68dbac31cfb56e65b/pkg/gatherers/clusterconfig/install_plans.go#L81

docs/gathered-data.md
rluders: I'll try to review it, but it is hard to get through all this stuff. There is no tool, you need to read very, very, very carefully and there is no tooling to help it, except your eyes.

docs/gathered-data.md
rluders: @jholecek-rh when you said: "What makes sure the gatherers do not collide?" - I was pointing this for @tremes on our channel, probably it will not be something frequent, but, IMHO, it is just a guest, we probably need to investigate it and make sure that collisions are not happening. There are other gathers that share the same path in the archive.

PS: this is just a note, I don't expect interaction about path collisions here, 'cause this is something that needs to be addressed in a separate task.

docs/gathered-data.md
rluders: Alerts use to be in our code, but now they are fetched from the service. I'll update it.

pkg/gatherers/clusterconfig/pod_network_connectivity_checks.go
jholecek-rh: Is it possible that `4.8.2` was the first GA/public release of 4.8?

pkg/gatherers/clusterconfig/operators_pods_and_events.go
jholecek-rh: Offering a more humane wording, although a bit more verbose:

```
The data gathered by `GatherClusterOperatorPodsAndEvents` were originally gathered by [`ClusterOperators`](#ClusterOperators). The [`ClusterOperators`](#ClusterOperators) gather was split at 4.8.2 and the change was backported to 4.7.11 and 4.6.35.
```

pkg/gatherers/clusterconfig/pod_network_connectivity_checks.go
rluders: Yep. It was.

docs/gathered-data.md
rluders: This line should not exist. It should be just the one with `< 4.7.0`. Thank you!

rluders:
🤞 
/retest

tremes:
These are mostly documentation changes. Overriding the upgrade test to unblock other PRs.

PR #706, opened 2023-01-03
OCPBUGS-5257: do not get disabled rules

tremes:
There are two ways to disable a rule. First is to do that for particular cluster. This would work with this original query parameter. Second is to do that for your organization (multiple clusters) and this is not currently supported by the API (and the query parameter).

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Privacy

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browseOCPBUGS-5257 


rluders (approved):
/lgtm

PR #709, opened 2023-01-04
OCPBUGS-5347: do not periodically update Available clusteroperator co…

tremes:
…ndition in disabled state

It doesn't make any sense to set some other conditions in disabled state, because it's still "healthy" state from operator perspective. Temporarily setting a different  `Available` condition status caused the "since" time reset every 2 mins and unnecessarily updated the transition time of the condition:

```
oc get co insights --watch
NAME       VERSION       AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
insights   4.13.0-ec.1   True        False         False      62s     
insights   4.13.0-ec.1   True        False         False      1s      
insights   4.13.0-ec.1   True        False         False      0s      
insights   4.13.0-ec.1   True        False         False      1s      
insights   4.13.0-ec.1   True        False         False      0s      
insights   4.13.0-ec.1   True        False         False      0s      
```

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

no new data

## Unit Tests

extended 
- `pkg/controller/status/controller_test.go`

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/OCPBUGS-5347

### DIFF
--- a/pkg/controller/status/conditions.go
+++ b/pkg/controller/status/conditions.go
@@ -70,11 +70,12 @@ func newConditions(cos *configv1.ClusterOperatorStatus, time metav1.Time) *condi
 }
 
 func (c *conditions) setCondition(conditionType configv1.ClusterStatusConditionType,
-	status configv1.ConditionStatus, reason, message string, lastTime metav1.Time) {
+	status configv1.ConditionStatus, reason, message string) {
 	originalCondition, ok := c.entryMap[conditionType]
+	transitionTime := metav1.Now()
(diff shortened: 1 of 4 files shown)


rluders (approved):
Yep, it makes sense.
/lgtm

tremes:
I found out that this is still not 100% in case when operator is degraded and then marked as disabled.
/hold

PR #714, opened 2023-01-10
OCPBUGS-5903: operators gatherer - handle ingresscontroller relatedObject & simplify

tremes:
the code

This adds new `/config/clusteroperator/operator.openshift.io/ingresscontroller/openshift-ingress-operator/default.json` file to the Insights archive. It also makes the code simple and improves the related test and test coverage.

## Categories

- [ ] Bugfix
- [X] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/config/clusteroperator/operator.openshift.io/ingresscontroller/openshift-ingress-operator/default.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `pkg/gatherers/clusterconfig/operators_test.go`

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/gatherers/clusterconfig/operators.go
+++ b/pkg/gatherers/clusterconfig/operators.go
@@ -32,9 +32,11 @@ type clusterOperatorResource struct {
 	Kind       string      `json:"kind"`
 	Name       string      `json:"name"`
 	Spec       interface{} `json:"spec"`
+	namespace  string
 }
 
-// GatherClusterOperators collects all the ClusterOperators definitions and their resources.
+// GatherClusterOperators collects all the ClusterOperators definitions and their related resources
+// from the `operator.openshift.io` group.
 //
 // The Kubernetes api https://github.com/openshift/client-go/blob/master/config/clientset/versioned/typed/config/v1/clusteroperator.go#L62
 // Response see https://docs.openshift.com/container-platform/4.3/rest_api/index.html#clusteroperatorlist-v1config-openshift-io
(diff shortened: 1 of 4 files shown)


pkg/gatherers/clusterconfig/operators.go
```diff
@@ -142,61 +153,66 @@ func collectClusterOperatorResources(ctx context.Context,
...
+			Resource: ro.Resource,
 		}
+		version := resVer[gr]
+		gvr := gr.WithVersion(version)
+		clusterResource, err := getRelatedObjectResource(ctx, dynamicClient, ro, co.Name, gvr)
+		if err != nil {
+			klog.V(2).Infof("Unable to get %s resource due to: %s", fmt.Sprintf("%s.%s", ro.Resource, ro.Group), err)
+			continue
+		}
+		spec, ok := clusterResource.Object["spec"]
```
ncaak, line 167: When I use this way of checking if a key exists in a map, I usually prefer to rename the second variable `exists` rather than `ok`. That way next line will display `if !exists {` what I find more readable.

However I saw this check several times here and on other repos. Out of curiosity, is this some kind of convention?

ncaak: BTW, not a change request or something like that. Just asking to feed my knowledge daemon.

tremes: Thanks. I don't have strong opinion about this to be honest. I don't think it's really a convention. We can go or have `exists` as well.

rluders (approved):
/lgtm

pkg/gatherers/clusterconfig/operators.go
rluders: I think that both works fine. AFAIK, in Go, the most common is to use `ok`, but not a strong opinion, if someone use `exists` instead of `ok` I'll be fine with it as well, once the verification happens in the following line, it is easy to read anyway.

JoaoFula:
retesting IO e2e tests since they've been updated to have the new file taken into account.

tremes:
@JoaoFula `insights-operator-e2e-tests` will always fail here, because the `test_running_gather_as_job` uses `quay.io/openshift/origin-base:latest` image for the testing, which doesn't contain these very last changes of course. 

I think I will override the tests for now and we should update the integration tests to use the same image for the gathering job as the given cluster uses for the Insights operator.

PR #721, opened 2023-01-25
arch docs update - explain disabled=true status more

tremes:
## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

no new data

## Privacy

## Breaking Changes

No

## References


docs/arch.md
```diff
@@ -244,7 +244,9 @@ oc get co insights -o=json | jq '.status.conditions'
 A condition is defined by its type. You may notice that there are some non-standard clusteroperator conditions. They are:
 - `SCAAvailable` - based on the SCA (Simple Content Access) controller in `pkg/ocm/sca/sca.go` and provides information about the status of downloading the SCA entitlements.
 - `ClusterTransferAvailable` - based on the cluster transfer controller in `pkg/ocm/clustertransfer/cluster_transfer.go` and provides information about the availability of cluster transfers.
-- `Disabled` - indicates whether data gathering is disabled or enabled.
+- `Disabled` - indicates whether data gathering is disabled or enabled. Note that when the operator is `Disabled=True`, it is still also `Available=True`, which is weird at first glance, but the Cluster Version Operator (CVO) checks that all the clusteroperators are `Available=True` during the OpenShift installation. If they are not, the installation will fail, which happens in disconnected environments/clusters where the Insights operator is usually `Disabled=True` (because there is no `cloud.openshift.com` token in the `pull-secret`). You can find more about this topic in:
```
JoaoFula, line 247: Is there any situation where Available=False and Disabled=True? Or are we forcing Available=True when Disabled=True to avoid the situation where the disable the operator while it's degraded?

Also, wouldn't it be better to have "which seems unusual" or "which is strange at a first glance"
"weird" just seems kinda strong IMO

tremes: >Is there any situation where Available=False and Disabled=True? Or are we forcing Available=True when Disabled=True to avoid the situation where the disable the operator while it's degraded?

I think there isn't, because right now when `Disabled=True` then the operator does nothing, but it's still running and available

> Also, wouldn't it be better to have "which seems unusual" or "which is strange at a first glance"
"weird" just seems kinda strong IMO

I am fine with strange word :) Thanks

JoaoFula: Alright, seems good to me then :)

tremes:
Minor docs enhancement.
/label px-approved
/label docs-approved

PR #723, opened 2023-01-26
OCPBUGS-6731: Anonymize env vars from containers: HTTP_PROXY, HTTPS_PROXY

ncaak:
This PR adds new functionality for container_images gatherer, where it obfuscates HTTP_PROXY and HTTPS_PROXY env variables from retrieved containers.

## Categories

- [x] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

N/A

## Documentation

N/A

## Unit Tests

- `/pkg/gatherers/clusterconfig/container_images_test.go`

## Privacy

## Changelog

Yes

## Breaking Changes

No

## References

[CCXDEV-10028](https://issues.redhat.com/browse/CCXDEV-10028)


pkg/gatherers/clusterconfig/container_images.go
```diff
@@ -68,6 +69,8 @@ func gatherContainerImages(ctx context.Context, coreClient corev1client.CoreV1In
 		for podIndex, pod := range pods.Items { //nolint:gocritic
 			podPtr := &pods.Items[podIndex]
 			if strings.HasPrefix(pod.Namespace, "openshift-") && check.HasContainerInCrashloop(podPtr) {
+				obfuscateContainerEnvVars(podPtr.Spec.Containers)
```
tremes, line 72: We will need to use this new function on more places - e.g somewhere here https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/clusterconfig/operators_pods_and_events.go#L218

pkg/gatherers/clusterconfig/container_images.go
```diff
@@ -202,3 +205,18 @@ func gatherImages(startMonth string, img2month2count img2Month2CountMap, contain
 		}
 	}
 }
+
+// obfuscateContainerEnvVars finds env variables within the given container list
+// and, if they are a target, it will obfuscate their value
+func obfuscateContainerEnvVars(containers []corev1.Container) {
+	targets := []string{"HTTP_PROXY", "HTTPS_PROXY"}
+	search := regexp.MustCompile(strings.Join(targets, "|"))
```
tremes, line 213: Nit: I think regex pattern is not required or is it? On the other hand it's probably not a big deal....

ncaak:
/hold
as per [this comment](https://github.com/openshift/insights-operator/pull/723#discussion_r1088643742), I found this logic could be applied to several other files.

pkg/gatherers/clusterconfig/container_images.go
ncaak: I accept the nit.

Actually, this way is not very costly. Regex is not using wildcards. Only check for exact notation is usually very fast. I did the code this way just in case more ENV VARS are suitable to be obfuscated in the future: you only need to add a new item to the slice.

If you prefer not involve any regexp, what is fair, I can reuse the slice approach and make a loop comparing strings with Env var names. Or, if you know any other magic or vendor package already on the project that could do this easier, please recommend.

tremes: OK fair point. Let's leave it as is

pkg/gatherers/clusterconfig/container_images.go
```diff
@@ -202,3 +205,18 @@ func gatherImages(startMonth string, img2month2count img2Month2CountMap, contain
...
+
+	for i := range containers {
+		for j := range containers[i].Env {
+			if search.MatchString(containers[i].Env[j].Name) {
+				containers[i].Env[j].Value = "<obfuscated>"
```
tremes, line 218: Nit again: We usually repeat `x` (for the value) in this case of "default" obfuscation. See https://github.com/openshift/insights-operator/blob/master/pkg/utils/anonymize/string.go.

pkg/utils/anonymize/envvars_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// When
+	SensitiveEnvVars(mock)
+
+	// Assert
+	t.Run("Non target env vars keep their original value", func(t *testing.T) {
+		test := mock[0].Env[0]
+		if test.Value != envOriginalValue {
+			t.Logf("\nexpected the variable '%s' to have 'original_value'\ngot: %s", test.Name, test.Value)
+			t.Fail()
```
tremes, line 30: Definitely not a blocker, but we have https://github.com/openshift/insights-operator/blob/master/go.mod#L17 so you can use that library for testing (I think it's more convenient, when we alread have the dependency)

tremes:
Just one minor comment, but it works as expected. Thanks
/lgtm

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 	Please see OpenShift release notes for official changes\n<!--Latest hash: 76cc46b739c27281db0f7b2a23eb091b2a70f698-->
 ## 4.13
 
+### Data Enhancement
+- [#723](https://github.com/openshift/insights-operator/pull/723) Obfuscate HTTP_PROXY and HTTPS_PROXY env variables on containers
```
tremes, line 6: First time someone updated changelog with a new PR. :) This is not a new data enhancement, but rather bugfix or other (and yes it corresponds to the PR template selection)

tremes: New data enhancement is e.g new gatherer for XYZ.

ncaak: Ok, updating it right now to BugFix. Thanks.

tremes:
I've already reviewed and verified this one and it stll 
/lgtm

rluders:
/restest
Lucky one... maybe?

JoaoFula:
This one is currently failing because https://github.com/openshift/insights-operator/pull/726 hasn't been merged. I can revert the tests and remove test_gather_metrics that checks the existence of openshift_apps_deploymentconfigs_strategy_total since we know exists and readd it once the PR has been merged.

PR #724, opened 2023-01-27
PoC for moving the basic config options to a new config map watched by a new informer

tremes:
This first PoC introducing a new `insights-config` config map, which should ideally replace all the config options from the `support` secret one day.  Currently the values defined in the "support" secret (if any exists) take precedence over values provided in the new config map. This is to preserve the backward compatibility. 

Please note that we don't want to expose all the default values in this new config map so some of the default values remain defined in the `config/pod.yaml` (e.g the `interval` for now). I think it would make sense to expose all the endpoints as well as `storagePath` options by default and keep all the duration options in the default config (but of course it's possible to override them using the config map)

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

no new data

## Unit Tests

TBD
## Privacy

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browse/CCXDEV-9910


pkg/controller/periodic/periodic.go
```diff
@@ -185,17 +188,32 @@ func (c *Controller) Gather() {
...
 		select {
 		case <-stopCh:
 			return
+		case <-cmConfigCh:
+			newInterval := c.configMapConfigurator.Config().DataReporting.Interval
+			if newInterval == interval {
+				continue
+			}
+			interval = newInterval
+			klog.Infof("Gathering cluster info every %s - from config map", interval)
```
tremes, line 214: ".... from config map" needs to be removed.

tremes:
We will not do this.

PR #728, opened 2023-02-07
test(clusterconfig): add unit test for silenced_alerts.go and rename it to gather_silenced_alerts.go

rluders:
This PR adds a simple unit test for `silenced_alerts` gatherer, and renames its file from `silenced_alerts.go -> gather_silenced_alerts.go`

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

None

## Documentation

None

## Unit Tests

- `pkg/gatherers/clusterconfig/gather_silenced_alerts_test.go`

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-10128


pkg/gatherers/clusterconfig/gather_silenced_alerts_test.go
```diff
@@ -0,0 +1,73 @@
...
+			expectedRecords:    nil,
+			expectedErrorCount: 1,
+		},
+	}
+
+	ctx := context.Background()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			records, errs := gatherSilencedAlerts(ctx, tt.mockAlertsClient.RestClient(t))
+			if len(errs) != tt.expectedErrorCount {
```
rluders, line 65: I'll replace with assert

pkg/gatherers/clusterconfig/gather_silenced_alerts_test.go
```diff
@@ -0,0 +1,70 @@
...
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if c.data == nil {
+			w.WriteHeader(http.StatusNotFound)
+		} else {
+			w.WriteHeader(http.StatusOK)
+			// nolint: errcheck
+			w.Write(c.data)
+		}
+	}))
+	return testRESTClient(t, ts)
```
tremes, line 31: This is defined in `node_logs_test.go`. I think this is not very "friendly" approach. Imagine that you're working and changing the `node_logs_test.go` and suddenly you also have failures in other unrelated test.

pkg/gatherers/clusterconfig/gather_silenced_alerts_test.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (c *mockAlertsClient) RestClient(t *testing.T) *rest.RESTClient {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if c.data == nil {
+			w.WriteHeader(http.StatusNotFound)
+		} else {
+			w.WriteHeader(http.StatusOK)
+			// nolint: errcheck
+			w.Write(c.data)
```
tremes, line 28: What is the reason to not check the error?

rluders: Well, it should not return an error, 'cause the data will never fail to be written. But, I can add the error handling, useless, but not a big deal.

pkg/gatherers/clusterconfig/gather_silenced_alerts_test.go
rluders: So, I agree. I'm considering to create a `testing/` folder at the root project (not sure if this is the best approach, but seems to be very common) to add some test utilities there. But, I'm not sure how many test utilities we have lost/shared across the project. I want to handle it in a separate task.

rluders: Another approach is to have some repeated code, between the tests, leaving them as independent as possible. But, as I said. I'm not sure yet, I want to check how many dependencies, and testing utilities we have before taking any decision.

tremes: I think little copying in this case is OK, second option of moving the helper functions and reuse them is also OK.

pkg/gatherers/clusterconfig/gather_silenced_alerts_test.go
```diff
@@ -0,0 +1,80 @@
...
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
+			}
+		}
+	}))
+
+	baseURL, _ := url.Parse(ts.URL)
+	client, err := rest.NewRESTClient(baseURL, "", rest.ClientContentConfig{}, nil, nil)
+	if err != nil {
+		t.Fatalf("failed to create a client: %v", err)
```
tremes, line 38: assert? :)

pkg/gatherers/clusterconfig/gather_silenced_alerts_test.go
```diff
@@ -0,0 +1,80 @@
...
+		} else {
+			w.WriteHeader(http.StatusOK)
+			_, err := w.Write(c.data)
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
+			}
+		}
+	}))
+
+	baseURL, _ := url.Parse(ts.URL)
```
tremes, line 35: I think we should check the error here.

pkg/gatherers/clusterconfig/gather_silenced_alerts_test.go
```diff
@@ -0,0 +1,80 @@
...
+			mockAlertsClient: &mockAlertsClient{
+				data: []byte(`[{"status": {"state": "suppressed"}}]`),
+			},
+			wantRecords: []record.Record{
+				{
+					Name: "config/silenced_alerts.json",
+					Item: marshal.RawByte(`[{"status": {"state": "suppressed"}}]`),
+				},
+			},
+			wantErrsCount: 0,
```
tremes, line 62: Suggestion. It would be more robust to check the particular error messages than just number of errors.

rluders: I added the test, but, TBH, in this specific case, I don't see much value/difference between checking the error type or the number of errors. 🤔

tremes:
Simple unit test. Reviewed
/lgtm

rluders:
This is just a unit test + file renaming should not affect operator functionality. No need for e2e.

rluders:
/override ci/prow/e2e-agnostic-upgrade ci/prow/insights-operator-e2e-tests

rluders:
This is just unit test.

PR #732, opened 2023-02-13
test(clusterconfig): add unit test for version.go and rename it to gather_cluster_version.go

rluders:
This PR adds a simple unit test for the `cluster version` gatherer and renames its file from `version.go -> gather_cluster_version.go`

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

None

## Documentation

None

## Unit Tests

- `pkg/gatherers/clusterconfig/gather_cluster_version_test.go`

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-10113


pkg/gatherers/clusterconfig/gather_cluster_version_test.go
```diff
@@ -0,0 +1,134 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+
+	"github.com/openshift/insights-operator/pkg/record"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_getClusterVersion(t *testing.T) {
+	lastTimestampEvent := metav1.Time{Time: time.Now().Add(2)}
+	compactEvents := eventListToCompactedEventList(&v1.EventList{
```
rluders, line 24: Basically, it needs to be here, 'cause when it passes to the `JSONMarshaller` if I don't pass it as a pointer, it doesn't match with the way that the info is generated by the gatherer.

pkg/gatherers/clusterconfig/gather_cluster_version_test.go
```diff
@@ -0,0 +1,134 @@
...
+				Items: []v1.Event{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Namespace: "openshift-cluster-version",
+						},
+						Reason:        "Unit test",
+						Message:       "This is a unit test event",
+						LastTimestamp: lastTimestampEvent,
+						Count:         1,
+					},
```
tremes, line 81: Just detail. This event definition repeats here so you could extract it to some variable as well and save few lines of code.

rluders: Done. I do have a feeling that this isn't something that we should do... But, I have no logical argument to say "why"... and in this specific case, I don't see benefits to not doing it. In my mind, it feels like the data in table-driven tests should test specific and the dups are expected and acceptable. But, as I said... I have no logical argument, just this feeling.

tremes:
Reviewed. 
/lgtm

rluders:
/override ci/prow/lint ci/prow/e2e ci/prow/insights-operator-e2e-tests

rluders:
Unit test, doesn't need e2e

PR #744, opened 2023-02-28
service_accounts.go Marshall fix

rhrmo:
This PR fixes the bug in service_accounts.go where in Marshall function there wouldn't be all service accounts saved.

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/config/serviceaccounts.json`

## Documentation

no update

## Unit Tests

- `pkg/gatherers/clusterconfig/service_accounts_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browse/CCXDEV-10265

### DIFF
--- a/pkg/gatherers/clusterconfig/service_accounts.go
+++ b/pkg/gatherers/clusterconfig/service_accounts.go
@@ -103,25 +103,29 @@ type ServiceAccountsMarshaller struct {
 // Marshal implements serialization of ServiceAccount
 func (a ServiceAccountsMarshaller) Marshal() ([]byte, error) {
 	// Creates map for marshal
-	sr := map[string]interface{}{}
-	st := map[string]interface{}{}
-	st["TOTAL_COUNT"] = a.totalServiceAccounts
-	sr["serviceAccounts"] = st
-	nss := map[string]interface{}{}
-	st["namespaces"] = nss
-	for i := range a.sa {
-		var ns map[string]interface{}
-		var ok bool
-		if _, ok = nss[a.sa[i].Namespace]; !ok {
-			ns = map[string]interface{}{}
-			nss[a.sa[i].Namespace] = ns
-		} else {
-			ns = nss[a.sa[i].Namespace].(map[string]interface{})
-		}
-		ns["name"] = a.sa[i].Name
-		ns["secrets"] = len(a.sa[i].Secrets)
+	serviceAccounts := map[string]any{}
+	namespaces := map[string][]map[string]any{}
+	serviceAccounts["serviceAccounts"] = map[string]any{
+		"TOTAL_COUNT": a.totalServiceAccounts,
+		"namespaces":  namespaces,
 	}
-	return json.Marshal(sr)
+
+	for i := range a.sa {
+		if _, ok := namespaces[a.sa[i].Namespace]; !ok {
+			namespaces[a.sa[i].Namespace] = []map[string]any{
+				{
+					"name":    a.sa[i].Name,
+					"secrets": len(a.sa[i].Secrets),
+				},
+			}
+		} else {
+			namespaces[a.sa[i].Namespace] = append(namespaces[a.sa[i].Namespace], map[string]any{
+				"name":    a.sa[i].Name,
+				"secrets": len(a.sa[i].Secrets),
+			})
+		}
+	}
+	return json.Marshal(serviceAccounts)
 }
 
 // GetExtension returns extension for anonymized openshift objects
--- a/pkg/gatherers/clusterconfig/service_accounts_test.go
+++ b/pkg/gatherers/clusterconfig/service_accounts_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubefake "k8s.io/client-go/kubernetes/fake"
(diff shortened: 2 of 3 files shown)


rluders (changes requested):

pkg/gatherers/clusterconfig/service_accounts.go
```diff
@@ -103,25 +103,29 @@ type ServiceAccountsMarshaller struct {
...
 	for i := range a.sa {
-		var ns map[string]interface{}
-		var ok bool
-		if _, ok = nss[a.sa[i].Namespace]; !ok {
-			ns = map[string]interface{}{}
-			nss[a.sa[i].Namespace] = ns
+		if _, ok := namespaces[a.sa[i].Namespace]; !ok {
+			namespaces[a.sa[i].Namespace] = []map[string]any{
+				{
+					"name":    a.sa[i].Name,
```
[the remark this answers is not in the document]
rluders: Also, it seems that this `[]map[string]any{...` is repeating a little bit.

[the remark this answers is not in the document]
rhrmo: a.sa[i] is service account, maybe I could change it's name in the struct :thinking:

pkg/gatherers/clusterconfig/service_accounts.go
```diff
@@ -96,32 +96,35 @@ func gatherServiceAccounts(ctx context.Context, coreClient corev1client.CoreV1In
...
-	st["namespaces"] = nss
-	for i := range a.sa {
-		var ns map[string]interface{}
-		var ok bool
-		if _, ok = nss[a.sa[i].Namespace]; !ok {
-			ns = map[string]interface{}{}
-			nss[a.sa[i].Namespace] = ns
-		} else {
-			ns = nss[a.sa[i].Namespace].(map[string]interface{})
+	serviceAccounts := map[string]any{}
```
rluders, line 106: This line seems useless. Can't you just do 106 and 108 in the same line?

rhrmo: I have tried that before and it doesn't work unfortunately :/

rhrmo: something like:
`serviceAccounts := map[string]any{
		"serviceAccounts" : map[string]any{
			"TOTAL_COUNT": a.totalServiceAccounts,
			"namespaces":  namespaces,
		},
	}`
would work but I think it's better the way it is now

tremes: You can't. Map has to be initialized in Go.

pkg/gatherers/clusterconfig/service_accounts.go
```diff
@@ -96,32 +96,35 @@ func gatherServiceAccounts(ctx context.Context, coreClient corev1client.CoreV1In
...
+		serviceAccount := []map[string]any{
+			{
+				"name":    a.serviceAccounts[i].Name,
+				"secrets": len(a.serviceAccounts[i].Secrets),
+			},
+		}
+		_, ok := namespaces[a.serviceAccounts[i].Namespace]
+		if !ok {
+			namespaces[a.serviceAccounts[i].Namespace] = serviceAccount
+			continue
```
tremes, line 123: I think we don't need this. Do we?

tremes, line 123: I think we don't need this. Do we?

rluders: I think that it does make sense. At line 125 it will append a new SA to the namespace map with the SA index. So, I'm assuming that we have > 1 SA per namespace, which means that you need first to create the index in the namespace map before appending it.

But probably what we don't need is line 114 to be a map as well...

rluders:
```go
// here comes the custom type that you suggested
serviceAccount := map[string]any{
    "name":    a.serviceAccounts[i].Name,
    "secrets": len(a.serviceAccounts[i].Secrets),
}
_, ok := namespaces[a.serviceAccounts[i].Namespace]
if !ok {
    namespaces[a.serviceAccounts[i].Namespace] = serviceAccount
    continue
}
namespaces[a.serviceAccounts[i].Namespace] = append(namespaces[a.serviceAccounts[i].Namespace], serviceAccount)
```

rluders: @rhrmo my fault here, that one should not be a slice. It is hard to write code directly to GitHub. lol

rhrmo: We need it to be slice of maps, because one namespace can have multiple service accounts, but I like the idea of introducing the type for the service account so it will get more clear after I update it.

rhrmo: nvm I was wrong, it does actually work without it being a slice and without ok

pkg/gatherers/clusterconfig/service_accounts.go
```diff
@@ -96,32 +96,35 @@ func gatherServiceAccounts(ctx context.Context, coreClient corev1client.CoreV1In
...
+		"TOTAL_COUNT": a.totalServiceAccounts,
+		"namespaces":  namespaces,
+	}
+
+	for i := range a.serviceAccounts {
+		serviceAccount := []map[string]any{
+			{
+				"name":    a.serviceAccounts[i].Name,
+				"secrets": len(a.serviceAccounts[i].Secrets),
+			},
```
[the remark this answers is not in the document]
rhrmo: @tremes after introducing type the json.Marshal function doesn't behave the same as with only maps, and it does not save the service accounts, it just leaves empty {} brackets. So I think we would need to change all of the maps to types and I'm not sure if that is useful :thinking:

pkg/gatherers/clusterconfig/service_accounts_test.go
```diff
@@ -52,32 +82,23 @@ func Test_ServiceAccounts_Gather(t *testing.T) {
 			t.Parallel()
 			coreClient := kubefake.NewSimpleClientset()
 			for _, d := range test.data {
-				_, err := coreClient.CoreV1().Namespaces().Create(
-					context.Background(),
-					&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: d.Namespace}}, metav1.CreateOptions{},
-				)
-				if err != nil {
-					t.Fatalf("unable to create fake ns %s", err)
+				namespace, _ := coreClient.CoreV1().Namespaces().Get(context.Background(), d.Namespace, metav1.GetOptions{})
```
tremes, line 85: Do not ignore the error please

tremes:
I reviewed and tested this one. Works as expected.
/lgtm

PR #745, opened 2023-03-06
update periodic_test.go & run unit tests with -race param

tremes:
Updates `periodic_test.go` to be more robust and readable and also adds `-race` param to unit tests make target.

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- no new data

## Unit Tests

- `pkg/controller/periodic/periodic_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ test: unit ## Run all the tests
 # Run the unit tests
 .PHONY: unit
 unit: ## Run the unit tests
-	go test $(GO_TEST_FLAGS) -coverprofile $(COVER_PROFILE) ./...
+	go test -race $(GO_TEST_FLAGS) -coverprofile $(COVER_PROFILE) ./...
 
 .PHONY: coverage
 coverage:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -58,7 +58,7 @@ test: unit ## Run all the tests
 # Run the unit tests
 .PHONY: unit
 unit: ## Run the unit tests
-	go test $(GO_TEST_FLAGS) -coverprofile $(COVER_PROFILE) ./...
+	go test -race $(GO_TEST_FLAGS) -coverprofile $(COVER_PROFILE) ./...
```
tremes, line 61: I am still wondering if we should also add e.g `-count 20`, which should help detect possible concurrency problems faster (if I am not mistaken), but the tests will also take longer to run.

tremes, line 61: I am still wondering if we should also add e.g `-count 20`, which should help detect possible concurrency problems faster (if I am not mistaken), but the tests will also take longer to run.

ncaak: Did you do a timed test on that to see how long could it take?

tremes: I guess 20x longer.....

PR #747, opened 2023-03-07
test(clusterconfig): Unit tests for kube controller manager logs

ncaak:
This PR implements a new unit tests for kube controller manager logs gatherer

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- No

## Documentation

- No need

## Unit Tests

- `pkg/gatherers/clusterconfig/gather_kube_controller_manager_logs_test.go`

## Privacy

N/A

## Changelog

Yes

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-10114


tremes:
It seems to me that this is in fact testing the `common.FilterLogFromScanner`, which we already have tests in https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/common/gather_logs_test.go or am I missing something?

rluders:
> It seems to me that this is in fact testing the `common.FilterLogFromScanner`, which we already have tests in https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/common/gather_logs_test.go or am I missing something?

Yes, but no. The idea here is not to test the `common.FilterLogFromScanner` but to test if the filtering parameters are working as expected. So, it could look like a dup, but it isn't. We are ensuring that the filters we use in this gatherer are working an returning the expected value.

rluders (approved):
/lgtm

ncaak:
Cancelling this PR to summarize Log Gatherers changes in the same PR.
Merged changes into [PR #751](https://github.com/openshift/insights-operator/pull/751/)

PR #778, opened 2023-05-16
chore: adds codecov

rluders:
This PR adds the codecov to the repository.

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

N/A

## Documentation

N/A

## Unit Tests

N/A

## Privacy

N/A

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-10925


ncaak (approved):

.codecov.yml
```diff
@@ -0,0 +1,25 @@
...
+    require_ci_to_pass: no
+
+coverage:
+  precision: 2
+  round: down
+  range: "20...100"
+  status:
+    project: no
+    patch: false
+    changes: false
```
tremes, line 12: Just out of curiosity. Is there some docs for these attributes?

rluders: Well, the Codecov documentation should have it.

rluders: https://docs.codecov.com/docs/codecov-yaml

.codecov.yml
```diff
@@ -0,0 +1,25 @@
+codecov:
+  notify:
+    require_ci_to_pass: no
+
+coverage:
+  precision: 2
+  round: down
+  range: "20...100"
```
tremes, line 8: I think the second number set to 100 doesn't make much sense when reading https://docs.codecov.com/docs/coverage-configuration#range. I would go with 70 or 75

.codecov.yml
tremes: Are `project` and `patch` attributes set correctly? I was looking at https://docs.codecov.com/docs/commit-status#disabling-a-status

.codecov.yml
rluders: @tremes check if DVO doesn’t have the same config. If I’m not wrong I copied the same config, maybe we want to change it there as well.

.codecov.yml
rluders: AFAIK, both should works but no problem to change it to `off`.

rluders:
```sh
➜ curl --data-binary @.codecov.yml https://codecov.io/validate

Valid!

{
  "codecov": {
    "notify": {},
    "require_ci_to_pass": false
  },
  "coverage": {
    "precision": 2,
    "round": "down",
    "range": [
      20.0,
      100.0
    ],
    "status": {
      "project": false,
      "patch": false,
      "changes": false
    }
  },
  "parsers": {
    "gcov": {
      "branch_detection": {
        "conditional": true,
        "loop": true,
        "method": false,
        "macro": false
      }
    }
  },
  "comment": {
    "layout": "reach,diff,flags,tree",
    "behavior": "default",
    "require_changes": false
  }
}
```

Never mind I'll not change it... I checked and the validator seems to be fine with it.

tremes: OK thanks and no problem, but now the range `range: "70...75"` seems to strict to me. :) I would go something like `40..75` or `30..75`

rluders: OK. Let me change it to 30...75. :)

ncaak (approved):
/retest

rluders:
/override lint

Lint is failing for other reasons, not 'cause of the changes on this PR.

rluders:
/override ci/prow/lint

Lint is failing for other reasons, not 'cause of the changes on this PR.

PR #779, opened 2023-05-19
OCPBUGS-13547: read featuregates from the shared API value

deads2k:
This brings featuregate access up to the 4.14 standard.

/hold

I need to locate the version


pkg/cmd/start/start.go
```diff
@@ -113,6 +119,41 @@ func runGather(operator *controller.GatherJob, cfg *controllercmd.ControllerComm
...
 		ctx, cancel := context.WithTimeout(context.Background(), operator.Interval)
+
+		// before calling the Gather logic, this command must check the featuregate status.
+		configClient, err := configv1client.NewForConfig(clientConfig)
+		if err != nil {
+			klog.Exit("error: building client: %v", err)
+		}
+		configInformers := configv1informers.NewSharedInformerFactory(configClient, 10*time.Minute)
+
+		desiredVersion := config.OperatorReleaseVersion
```
deads2k, line 130: this needs to be teh version that we're going to write into the clusteroperator.

The gather command needs to know the version too.

pkg/cmd/start/start.go
```diff
@@ -113,6 +119,41 @@ func runGather(operator *controller.GatherJob, cfg *controllercmd.ControllerComm
...
 		ctx, cancel := context.WithTimeout(context.Background(), operator.Interval)
+
+		// before calling the Gather logic, this command must check the featuregate status.
+		configClient, err := configv1client.NewForConfig(clientConfig)
+		if err != nil {
+			klog.Exit("error: building client: %v", err)
+		}
+		configInformers := configv1informers.NewSharedInformerFactory(configClient, 10*time.Minute)
+
+		desiredVersion := os.Getenv("RELEASE_VERSION")
```
deads2k, line 130: What launches the gather job?  Having this match is important to get consistent feature gate behavior.

deads2k:
I think this is updated enough to be functional

/hold cancel

pkg/cmd/start/start.go
tremes: This can be launched "manually" by the user (kubeadmin). Basically as an one-off operation - see https://docs.openshift.com/container-platform/4.12/support/remote_health_monitoring/remote-health-reporting-from-restricted-network.html. I think this is very edge case and I plan to remove it once the on-demand gathering is GA.

pkg/controller/operator.go
```diff
@@ -82,24 +84,45 @@ func (s *Operator) Run(ctx context.Context, controller *controllercmd.Controller
 		return err
 	}
 
+	desiredVersion := os.Getenv("RELEASE_VERSION")
+	missingVersion := "0.0.1-snapshot"
```
tremes, line 88: This corresponds to https://github.com/openshift/insights-operator/blob/master/manifests/07-cluster-operator.yaml#L15, right? And it's updated by the CVO, right?

deads2k:
finishing off the external cloud provider featuregate shift.

pkg/controller/operator.go
```diff
@@ -82,24 +84,45 @@ func (s *Operator) Run(ctx context.Context, controller *controllercmd.Controller
...
+	}
+	insightsConfigAPIEnabled := featureGates.Enabled(v1.FeatureGateInsightsConfigAPI)
+
 	// ensure the insight snapshot directory exists
 	if _, err = os.Stat(s.StoragePath); err != nil && os.IsNotExist(err) {
 		if err = os.MkdirAll(s.StoragePath, 0777); err != nil {
 			return fmt.Errorf("can't create --path: %v", err)
 		}
 	}
-	tpEnabled, err := isTechPreviewEnabled(ctx, configClient)
```
tremes, line 91: Can you please also remove the function `isTechPreviewEnabled` definition?

tremes:
@deads2k I think we need to update the permissions for the operator role, because I wasn't able to get a cluster (from your PR):
```
E0601 07:52:41.026679       1 reflector.go:148] github.com/openshift/client-go/config/informers/externalversions/factory.go:101: Failed to watch *v1.ClusterVersion: failed to list *v1.ClusterVersion: clusterversions.config.openshift.io is forbidden: User "system:serviceaccount:openshift-insights:operator" cannot list resource "clusterversions" in API group "config.openshift.io" at the cluster scope
```
you can see https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1664160810571665408

PR #780, opened 2023-05-21
gather PDBs only from openshift namespaces

rhrmo:

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

### DIFF
--- a/pkg/gatherers/clusterconfig/pod_disruption_budgets.go
+++ b/pkg/gatherers/clusterconfig/pod_disruption_budgets.go
@@ -3,6 +3,7 @@ package clusterconfig
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	policyclient "k8s.io/client-go/kubernetes/typed/policy/v1"
@@ -11,7 +12,8 @@ import (
 )
 
 const (
-	gatherPodDisruptionBudgetLimit = 5000
+	gatherPodDisruptionBudgetLimit = 100
+	serverRequestLimit             = 5000
 )
 
 // GatherPodDisruptionBudgets Collects the cluster's `PodDisruptionBudgets`.
@@ -37,7 +39,7 @@ const (
 // - 4.5.15+
 //
 // ### Changes
-// None
+// - In 4.14 we changed the gatherer to gather pdbs only from namespaces with "openshift-" prefix and the limit to 100.
 func (g *Gatherer) GatherPodDisruptionBudgets(ctx context.Context) ([]record.Record, []error) {
 	gatherPolicyClient, err := policyclient.NewForConfig(g.gatherKubeConfig)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/gatherers/clusterconfig/pod_disruption_budgets.go
```diff
@@ -48,20 +51,24 @@ func (g *Gatherer) GatherPodDisruptionBudgets(ctx context.Context) ([]record.Rec
...
 	if err != nil {
 		return nil, []error{err}
 	}
 	var records []record.Record
+	limit := 0
 	for i := range pdbs.Items {
-		recordName := fmt.Sprintf("config/pdbs/%s", pdbs.Items[i].GetName())
-		if pdbs.Items[i].GetNamespace() != "" {
-			recordName = fmt.Sprintf("config/pdbs/%s/%s", pdbs.Items[i].GetNamespace(), pdbs.Items[i].GetName())
+		if strings.HasPrefix(pdbs.Items[i].GetNamespace(), "openshift") {
```
tremes, line 61: Better to extract `pdbs.Items[i]` to a new variable and check `openshift-`

pkg/gatherers/clusterconfig/pod_disruption_budgets_test.go
```diff
@@ -5,55 +5,114 @@ import (
...
+			for namespace, name := range test.pdbsToNamespace {
+				_, err := coreClient.PolicyV1().
+					PodDisruptionBudgets(namespace).
+					Create(context.Background(), &policyv1.PodDisruptionBudget{
+						ObjectMeta: metav1.ObjectMeta{
+							Namespace: namespace,
+							Name:      name,
+						},
+						Spec: policyv1.PodDisruptionBudgetSpec{
+							MinAvailable: &intstr.IntOrString{IntVal: int32(test.minAvailableToPdb[name])},
```
tremes, line 91: Is this required?

pkg/gatherers/clusterconfig/pod_disruption_budgets_test.go
```diff
@@ -5,55 +5,114 @@ import (
...
+						},
+					}, metav1.CreateOptions{})
+				if err != nil {
+					t.Fatalf("unable to create fake pdbs: %v", err)
```
tremes, line 95: Let's use `assert` please

pkg/gatherers/clusterconfig/pod_disruption_budgets.go
```diff
@@ -11,7 +12,8 @@ import (
 )
 
 const (
-	gatherPodDisruptionBudgetLimit = 5000
+	gatherPodDisruptionBudgetLimit = 100
+	serverRequestLimit             = 5000
```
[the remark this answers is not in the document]
rhrmo: Only if the server would return too many PDBs, but after reconsidering it I don't think that such case would happen.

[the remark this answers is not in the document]
tremes: I think we should either use the request limit and then use `continue` or don't use the limit at all.

pkg/gatherers/clusterconfig/pod_disruption_budgets_test.go
rhrmo: The old unit test tested if the min. available was asserted to the correct PDB. I left it in there just in case, but I don't think gatherer can mismatch the pdb and its spec (MinAvailable).

tremes: Ahh OK.

pkg/gatherers/clusterconfig/pod_disruption_budgets.go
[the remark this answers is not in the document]
rhrmo: I think it will be simpler to just not use the limit at all in that case, since it is set as an attribute in metav1.ListOptions{} struct.

tremes:
I reviewed and briefly tested. Thanks.
/lgtm

PR #792, opened 2023-06-09
OCPBUGS-14771: run an extra config informer in the tech preview

tremes:
https://github.com/openshift/insights-operator/pull/779 changed the logic in `operator.go` so that the `configinformer` runs every time (and not only in the techpreview). This caused the problem (unable to sync cache -> restart every 10 min) in the techpreview, because it doesn't seem to be possible to add an additional informer, when it's already started. 

This creates (and starts) a new extra config informer only in the techpreview when the feature is enabled. 

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

-no new data
## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/OCPBUGS-14771

### DIFF
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -88,7 +88,7 @@ type Anonymizer struct {
 	ipNetworkRegex     *regexp.Regexp
 	secretsClient      corev1client.SecretInterface
 	secretConfigurator configobserver.Configurator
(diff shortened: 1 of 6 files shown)


tremes:
We need to exclude the `depguard` lint.
/override ci/prow/lint

ncaak (approved):
/lgtm

PR #805, opened 2023-06-28
Implement DataGather status conditions and status propagation

tremes:

## Categories

- [ ] Bugfix
- [X] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

### DIFF
--- a/pkg/controller/periodic/periodic.go
+++ b/pkg/controller/periodic/periodic.go
@@ -52,6 +52,7 @@ type Controller struct {
 	image               string
 	jobController       *JobController
 	pruneInterval       time.Duration
+	techPreview         bool
 }
 
 func NewWithTechPreview(
@@ -64,6 +65,8 @@ func NewWithTechPreview(
 	insightsOperatorCLI operatorv1client.InsightsOperatorInterface,
 ) *Controller {
 	statuses := make(map[string]controllerstatus.StatusController)
+
+	statuses["tech-preview-test"] = controllerstatus.New("tech-preview-test")
 	jobController := NewJobController(kubeClient)
 	return &Controller{
 		reportRetriever:     reportRetriever,
@@ -76,6 +79,7 @@ func NewWithTechPreview(
 		jobController:       jobController,
 		insightsOperatorCLI: insightsOperatorCLI,
 		pruneInterval:       1 * time.Hour,
+		techPreview:         true,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/controller/periodic/periodic.go
```diff
@@ -493,6 +504,41 @@ func (c *Controller) createDataGatherAttributeValues() ([]string, insightsv1alph
...
+	if dataUploadedCon.Status == metav1.ConditionFalse {
+		statusSummary.Healthy = false
+		statusSummary.Count = 5
+		statusSummary.Reason = dataUploadedCon.Reason
+		statusSummary.Message = dataUploadedCon.Message
+		c.statuses["insightsuploader"].UpdateStatus(statusSummary)
+		return false
+	}
+	c.statuses["insightsuploader"].UpdateStatus(statusSummary)
+	return true
```
ncaak, line 539: nit:
I think this could be refactored a bit to be a little clear. Not bad code, only a recommendation.
* You could modify the conditional part from two `if`, to a `if - else if`
* Then not using several returns, but to set `return statusSummary.Healthy` (as it's already set false or true depending on the branch)
* only one `c.statuses["insightsuploader"].UpdateStatus(statusSummary)` before the return

In any case, the code is not that complex as to **need** this refactor.

ncaak: e.g.
```
if dataUploadedCon == nil {
		statusSummary.Healthy = false
		statusSummary.Count = 5
		statusSummary.Reason = "DataUploadedConditionNotAvailable"
		statusSummary.Message = fmt.Sprintf("did not find any %q condition in the %s dataGather resource",
			status.DataUploaded, dataGather.Name)
	} else if dataUploadedCon.Status == metav1.ConditionFalse {
		statusSummary.Healthy = false
		statusSummary.Count = 5
		statusSummary.Reason = dataUploadedCon.Reason
		statusSummary.Message = dataUploadedCon.Message
	}

	c.statuses["insightsuploader"].UpdateStatus(statusSummary)
	return statusSummary.Healthy
```

tremes: Thanks a lot for the suggestion. I'll take a look.

tremes: Updated. Please check

ncaak: Thanks, it looks great

ncaak (approved):
/lgtm

PR #807, opened 2023-06-30
OCPBUGS-13915: create Prometheus rules programmatically according the config option

tremes:
Prior to this PR the Prometheus rules (i.e our Insights Alert definitions) were created by the ClusterVersion Operator based on the `manifests/08-prometheus_rule.yaml` file. This means that we didn't have much control over them. 

This PR removes the Prometheus rule manifest and creates them programmatically according the corresponding `disableInsightsAlert` config option. In other words, previously when `disableInsightsAlerts=true` then only `InsightsRecommendationActive` alert was not available, but now no Insights alert (`InsightsDisabled`, `SimpleContentNotAvailable`) is available in this case.

This adds a new dependency to the prometheus-operator APIs & client...thus the vendor files.

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

no new data

## Unit Tests

- `pkg/insights/prometheus_rules_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browse/OCPBUGS-13915


pkg/insights/prometheus_rules.go
```diff
@@ -0,0 +1,153 @@
...
+
+// checkAlertsDisabled reads the actual config and either creates (if they don't exist) or removes (if they do exist)
+// the "insights-prometheus-rules" definition
+func (p *PrometheusRulesController) checkAlertsDisabled(ctx context.Context) {
+	if p.configurator.Config().DisableInsightsAlerts {
+		if p.promRulesExist {
+			err := p.removeInsightsAlerts(ctx)
+			if err != nil {
+				klog.Errorf("Failed to remove Insights Prometheus rules definition: %v ", err)
+			} else {
```
ncaak, line 70: nit: I think we can get rid of this else branch, just returning after the klog.Errorf. Then leaving the last statement with no extra tabulation.
e.g.
```
    if err != nil {
        klog.Errorf("Failed to remove Insights Prometheus rules definition: %v ", err)
        return
    }

    klog.Info("Prometheus rules successfully removed")
    p.promRulesExist = false

}
```

rluders: TBH, I think that this whole `checkAlertsDisabled` has a lot of space to improve. I'm not a big fan of if-else-if-else-if-else and a big fan of early return...

```go
func (p *PrometheusRulesController) checkAlertsDisabled(ctx context.Context) {
	disableInsightsAlerts := p.configurator.Config().DisableInsightsAlerts

	if disableInsightsAlerts && p.promRulesExist {
		if err := p.removeInsightsAlerts(ctx); err != nil {
			klog.Errorf("Failed to remove Insights Prometheus rules definition: %v", err)
			return
		}
		klog.Info("Prometheus rules successfully removed")
		p.promRulesExist = false
		return
	}

	if !disableInsightsAlerts && !p.promRulesExist {
		if err := p.createInsightsAlerts(ctx); err != nil {
			klog.Errorf("Failed to create Insights Prometheus rules definition: %v", err)
			return
		}
		klog.Info("Prometheus rules successfully created")
		p.promRulesExist = true
	}
}
```

☝️ I didn't tested it.

rluders: Or if you want, you could use `else if` for the second if block and then remove the early returns.

tremes: Thanks. Updated. Please check

tremes:
OCPBUGS should not require `qe-approved` label since it is required to be verified by QE anyway.
/label qe-approved

PR #808, opened 2023-07-13
update registration of the metrics to work in techpreview too

tremes:
This PR updates the logic of Prometheus metrics registration to work in techpreview (when data gathering runs in a job) as well. The registration and incrementation of the `insightsclient_request_send_total` metric is moved from the uploader and the metric in the techpreview is updated based on the conditions of the latest `datagather` CR (which provides information including HTTP status code in its conditions).  

This also removes the `insightsclient_last_gather_time` metric. This data is available in the Insights CRs. 

This also removes  

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive

no new data

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/insights/insightsclient/requests.go
+++ b/pkg/insights/insightsclient/requests.go
@@ -12,21 +12,22 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/insights-operator/pkg/authorizer"
+	"github.com/openshift/insights-operator/pkg/insights"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-func (c *Client) SendAndGetID(ctx context.Context, endpoint string, source Source) (string, error) {
+func (c *Client) SendAndGetID(ctx context.Context, endpoint string, source Source) (string, int, error) {
 	cv, err := c.GetClusterVersion()
 	if apierrors.IsNotFound(err) {
-		return "", ErrWaitingForVersion
+		return "", 0, ErrWaitingForVersion
 	}
 	if err != nil {
-		return "", err
+		return "", 0, err
 	}
 
 	req, err := c.prepareRequest(ctx, http.MethodPost, endpoint, cv)
 	if err != nil {
-		return "", err
+		return "", 0, err
 	}
 
 	bytesRead := make(chan int64, 1)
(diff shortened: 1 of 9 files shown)


pkg/insights/insightsclient/requests.go
```diff
@@ -12,21 +12,22 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/insights-operator/pkg/authorizer"
+	"github.com/openshift/insights-operator/pkg/insights"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-func (c *Client) SendAndGetID(ctx context.Context, endpoint string, source Source) (string, error) {
+func (c *Client) SendAndGetID(ctx context.Context, endpoint string, source Source) (string, int, error) {
```
ncaak, line 19: I'm having some concerns about confusing use of second return parameter. This is part of the same nit with upcoming comments. I prefer to leave the comments inline to avoid more confusion :D

As a summary, I'd assign a variable to default returning code, to improve readability. Then later use the same codes as parameters for a control function, moving the transformation of the data to that function itself. That way code is more straightforward IMHO, and avoids confusion on why transforming the data, why sending a 0 as a string, 0 int returnings with no context, etc...

I think it's a simple change but that improves readability and avoids any confusion on why or what the data is used for.

Please, check upcoming messages :point_down:

pkg/insights/insightsclient/requests.go
```diff
@@ -42,9 +43,9 @@ func (c *Client) SendAndGetID(ctx context.Context, endpoint string, source Sourc
 	resp, err := c.client.Do(req)
 	if err != nil {
 		klog.V(4).Infof("Unable to build a request, possible invalid token: %v", err)
-		// if the request is not build, for example because of invalid endpoint,(maybe some problem with DNS), we want to have recorded about it in metrics as well.
-		counterRequestSend.WithLabelValues(c.metricsName, "0").Inc()
-		return "", fmt.Errorf("unable to build request to connect to Insights server: %v", err)
+		// if the request is not build, for example because of invalid endpoint,(maybe some problem with DNS), we want to have record about it in metrics as well.
+		insights.IncrementCounterRequestSend("0")
```
ncaak, line 47: Send `defaultCode` (variable or const for default 0 value) as it is, the data transformation should take place inside `IncrementCounterRequestSend` function

pkg/insights/metrics.go
```diff
@@ -32,7 +36,11 @@ func MustRegisterMetrics(registrables ...metrics.Registerable) {
 }
 
 func init() {
-	MustRegisterMetrics(RecommendationCollector)
+	MustRegisterMetrics(RecommendationCollector, counterRequestSend)
+}
+
+func IncrementCounterRequestSend(status string) {
```
ncaak, line 42: change signature to accept `int` instead. Http codes type is `int` and we can play with a default value of the same type. Then before sending the data, transform it with `strconv` or `fmt` as string.

pkg/insights/insightsclient/requests.go
ncaak: e.g. insights.IncrementCounterRequestSend(defaultCode)

I think it's more clear what the 0 means, without even reading the comment above.

ncaak (approved):
/lgtm

tremes:
techpreview testsuite passed including the basic metrics test. I am adding the qe-approved label myself.
/label qe-approved

PR #810, opened 2023-07-26
check new processing status endpoint after uploading data in the job

tremes:
This PR implements a new check (HTTP GET) for the `processingStatusEndpoint` (better name suggestions are welcome), which requires cluster ID as well as the Insights request ID (which is part of the data upload HTTP response). 

The endpoint is polled three times (in case of error or response other than HTTP 200) with initial delay of 1 min (same as previously). If the check is successful then the HTTP response body is read and it's used to evaluate whether the data/archive processing was finished or not. The result is provided as the `DataProcessed` condition in the respective `DataGather` CR. 

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive

no new data

## Unit Tests

- `pkg/controller/gather_commands_test.go` - basic test for the new function

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/controller/gather_commands.go
+++ b/pkg/controller/gather_commands.go
@@ -2,10 +2,14 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/version"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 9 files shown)


pkg/controller/gather_commands.go
```diff
@@ -271,3 +289,56 @@ func record(functionReports []gather.GathererFunctionReport,
...
+
+	var resp *http.Response
+	err := wait.PollUntilContextCancel(ctx, delay, false, func(ctx context.Context) (done bool, err error) {
+		resp, err = insightsCli.GetDataProcessingStatus(ctx, // nolint: bodyclose
+			controllerConf.ProcessingStatusEndpoint, insightsRequestID) // response body is closed later
+		if err != nil {
+			return false, err
+		}
+		if resp.StatusCode == http.StatusOK || retryCounter == 2 {
+			return true, nil
```
ncaak, line 317: I'm a bit confused.
Why if we reach 2 retries (maximum allowed retries?) but no statusOK from the request, we return "true"?

tremes: Yes. Note that this return value is for the `wait.ConditionWithContextFunc` and not for the `wasDataProcessed` function and it means that we don't want to poll the endpoint again - i.e the condition was satisfied.

ncaak (approved):
/lgtm

PR #827, opened 2023-09-27
First minimal PoC version for moving the configuration to configmap

tremes:
This is minimal PoC for introducing the Insights Operator's configuration in the new `insights-config` configmap instead of the `support` secret configuration. The `support` secret configuration continues to work (to keep backward compatibility), but the configmap takes precedence if exists.

This PR introduces following new files:
- `pkg/config/configobserver/configmapobserver.go` - basic configmap observer watching the configmap's changes using Kubernetes informer
- `pkg/config/configobserver/config_aggregator.go` - this is a listener for both - the original "legacy" secret configobserver as well as the new configmap observer. The idea here is to aggregate or merge the config from both sources at one place. 

The `insights-config` configmap is supposed to live in the `openshift-insights` namespace and right now it can look like:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: 'insights-config'
  namespace: openshift-insights
data:
 config.yaml: |
   dataReporting:
     interval: 1h
     uploadEndpoint: https://console.redhat.com/api/ingress/v1/upload
binaryData: {}
immutable: false

```
Right now the config map allows you to override the following config options:
-  interval 
- upload endpoint
- conditional gatherer endpoint
- storage path
- download endpoint for tech preview
- processing status endpoint (tech preview only)

The configmap (as well as the config) structure is defined in the **`pkg/config/types.go`** so please review this one very carefully and feel free to suggest changes (for example naming, structure, etc.)

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive

- no new data

## Privacy

## Breaking Changes

Yes/No

## References

### DIFF
--- a/pkg/config/configobserver/config_aggregator.go
+++ b/pkg/config/configobserver/config_aggregator.go
@@ -3,7 +3,6 @@ package configobserver
 import (
 	"context"
 	"sync"
-	"time"
 
 	"github.com/openshift/insights-operator/pkg/config"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -104,7 +103,7 @@ func (c *ConfigAggregator) mergeStatically() {
 
 func (c *ConfigAggregator) merge(defaultCfg, newCfg *config.InsightsConfiguration) {
 	// read config map values and merge
-	if newCfg.DataReporting.Interval != 0*time.Minute {
+	if newCfg.DataReporting.Interval != 0 {
 		defaultCfg.DataReporting.Interval = newCfg.DataReporting.Interval
 	}
 
--- a/pkg/config/configobserver/config_aggregator.go
+++ b/pkg/config/configobserver/config_aggregator.go
@@ -6,6 +6,9 @@ import (
 	"time"
 
 	"github.com/openshift/insights-operator/pkg/config"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 )
 
@@ -21,8 +24,10 @@ type ConfigAggregator struct {
 	lock               sync.Mutex
 	legacyConfigurator Configurator
 	configMapInformer  ConfigMapInformer
-	configAggregated   *config.InsightsConfiguration
+	config             *config.InsightsConfiguration
 	listeners          map[chan struct{}]struct{}
+	usingInformer      bool
+	kubeClient         kubernetes.Interface
 }
 
 func NewConfigAggregator(ctrl Configurator, configMapInf ConfigMapInformer) Interface {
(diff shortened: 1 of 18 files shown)


ncaak (approved):

pkg/config/configobserver/config_aggregator.go
```diff
@@ -0,0 +1,194 @@
...
+		klog.Error("Failed to read configmap configuration: %v", err)
+		return
+	}
+
+	c.merge(conf, cmConf)
+}
+
+func (c *ConfigAggregator) merge(defaultCfg, newCfg *config.InsightsConfiguration) {
+	// read config map values and merge
+	if newCfg.DataReporting.Interval != 0*time.Minute {
```
rluders, line 95: You can create a helper function here.

```go
func updateFieldIfNotEmpty(field interface{}, newValue interface{}) {
	fieldValue := reflect.ValueOf(field)
	newValueIsValid := !reflect.DeepEqual(newValue, reflect.Zero(reflect.TypeOf(newValue)).Interface())

	if newValueIsValid {
		fieldValue.Elem().Set(reflect.ValueOf(newValue))
	}
}
```
And replace all the Ifs with:

```go
updateFieldIfNotEmpty(&defaultCfg.DataReporting.Interval, newCfg.DataReporting.Interval)
updateFieldIfNotEmpty(&defaultCfg.DataReporting.UploadEndpoint, newCfg.DataReporting.UploadEndpoint)
// more...
```

pkg/config/configobserver/configmapobserver.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	ctrl := &ConfigMapObserver{
+		kubeCli:        kubeClient,
+		insightsConfig: nil,
+		listeners:      make(map[chan struct{}]struct{}),
+	}
+	factoryCtrl := factory.New().WithInformers(cmInformer).
+		WithSync(ctrl.sync).
+		ResyncEvery(10*time.Minute).
```
rluders, line 59: Magic value. Convert to a constant so it is easy to find/change. Best if we could configure it, but not necessary, constant is good enough.

pkg/config/configobserver/configmapobserver.go
```diff
@@ -0,0 +1,133 @@
...
+
+// sync is called by the informer with every config map update
+func (c *ConfigMapObserver) sync(ctx context.Context, _ factory.SyncContext) error {
+	cm, err := getConfigMap(ctx, c.kubeCli)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+		// config map doesn't exist so clear the config, notify and return
+		klog.Info(err)
```
rluders, line 74: Maybe improve the logging message here?

pkg/config/configobserver/configmapobserver.go
```diff
@@ -0,0 +1,133 @@
...
+	ctrl.Controller = factoryCtrl
+	return ctrl, nil
+}
+
+// sync is called by the informer with every config map update
+func (c *ConfigMapObserver) sync(ctx context.Context, _ factory.SyncContext) error {
+	cm, err := getConfigMap(ctx, c.kubeCli)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return err
```
rluders, line 71: Better log message?

pkg/config/configobserver/configmapobserver.go
```diff
@@ -0,0 +1,133 @@
...
+	for ch := range c.listeners {
+		if ch == nil {
+			continue
+		}
+		ch <- struct{}{}
+	}
+}
+
+func (c *ConfigMapObserver) Config() *config.InsightsConfiguration {
+	c.lock.Lock()
```
rluders, line 101: I think that if we use `sync.RWMutex` we can use `RLock` and `RUnlock` here... it would be probably better.

pkg/config/configobserver/configmapobserver.go
```diff
@@ -0,0 +1,133 @@
...
+	cfg := cm.Data["config.yaml"]
+	err := yaml.Unmarshal([]byte(cfg), insightsConfig)
+	if err != nil {
+		return nil, err
+	}
+	return insightsConfig.ToConfig(), nil
+}
+
+func getConfigMap(ctx context.Context, kubeCli *kubernetes.Clientset) (*v1.ConfigMap, error) {
+	return kubeCli.CoreV1().ConfigMaps("openshift-insights").Get(ctx, insightsConfigMapName, metav1.GetOptions{})
```
rluders, line 132: Constant for `openshift-insights`

pkg/config/types.go
```diff
@@ -0,0 +1,89 @@
...
+			ProcessingStatusEndpoint:    i.DataReporting.ProcessingStatusEndpoint,
+		},
+	}
+	if i.DataReporting.Interval != "" {
+		interval, err := time.ParseDuration(i.DataReporting.Interval)
+		if err != nil {
+			klog.Errorf("Cannot parse interval time duration: %v. Using default value 2h", err)
+			interval = 2 * time.Hour
+		}
+		if interval <= 0 {
```
rluders, line 68: Can't we handle the `err != nil` and `interval <= 0` at the same if-statement? Or do we really want to have just the log for parsing error?

If we want, I would probably do something like:

```go
if err != nil {
    klog.Errorf(.....)
}

if err != nil || interval <= 0 {
    interval = 2 * time.Hour // I would also consider to convert it to a constant
}
```

pkg/config/types.go
```diff
@@ -0,0 +1,89 @@
...
+			interval = 2 * time.Hour
+		}
+		ic.DataReporting.Interval = interval
+	}
+	return ic
+}
+
+func (i *InsightsConfiguration) String() string {
+	s := fmt.Sprintf("interval=%s, "+
```
rluders, line 77: I would probably not use `+` to concatenate the strings, instead I would use "`", like:

```go
return fmt.Sprintf(`interval=%s,
upload_endpoint=%s,
storage_path=%s,
// and so...
`, ....)
```

rluders (changes requested):

pkg/config/configobserver/config_aggregator.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	if newCfg.DataReporting.StoragePath != "" {
+		defaultCfg.DataReporting.StoragePath = newCfg.DataReporting.StoragePath
+	}
+	c.config = defaultCfg
+}
+
+func (c *ConfigAggregator) Config() *config.InsightsConfiguration {
+	if c.usingInformer {
```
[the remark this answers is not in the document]
tremes: Why? Lock is on the level of the `merge` function.

pkg/config/configobserver/config_aggregator.go
```diff
@@ -0,0 +1,194 @@
...
+// new config map informer. When any configuration change is observed then all the listeners
+// are notified.
+func (c *ConfigAggregator) Listen(ctx context.Context) {
+	legacyCh, legacyCloseFn := c.legacyConfigurator.ConfigChanged()
+	cmCh, cmICloseFn := c.configMapInformer.ConfigChanged()
+	defer func() {
+		legacyCloseFn()
+		cmICloseFn()
+	}()
+
```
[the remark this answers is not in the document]
tremes: `<-ctx.Done()` case is handled below in the `select` statement so I don't understand much here

pkg/config/configobserver/configmapobserver.go
```diff
@@ -0,0 +1,133 @@
...
+	lock           sync.Mutex
+	kubeCli        *kubernetes.Clientset
+	insightsConfig *config.InsightsConfiguration
+	listeners      map[chan struct{}]struct{}
+}
+
+func NewConfigMapObserver(kubeConfig *rest.Config,
+	eventRecorder events.Recorder,
+	kubeInformer v1helpers.KubeInformersForNamespaces) (ConfigMapInformer, error) {
+	cmInformer := kubeInformer.InformersFor("openshift-insights").Core().V1().ConfigMaps().Informer()
```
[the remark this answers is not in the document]
tremes: yes we can :)

pkg/config/configobserver/configmapobserver.go
tremes: I think we don't need to log this in every reconciliation. The message that the configmap is not found is already logged.

pkg/config/configobserver/config_aggregator.go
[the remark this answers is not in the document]
tremes: Sorry I was wrong. Yes I think it should be locked on the level of `mergeUsingInformer` and `mergeStatically` functions.

pkg/config/configobserver/config_aggregator.go
```diff
@@ -0,0 +1,194 @@
...
+	c.merge(conf, newConf)
+}
+
+func (c *ConfigAggregator) mergeStatically() {
+	conf := c.legacyConfigToInsightsConfiguration()
+	c.config = conf
+	cm, err := c.kubeClient.CoreV1().ConfigMaps("openshift-insights").Get(context.Background(), insightsConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return
```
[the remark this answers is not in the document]
tremes: This is logged as:
```
configmapobserver.go:74] configmaps "insights-config" not found
```

pkg/config/configobserver/configmapobserver.go
tremes: See the https://github.com/openshift/insights-operator/pull/827/files#r1349992228 I think it's OK

pkg/config/configobserver/config_aggregator.go
```diff
@@ -0,0 +1,194 @@
...
+type Interface interface {
+	Config() *config.InsightsConfiguration
+	ConfigChanged() (<-chan struct{}, func())
+	Listen(ctx context.Context)
+}
+
+// ConfigAggregator is an auxiliary structure that should obviate the need for the use of
+// legacy secret configurator and the new config map informer
+type ConfigAggregator struct {
+	lock               sync.Mutex
```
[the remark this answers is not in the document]
tremes: Yeah I think it should be OK to change to RWMutext

pkg/config/configobserver/config_aggregator.go
```diff
@@ -0,0 +1,194 @@
...
+// precedence if it exists and is not empty.
+func (c *ConfigAggregator) mergeUsingInformer() {
+	newConf := c.configMapInformer.Config()
+	conf := c.legacyConfigToInsightsConfiguration()
+
+	if newConf == nil {
+		c.config = conf
+		return
+	}
+
```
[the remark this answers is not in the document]
tremes: yes, Good point

pkg/config/configobserver/configmapobserver.go
tremes: I am OK with constant, but this should not be configurable.

pkg/config/types.go
tremes: Thinking....maybe we could make it more simple, but at the same time the error message might be helpful here.

pkg/config/configobserver/config_aggregator.go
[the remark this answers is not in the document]
rluders: No worries. Mutex is always tricky.

pkg/config/configobserver/configmapobserver.go
rluders: Make sense.

pkg/config/configobserver/config_aggregator.go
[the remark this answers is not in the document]
rluders: I think that you are right here.

pkg/config/configobserver/config_aggregator.go
```diff
@@ -0,0 +1,194 @@
...
+	cmCh, cmICloseFn := c.configMapInformer.ConfigChanged()
+	defer func() {
+		legacyCloseFn()
+		cmICloseFn()
+	}()
+
+	for {
+		select {
+		case <-legacyCh:
+			c.notifyListeners()
```
rluders, line 160: I was thinking if isn't worth adding some log here to each case:

```go
klog.Infof("Legacy configuration change detected")
```

pkg/config/configobserver/config_aggregator.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (c *ConfigAggregator) mergeStatically() {
+	conf := c.legacyConfigToInsightsConfiguration()
+	c.config = conf
+	cm, err := c.kubeClient.CoreV1().ConfigMaps("openshift-insights").Get(context.Background(), insightsConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return
+		}
+		klog.Error(err)
```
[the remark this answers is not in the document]
tremes: I think this is addressed with https://github.com/openshift/insights-operator/pull/827#discussion_r1350038681

pkg/config/configobserver/config_aggregator.go
tremes: Yes it's an interesting suggestion, but I wouldn't complicate it with using the `reflect` package for now.

rluders: Well, you could probably just create two separate functions:

```go
func updateDurationFieldIfNotEmpty(field *time.Duration, newValue time.Duration) {
	if newValue != 0*time.Minute {
		*field = newValue
	}
}

func updateStringFieldIfNotEmpty(field *string, newValue string) {
	if newValue != "" {
		*field = newValue
	}
}
```

Also, I'm pretty sure that you could compare the `Interval` to `0` without having to do `0*time.Minute`.

pkg/config/configobserver/configmapobserver.go
rluders: How about this one?

pkg/config/configobserver/config_aggregator.go
tremes: Good point with zero, but I am not sure about the function suggestion and pointers usage. I think I still find it more straightforward as it is.....but yeah I am still thinking of a nicer solution.

rluders (approved):
/lgtm

tremes:
One test failure - I am working on the test fix.
/override ci/prow/e2e-gcp-ovn-techpreview

PR #861, opened 2023-11-09
OCPBUGS-23097: DVO gatherer - do the retry request as HEAD and not stream

tremes:
This updates the original `queryMetricsService` function in the DVO gatherer. I think the problem was that the timeout for the context was only 5 seconds and this context was also used for streaming the data. 5 seconds is not enough when streaming larger DVO metrics data. 

This changes the logic to use HEAD request to check whether the DVO service is up and running and then, if it's OK, start reading/streaming the data.

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Privacy

## Breaking Changes

Yes/No

## References

### DIFF
--- a/pkg/gatherers/clusterconfig/gather_dvo_metrics.go
+++ b/pkg/gatherers/clusterconfig/gather_dvo_metrics.go
@@ -137,15 +137,21 @@ func gatherDVOMetricsFromEndpoint(
 		http.DefaultClient,
 	)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ncaak (approved):
/lgtm

PR #863, opened 2023-11-10
CCXDEV-11387: feat(gather): gathering kubelet journal logs

rluders:
This PR extends the `node_logs` gathering to filter extra messages from cluster kubelet logs.

## Categories

- [ ] Bugfix
- [X] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `docs/insights-archive-sample/config/node/logs/ip-10-0-135-228.us-east-2.compute.internal.log` 🆙 

## Documentation

- `docs/gathered-data.md` 🆙 

## Unit Tests

Tests already exist.

## Privacy

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browse/CCXDEV-11387


rluders:
I need to check why the unit test is failing. 
/hold

pkg/gatherers/clusterconfig/gather_node_logs.go
```diff
@@ -105,24 +118,44 @@ func nodeLogString(ctx context.Context, req *rest.Request) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer in.Close()
+	defer func() {
+		_ = in.Close()
```
tremes, line 122: Should we log the error?

pkg/gatherers/clusterconfig/gather_node_logs.go
```diff
@@ -50,7 +63,7 @@ func (g *Gatherer) GatherNodeLogs(ctx context.Context) ([]record.Record, []error
 }
 
 func gatherNodeLogs(ctx context.Context, client corev1client.CoreV1Interface) ([]record.Record, []error) {
-	nodes, err := client.Nodes().List(ctx, metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/master"})
+	nodes, err := client.Nodes().List(ctx, metav1.ListOptions{})
```
tremes, line 66: The reason for this was to get only logs from master nodes, I guess. We want to filter everyrthing right?

tremes: I think this might mean much, much more data than before. Can't we filter just master node logs? I think usually there's more worker nodes than master nodes.

rluders: This is correct. The task doesn't limit the log gathering by the cluster type.

pkg/gatherers/clusterconfig/gather_node_logs.go
rluders: Yeah, this is probably a good idea. Let me fix it.

pkg/gatherers/clusterconfig/gather_node_logs.go
```diff
@@ -105,24 +118,44 @@ func nodeLogString(ctx context.Context, req *rest.Request) (string, error) {
...
+		`ContainerStateWaiting{Reason:ContainerCreating`,
+		`ContainersNotReady Message:containers with unready status`,
+		`MountVolume.MountDevice failed for volume`,
+		`kubernetes.io/csi: attacher.MountDevice failed to create newCsiDriverClient`,
+		`Unable to attach or mount volumes: unmounted volumes`,
+		`timed out waiting for the condition`,
+		`CreateContainerError: context deadline exceeded`,
+		`rpc error: code = ResourceExhausted desc = grpc: received message larger than max`,
+	}
+}
```
tremes, line 161: Minor: perhaps it would be slightly better to move this to some constant or share variable instead of creating the slice multiple times in the for loop (for each node)

rluders: That is a good point. I considered doing that but didn't want to leave it as a constant on the package scope. It should be called just once at the right before the loop to create a variable, then the GC can handle it, but it would require some refactoring - and adjusting unit tests as well. Not sure if I want to address it on this PR.

ncaak (approved):
/lgtm

PR #867, opened 2023-11-16
feat: conditional log filter

rluders:

Condition format example:
```json
{
  "conditions": [
    {
      "type": "cluster_version_matches",
      "cluster_version_matches": {
        "version": "> 4.12.0"
      }
    }
  ],
  "gathering_functions": {
    "pod_logs": {
      "resource_filter": {
        "namespace": ["openshift-monitoring"],
        "pod_name_regex_filter": "thanos-"
      },
      "log_filter": {
        "messages_to_seach": [
          "connect: connection refused"
        ],
        "tail_lines": 100,
        "previous": false
      }
    }
  }
}
```

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browse/CCXDEV-9909

### DIFF
--- a/docs/conditional-gatherer/README.md
+++ b/docs/conditional-gatherer/README.md
@@ -1,185 +1,137 @@
-# Conditional Gatherer
+# Conditional Gathering
 
-Conditional gatherer is a special gatherer which uses a set of rules describing which gathering functions to activate.
-The conditional rules are defined in the [insights-operator-gathering-conditions GitHub repository](https://github.com/RedHatInsights/insights-operator-gathering-conditions). This content is consumed and exposed by the [insights-operator-gathering-conditions-service](https://github.com/RedHatInsights/insights-operator-gathering-conditions-service). A new version of the `insights-operator-gathering-conditions` requires a new release version of the `insights-operator-gathering-conditions-service`.
-The Insights Operator connects to this service and consumes the conditional rules from it. The connection endpoint is defined in the [pod.yaml config file](../../config/pod.yaml) in the `conditionalGathererEndpoint` attribute (the value can be overriden in the `support` secret). Authentication is required and the `pull-secret` token is used for this purpose.
+The conditional gathering system is a specialized gatherer that utilizes a set of rules to determine which gathering functions to activate. These rules are defined in the [insights-operator-gathering-conditions GitHub repository](https://github.com/RedHatInsights/insights-operator-gathering-conditions) and are consumed and exposed by the [insights-operator-gathering-conditions-service](https://github.com/RedHatInsights/insights-operator-gathering-conditions-service). Any updates or changes to the conditional rules necessitate a new release version of the `insights-operator-gathering-conditions-service`.
 
-## Validation of the conditional rules
+The Insights Operator establishes a connection with this service to consume the conditional rules, and the connection endpoint is specified in the [pod.yaml config file](../../config/pod.yaml) under the `conditionalGathererEndpoint` attribute. It's important to note that the value of this attribute can be overridden in the `support` secret. Authentication is mandatory, and the `pull-secret` token is utilized for this purpose.
 
-The Insights Operator internally validates the conditional rules JSON against the JSON schema. [The schema](../../pkg/gatherers/conditional/gathering_rules.schema.json) is available in the `pkg/gatherers/conditional` package. You can see that this schema refers to the second [gathering_rule.schema.json](../../pkg/gatherers/conditional/gathering_rule.schema.json). This second schema defines the more important restrictions on the specific rules. **If the validation fails, no conditional data will be gathered!**!
+If you are looking for extra develoment information, please, [check this document](./development.md).
 
-The following are some examples of validation failures (which will show up in the log):
+## Conditions and Gathering Functions
 
-Non-existing gathering function:
-```
-E0808 16:29:51.864716  241084 parsing.go:22] skipping a rule because of an error: unable to create params for conditional.GatheringFunctionName: containers_log {[{alert_is_firing 0xc0004639c0 <nil>}] map[]}
-```
+The conditional gathering system allows you to tailor data collection based on specific conditions and gathering functions. Conditions define when data gathering should be triggered, and gathering functions specify how data should be collected.
 
-Missing alert name:
-```
-E0808 16:38:09.453211  242327 periodic.go:137] conditional failed after 14ms with: got invalid config for conditional gatherer: 0.conditions.0: Must validate at least one schema (anyOf), 0.conditions.0: alert is required
-```
+Here is an example JSON structure for defining conditions:
 
-Gathering function missing required parameter:
-```
-E0808 16:41:35.184585  242636 periodic.go:137] conditional failed after 20ms with: got invalid config for conditional gatherer: 0.gathering_functions.containers_logs.tail_lines: Must be greater than or equal to 1
-```
-
-Failed to parse the provided cluster version:
-
-```
-E0809 10:02:16.383643   37430 conditional_gatherer.go:140] error checking conditions for a gathering rule: Could not parse Range "4-11.12": Could not parse version "4-11.12" in "4-11.12": No Major.Minor.Patch elements found
-```
-
-One of the common conditions type (see below) is the `alert_is_firing`. This condition depends on availability of Prometheus alerts - i.e. connection to the in-cluster Prometheus instance. If the connection is not available, this may manifest in the log as follows for example:
-
-```
-E0809 11:56:48.491346   46838 conditional_gatherer.go:226] unable to update alerts cache: open /var/run/configmaps/service-ca-bundle/service-ca.crt: no such file or directory
-```
-
-If the error message `there are no conditional rules` is shown, that means that there are no conditional rules or that the format used to assign the conditional rules had an invalid json format. Log message will look something like:
-
-```
-E0915 09:01:03.349317   26966 periodic.go:158] conditional failed after 2ms with: got invalid config for conditional gatherer: there are no conditional rules
-```
-
-## Basic structure of the conditional rules
-
-From the schemas mentioned above, you can see that each rule consists of `conditions` array and `gathering_functions` object. The `conditions` array defines conditions that must be met and the `gathering_functions` object tells what functions are called in the Insights Operator source code. The current conditions are defined in the [`pkg/gatherers/conditional/conditions.go`](../../pkg/gatherers/conditional/conditions.go) (see the `ConditionType` and its use) and the gathering functions are defined in the [`pkg/gatherers/conditional/gathering_functions.go`](../../pkg/gatherers/conditional/gathering_functions.go)
-
-
-## Manual Testing
-
-To test that conditional gatherer provides some data, follow the next steps:
-
-1. Downscale CVO, CMO and Prometheus CRD:
-```bash
-oc scale deployment -n openshift-cluster-version cluster-version-operator --replicas=0
-oc scale deployment -n openshift-monitoring cluster-monitoring-operator --replicas=0
-oc patch prometheus -n openshift-monitoring k8s --type "json" -p '[{"op": "replace", "path": "/spec/replicas", "value": 1}]'
-```
-
-2. Backup prometheus rules:
-```bash
-oc get prometheusrule -n openshift-cluster-samples-operator samples-operator-alerts -o json > prometheus-rules.back.json
-```
-
-3. Make SamplesImagestreamImportFailing alert to fire by setting `SamplesImagestreamImportFailing`'s
-`expr` value to `1 > bool 0` and `for` to `1s`:
-```bash
-echo '{
-    "apiVersion": "monitoring.coreos.com/v1",
-    "kind": "PrometheusRule",
-    "metadata": {
-        "name": "samples-operator-alerts",
-        "namespace": "openshift-cluster-samples-operator"
-    },
-    "spec": {
-        "groups": [
-            {
-                "name": "SamplesOperator",
-                "rules": [
-                    {
-                        "alert": "SamplesImagestreamImportFailing",
-                        "annotations": {
-                            "message": "Always firing"
-                        },
-                        "expr": "1 > bool 0",
-                        "for": "1s",
-                        "labels": {
-                            "severity": "warning"
-                        }
-                    }
-                ]
-            }
-        ]
-    }
-}' | oc apply -f -
-```
-
-4. Wait for the alert to fire:
-```bash
-export PROMETHEUS_HOST=$(oc get route -n openshift-monitoring prometheus-k8s -o jsonpath='{@.spec.host}')
-export INSECURE_PROMETHEUS_TOKEN=$(oc get secret $(oc get sa prometheus-k8s -n openshift-monitoring -o json | jq .secrets[0].name | tr --delete \") -n openshift-monitoring -o json | jq .metadata.annotations.\"openshift.io/token-secret.value\" | tr --delete \")
-curl -g -s -k -H 'Cache-Control: no-cache' -H "Authorization: Bearer $INSECURE_PROMETHEUS_TOKEN" "https://$PROMETHEUS_HOST/api/v1/query" --data-urlencode 'query=ALERTS{alertstate="firing",alertname="SamplesImagestreamImportFailing"}' | jq ".data.result[]"
-```
-
-The output should be:
 ```json
 {
-  "metric": {
-    "__name__": "ALERTS",
-    "alertname": "SamplesImagestreamImportFailing",
-    "alertstate": "firing",
-    "severity": "critical"
-  },
-  "value": [
-    1652363876.855,
-    "1"
-  ]
+    "conditions": [
+        {
+            "type": "cluster_version_matches",
+            "cluster_version_matches": {
+                "version": "> 4.12.0"
+            }
+        }
+    ],
+    "gathering_functions": {
+        "pod_logs": {
+            "resource_filter": {
+                "namespaces": ["openshift-monitoring"],
+                "pod_name_regex_filter": "thanos-"
+            },
+            "log_filter": {
+                "messages_to_search": ["connect: connection refused"],
+                "tail_lines": 100,
+                "previous": false
+            }
+        }
+    }
 }
 ```
 
-5. Make metrics work by forwarding the endpoint and setting INSECURE_PROMETHEUS_TOKEN environment variable:
-```bash
-export INSECURE_PROMETHEUS_TOKEN=$(oc get secret $(oc get sa prometheus-k8s -n openshift-monitoring -o json | jq .secrets[0].name | tr --delete \") -n openshift-monitoring -o json | jq .metadata.annotations.\"openshift.io/token-secret.value\")
-```
-```bash
-# run this command in a separate terminal
-sudo kubefwd svc -n openshift-monitoring -d openshift-monitoring.svc -l app.kubernetes.io/instance=k8s --kubeconfig $KUBECONFIG
-```
+## Conditions
 
-6. Run the operator and wait for an archive containing `conditional/` directory.
+Conditions determine when data gathering should be triggered within the conditional data gathering system. These conditions are configured based on specific events or states within the OpenShift cluster.
 
-7. Restore the backup:
-```bash
-oc apply -f prometheus-rules.back.json
-```
+For example, the `cluster_version_matches` condition allows you to trigger data gathering based on the OpenShift cluster version meeting specific criteria. Understand and utilize these conditions to precisely control when data collection activities occur, ensuring they align with your cluster's operational requirements.
 
-8. Fix CVO back
-```bash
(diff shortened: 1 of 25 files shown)


pkg/gatherers/conditional/gather_pod_logs.go
```diff
@@ -0,0 +1,70 @@
...
+func (g *Gatherer) gatherPodLogs(ctx context.Context, params GatherPodLogsParams, coreClient v1.CoreV1Interface) ([]record.Record, []error) {
+	// TODO filter namespace, it only allows `openshift-*` and `kubernetes-*`
+	records, err := common.CollectLogsFromContainers(
+		ctx,
+		coreClient,
+		params.ResourceFilter,
+		params.LogMessageFilter,
+		func(namespace string, podName string, containerName string) string {
+			return fmt.Sprintf(
+				"%v/logs/%v/%v/%v.log",
```
rluders, line 60: Self-review: this is not what I want... I need to fix it :)
The desired paths would be:

- conditional/pod_logs/{namespace}/{pod-name}/{container}/current.log
- conditional/pod_logs/{namespace}/{pod-name}/{container}/previous.log

pkg/gatherers/clusterconfig/gather_node_logs.go
```diff
@@ -119,7 +121,9 @@ func nodeLogString(ctx context.Context, req *rest.Request) (string, error) {
 	messagesToSearch := []string{
 		"E\\d{4} [0-9]{1,2}:[0-9]{1,2}:[0-9]{1,2}", //  Errors from log
 	}
-	return common.FilterLogFromScanner(scanner, messagesToSearch, true, func(lines []string) []string {
+	messagesRegex := regexp.MustCompile(strings.Join(messagesToSearch, "|"))
```
tremes, line 124: It would be better to compile the regex once and not for every node.

pkg/gatherers/conditional/gather_pod_logs.go
tremes: We should take a look at must-gather archive structure and try do the same in the future.

pkg/gatherers/conditional/gather_pod_logs.go
```diff
@@ -0,0 +1,70 @@
...
+				return nil, []error{err}
+			}
+			coreClient := kubeClient.CoreV1()
+			return g.gatherPodLogs(ctx, params, coreClient)
+		},
+	}, nil
+}
+
+func (g *Gatherer) gatherPodLogs(ctx context.Context, params GatherPodLogsParams, coreClient v1.CoreV1Interface) ([]record.Record, []error) {
+	// TODO filter namespace, it only allows `openshift-*` and `kubernetes-*`
```
tremes, line 52: This should be enforced by the JSON schema validation IMO.

pkg/gatherers/common/gather_logs.go
```diff
@@ -79,23 +80,33 @@ func CollectLogsFromContainers( //nolint:gocyclo
 	var skippedContainers int
 	var records []record.Record
 
+	var podNameRegex *regexp.Regexp
+	if len(containersFilter.PodNameRegexFilter) > 0 {
```
tremes, line 84: I think this should be validated/enforced by the JSON schema validation or? In other words we require some identificator of a Pod name.

pkg/gatherers/conditional/gathering_rule.schema.json
```diff
@@ -217,6 +217,86 @@
                             "pattern": "^[a-zA-Z0-9_]{1,128}$"
                         }
                     }
+                },
+                "^pod_logs$": {
+                    "type": "object",
+                    "title": "GatherPodLogsParams",
+                    "required": [
+                        "resource_filter"
```
tremes, line 225: Shouldn't the `log_filter` be required as well or at least the `message_to_search` should be required IMO

pkg/gatherers/clusterconfig/gather_node_logs.go
rluders: Oh yeah, for sure. I forgot to add the `TODO` here. I want to do something similar to what I did for `CollectLogsFromContainers`. Thanks for pointing it out.

pkg/gatherers/conditional/gather_pod_logs.go
rluders: Ohhhhhh!!! That is actually a pretty good idea.

pkg/gatherers/conditional/gathering_rule.schema.json
rluders: Maybe? I'm trying to remember if we have some log gatherer that doesn't filter the errors. 🤔  But, I'll check.

pkg/gatherers/common/gather_logs.go
rluders: Not sure If I got it right. I just added this validation to not compile the regex with `nil` value... maybe there is no difference? Or are you saying that this should be required?

tremes: Yes I think it should be required.

pkg/gatherers/common/gather_logs.go
```diff
@@ -156,38 +178,35 @@ func filterLogs(
...
 			matches := messagesRegexp.MatchString(line)
 			if matches {
 				result = append(result, line)
 			}
-		} else {
-			for _, messageToSearch := range messagesToSearch {
-				if strings.Contains(strings.ToLower(line), strings.ToLower(messageToSearch)) {
-					result = append(result, line)
-				}
+			continue
```
tremes, line 197: So when there's some `messageRegexp` provided then the `messagesToSearch` are not taken into account at all right? If so then I would probably split it in two functions or at least provide some better documentation for the function.

pkg/gatherers/common/gather_logs.go
```diff
@@ -156,38 +178,35 @@ func filterLogs(
...
+		for _, messageToSearch := range messagesToSearch {
+			if strings.Contains(strings.ToLower(line), strings.ToLower(messageToSearch)) {
+				result = append(result, line)
 			}
 		}
+
+		if len(messagesToSearch) == 0 {
+			result = append(result, line)
+			continue
+		}
```
tremes, line 209: I don't understand the need of this 4 lines of code. Can you elaborate please?

pkg/gatherers/common/gather_logs.go
rluders: OK, but... If we make it required it would cause some issues to existing log gatherers.

pkg/gatherers/common/gather_logs.go
rluders: I'll remove this block... We never collect any log without filtering it and we probably never want to do it.

rluders: Oh, actually we need this block, but I kind of changed it a little bit. There is one gather that kind of doesn't filter the log, actually, it gets just the last 50 lines.

pkg/gatherers/common/gather_logs.go
rluders: @tremes I extended the log functions a little bit now and also added some comments. Please, check it. I think it looks nice...

pkg/gatherers/conditional/gathering_rule.schema.json
rluders: (I'm double checking it, but probably no problem to leave it as required for conditions)

pkg/gatherers/common/gather_logs.go
rluders: We can't make it required, if we do we can't move the current log gathers to the conditional, but, on the other hand, we can make it required at the conditional scheme if we decide to not migrate the existing log gatherers to the condition or change them.

pkg/gatherers/common/gather_logs.go
```diff
@@ -17,25 +17,32 @@ import (
...
-	IsRegexSearch    bool
-	SinceSeconds     int64
-	LimitBytes       int64
-	TailLines        int64
-	Previous         bool
+	MessagesToSearch []string `json:"messages_to_search,omitempty"`
+	IsRegexSearch    bool     `json:"is_regex_search,omitempty"`
+	SinceSeconds     int64    `json:"since_seconds,omitempty"`
+	LimitBytes       int64    `json:"limit_bytes,omitempty"`
+	TailLines        int64    `json:"tail_lines,omitempty"`
```
tremes, line 36: Does it makes sense to still have `SinceSeconds` and `TailLines` parameters here?

pkg/gatherers/common/gather_logs.go
```diff
@@ -17,25 +17,32 @@ import (
...
+	Previous         bool     `json:"previous,omitempty"`
 }
 
+// FilterLogCallbackFunc is used to apply extra filtering to log lines
+type FilterLogCallbackFunc func(lines []string) []string
+
+// FilterLogOption is used to apply the filter engine (e.g: Regex, Substring)
+type FilterLogOption func(line string) bool
```
tremes, line 44: It's not clear to me what the return value means? Filter or not filter?

pkg/gatherers/common/gather_logs.go
rluders: This is for compatibility with the existing gatherers, if we will not use the conditions we can remove it by limiting the schema.

pkg/gatherers/common/gather_logs.go
rluders: // FilterLogOption is a functional option used to apply the filtering engine, such as Regex or Substring matching,
// to determine whether a given log line satisfies the specified condition.
// When applied, it returns true if the provided regex or substring is found in the log line, indicating a match.
// Conversely, it returns false if no match is detected.

pkg/gatherers/common/gather_logs.go
rluders: I'll track this as part of the Epic. I think that this is something that we should only remove from the scheme, not from the code.

pkg/gatherers/conditional/gathering_rule.schema.json
rluders: I'll move this to the epic. I think that we will need to validate the schema definition with the new requirements.

rluders:
Converting it to draft for now, since we need some follow-up discussions.

pkg/gatherers/common/gather_logs.go
```diff
@@ -17,25 +17,37 @@ import (
...
-	Namespace                string
-	LabelSelector            string
-	FieldSelector            string
-	ContainerNameRegexFilter string
-	MaxNamespaceContainers   int
+// LogResourceFilter allows you to filter containers
+type LogResourceFilter struct {
+	// Deprecated: Namespace would be replaced by Namespaces in the future
+	Namespace                string   `json:"namespace"`
+	Namespaces               []string `json:"namespaces,omitempty"`
```
tremes, line 24: It's not clear to me how this is supposed to work. Do you plan to iterate over the namespaces for the label selector value?
I think it will be essential to compile a list of messages for a given pod for a given namespace. I'm not sure how you're going to accomplish that. Can you please elaborate?

docs/conditional-gatherer/README.md
```diff
@@ -1,185 +1,137 @@
...
+In this example:
 
-7. Restore the backup:
-```bash
-oc apply -f prometheus-rules.back.json
-```
+- `resource_filter.namespaces`: Specifies the namespaces to search for resources.
+- `resource_filter.pod_name_regex_filter`: Uses a regex to filter pod names.
+- `log_filter.messages_to_search`: Lists messages to search for in the logs.
+- `log_filter.tail_lines`: Specifies the number of lines to retrieve from the end of logs.
```
jholecek-rh, line 109: How does it work with `messages_to_search`?

docs/conditional-gatherer/README.md
```diff
@@ -1,185 +1,137 @@
...
-```bash
-export INSECURE_PROMETHEUS_TOKEN=$(oc get secret $(oc get sa prometheus-k8s -n openshift-monitoring -o json | jq .secrets[0].name | tr --delete \") -n openshift-monitoring -o json | jq .metadata.annotations.\"openshift.io/token-secret.value\")
-```
-```bash
-# run this command in a separate terminal
-sudo kubefwd svc -n openshift-monitoring -d openshift-monitoring.svc -l app.kubernetes.io/instance=k8s --kubeconfig $KUBECONFIG
-```
-
-6. Run the operator and wait for an archive containing `conditional/` directory.
+In this example:
```
jholecek-rh, line 104: Examples are useful, but I am rather looking for a reference guide that provides a comprehensive documentation of each gathering function and all its parameters.

jholecek-rh: Nvm, a complete list is in one of the next sections.

docs/conditional-gatherer/README.md
```diff
@@ -1,185 +1,137 @@
...
-}' | oc apply -f -
-```
+#### Log filter
+The `log_filter` structure enables users to filter log messages. It includes the following fields:
 
-3. Enjoy your conditional rules from the stage endpoint
+| Field                   | Type   | Description                                           | Required |
+|-------------------------|--------|-------------------------------------------------------|----------|
+| `messages_to_search`    | array  | Array of messages to search for.                      | Required |
+| `is_regex_search`       | bool   | Boolean flag for regex search.                        | Optional |
```
jholecek-rh, line 133: I would prefer to have `messages` and `messages_regex` parameters over having an optional parameter whose sole purpose is to modify the behavior of another optional parameter.  I would be fine with optional parameters modifying the behavior of mandatory parameters, but this is not the case.

docs/conditional-gatherer/development.md
```diff
@@ -0,0 +1,181 @@
+# Conditional Gatherer Development
+
+## Validation of the conditional rules
+
+The Insights Operator internally validates the conditional rules JSON against the JSON schema. [The schema](../../pkg/gatherers/conditional/gathering_rules.schema.json) is available in the `pkg/gatherers/conditional` package. You can see that this schema refers to the second [gathering_rule.schema.json](../../pkg/gatherers/conditional/gathering_rule.schema.json). This second schema defines the more important restrictions on the specific rules. **If the validation fails, no conditional data will be gathered!**!
```
jholecek-rh, line 5: > If the validation fails, no conditional data will be gathered!

Does this mean that an error in one gathering rule prevents all other gathering rules from being executed?

pkg/gatherers/common/gather_logs.go
rluders: > It's not clear to me how this is supposed to work. Do you plan to iterate over the namespaces for the pod name filter value?

Instead, one condition gathers logs from containers under one namespace it can get from multiple namespaces. This was one of the requests to reduce the need to create one condition per namespace.

> I think it will be essential to compile a list of messages for a given pod for a given namespace. I'm not sure how you're going to accomplish that. Can you please elaborate?

This is simple, I already implemented it on this PR:

```go
for _, namespace := range containersFilter.Namespaces {
```

It just iterates all the given Namespaces and fetches the logs. (yes, need probably some optimization as next step, but first we need to validate the whole implementation)

docs/conditional-gatherer/README.md
```diff
@@ -1,185 +1,137 @@
...
+| `since_seconds`         | int64  | Time duration for log search (in seconds).            | Optional |
+| `limit_bytes`           | int64  | Maximum size of log data to collect.                  | Optional |
+| `tail_lines`            | int64  | Number of lines to retrieve from the end of logs.     | Optional |
```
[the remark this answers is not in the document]
rluders: I'll update the description but removing parameters, at this point, is out of scope.

docs/conditional-gatherer/development.md
rluders: Yep. We can address it on a separate task, but sure, we can change it. I'll probably drag this as a 'next step' point of improvement, since we may also need to track the failings conditions somehow.

docs/conditional-gatherer/README.md
rluders: It doesn't impact the `messages_to_seach`. It only limits the number of lines fetched from the Openshift API per container.

docs/conditional-gatherer/README.md
rluders: Not sure if part of the documentation existing feature task, however, I'm taking these notes for future improvements, since it needs some effort to change.

pkg/gatherers/common/gather_logs.go
tremes: Sorry but I can't see any benefit of this approach. Why should a rule developer provide a list of namespaces? I assume he/she should know the corresponding namespace for a particular pod.

jholecek-rh: I can imagine two use cases where we would need to specify multiple namespaces:

* A platform pod has moved (or its namespace has been renamed) between OCP versions. This would be a fixed-length list of specific namespace names. IIUC, the current implementation supports this use case. On the other hand, we could solve this use case by splitting the gathering rule into multiple ones - one for each namespace. This case is quite rare so it shouldn't increase the config size too much.
* There is a grey zone of optional Red Hat-developed operators which customers can install with OLM into namespaces with custom names. In this case, we would need to search for the pod(s) in namespaces specified by other means then their names. Without some standardization on detecting such namespaces, these namespaces will be outside the hard-coded boundaries of what Insights Operator will be willing to gather. We don't need to worry about this use case right now, imho.

Having said that, how/where does the current implementation define the hard-coded boundaries?

docs/conditional-gatherer/README.md
jholecek-rh: > It doesn't impact the messages_to_search.

IIUC, parameters like `tail_lines` specify a portion of the log where lines matching `messages_to_search` will be searched.

Note that this is very different from "IO gathers the last `tail_lines` lines matching `messages_to_search`". The current documentation allows this interpretation.

docs/conditional-gatherer/README.md
jholecek-rh: I don't understand why you are so opposed to iterating over the rule specification. I don't care about the code that implements the gathering rules at this point. Users should not be expected to read the Insights Operator source code to be able to write gathering rules.

rluders: OK. Let's do it... We can split it into `messages` and `messages_regex` for the conditions, but basically, the code would still behavior the same, it would be only an interface to interact with the code that we already have, so we can mitigate the effort to change it.

pkg/gatherers/common/gather_logs.go
rluders: Currently, we are limited by `openshift-*` and `kube-*` namespaces.

> customers can install with OLM into namespaces with custom names

And based on this quote it seems that this boundary could be a problem... right?

docs/conditional-gatherer/README.md
jholecek-rh: I repeat: I don't care about the code in this PR. The code doesn't do what we need. If nothing else, it does not provide the logs in the format that was requested. That alone might mean that you will need to reconsider the whole gathering approach.

rluders: I know that you don't care about the code, but I do. 😆 

I see no reason why `message` + `message_regex` instead `message_to_search` + `is_regex_seach` would not "provide the logs in the format that was requested". But, as I said, looking a the code it is possible to adjust to support both parameters `message` + `message_regex` instead of the current implementation, and from the implementation perspective it would not cause any impact, since under the hood it would still using the same implementation.

So, what I'm saying is: yes, let's change to `message` and `message_regex`, it doesn't seem to be such a difficult implementation as I was thinking.

docs/conditional-gatherer/README.md
jholecek-rh: P.S. Also note that Ivan requested to be able to gather a log tail (e.g. for terminating pods). So it could also mean "gather all lines that match messages and a log tail specified by the other parameters". 

I don't understand what your emoji reaction means. Are you planning to take any action based on this feedback?

pkg/gatherers/common/gather_logs.go
jholecek-rh: I would say let's limit the scope to 1 namespace per resource filter and `openshift-*` and `kube-*` namespaces for now.

We will need to deal with multiple rules for one pod anyway so we can solve the first use case by having multiple rules if we need it.

I wouldn't worry about the second use case right now. It is a separate issue (and probably a rather large one).

docs/conditional-gatherer/README.md
jholecek-rh: You are providing feedback on how my proposal affects the current implementation. I am expecting feedback on the API that the gathering rules comprise - what is more intuitive/clear/powerful?

Having said that, we could also say that the messages have to be regexes and handle non-regex patters on the gathering rule creation side.

tremes:
@rluders I think we can close this PR right?

PR #868, opened 2023-11-16
feat: gather helm information

rluders:

## Categories

- [ ] Bugfix
- [X] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

Yes/No

## References


pkg/gatherers/workloads/workloads_gatherer.go
```diff
@@ -36,6 +38,11 @@ func (g *Gatherer) GetGatheringFunctions(context.Context) (map[string]gatherers.
 				return g.GatherWorkloadInfo(ctx)
 			},
 		},
+		"helm_info": {
+			Run: func(ctx context.Context) ([]record.Record, []error) {
+				return g.GatherHelmInfo(ctx)
+			},
+		},
```
tremes, line 45: workloads gatherer runs only each 12h. Is it desired for this data?

pkg/gatherers/workloads/gather_helm_info.go
```diff
@@ -0,0 +1,129 @@
...
+				if labels[key] != value {
+					isHelmManaged = false
+					break
+				}
+			}
+
+			if isHelmManaged {
+				records = append(records, record.Record{
+					Name: fmt.Sprintf("config/helm_info/%s/%s", resource.Resource, item.GetName()),
+					Item: record.JSONMarshaller{Object: item},
```
tremes, line 118: Do we need to gather the full object definition? I guess this would be data explosion in the archive....

pkg/gatherers/workloads/workloads_gatherer.go
rluders: It is not, it needs to be executed once a day. I'm moving it to a separated gather.

tremes: hmm the question is if it's worth creating a new gatherer....maybe we should check with processing team if there's some extra logic for those data which is not in every archiuve

docs/insights-archive-sample/config/helmchart_info.json
```diff
@@ -0,0 +1,19 @@
...
+            "name": "nodejs",
+            "version": "",
+            "resources": { "deployments": 1, "replicasets": 1, "services": 1 }
+        }
+    ],
+    "e976fcc461dc1b1ad177c083135393e90ded18b3707987d4b6adf78e86e687ed": [
+        {
+            "name": "nodejs",
+            "version": "",
+            "resources": { "deployments": 1, "replicasets": 1, "services": 1 }
```
tremes, line 16: As I mentioned. If there's no use case for the actual resource type, then I would not include it.

rluders: I prefer to keep this one since it was requested by the client and it also can help to identify the number of apps using the helm on the same namespace.

pkg/gatherers/workloads/workloads_gatherer.go
rluders: We will keep using the `workloads`

pkg/gatherers/workloads/gather_helm_info.go
```diff
@@ -0,0 +1,171 @@
...
+// `workloads/helmchart_info`
+//
+// ### Released version
+// - 4.15.0
+//
+// ### Backported versions
+// None
+func (g *Gatherer) GatherHelmInfo(ctx context.Context) ([]record.Record, []error) {
+	dynamicClient, err := dynamic.NewForConfig(g.gatherKubeConfig)
+	// dynamicClient, err := dynamic.NewForConfig(g.gatherProtoKubeConfig)
```
rluders, line 45: Removing it next commit

pkg/gatherers/workloads/gather_helm_info.go
```diff
@@ -0,0 +1,170 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/klog/v2"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+const LabelChartNameKey = "helm.sh/chart"
```
tremes, line 21: Does it need to be exported?

pkg/gatherers/workloads/gather_helm_info.go
```diff
@@ -0,0 +1,170 @@
...
+			klog.V(2).Infof("Unable to list %s resource due to: %s", resource, err)
+			errs = append(errs, err)
+			continue
+		}
+
+		for _, item := range items.Items {
+			labels := item.GetLabels()
+			if labels == nil {
+				labels = map[string]string{}
+			}
```
tremes, line 85: I don't understand this condition. Wouldn't make sense to just continue in case of `labels == nil`? The map in this case is not initialized, so I assume it will fail later or return an empty string.

pkg/gatherers/workloads/gather_helm_info.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	var errs []error
+	var records []record.Record
+	helmList := newHelmChartInfoList()
+
+	for _, resource := range resources {
+		listOptions := metav1.ListOptions{LabelSelector: "app.kubernetes.io/managed-by=Helm"}
+
+		items, err := dynamicClient.Resource(resource).List(ctx, listOptions)
```
tremes, line 71: Would it make sense to exclude OpenShift and Kubernetes namespaces?

pkg/gatherers/workloads/gather_helm_info.go
rluders: Forgot to remove this `if`. Thanks.

pkg/gatherers/workloads/gather_helm_info.go
rluders: That is a good question. Let me investigate it.

pkg/gatherers/workloads/gather_helm_info.go
```diff
@@ -0,0 +1,167 @@
...
+			continue
+		}
+
+		for _, item := range items.Items {
+			labels := item.GetLabels()
+
+			// Anonymize the namespace to make it unique identifier
+			hash, err := createHash(item.GetNamespace())
+			if err != nil {
+				klog.Errorf("unable to hash the HelmChart's name '%s': %v", labels[labelChartNameKey], err)
```
tremes, line 87: nit: it's not HelmChart name hash, but namespace name hash right

pkg/gatherers/workloads/gather_helm_info_test.go
```diff
@@ -0,0 +1,232 @@
...
+							"metadata": map[string]interface{}{
+								"name":      "myservice",
+								"namespace": "mynamespace",
+								"labels": map[string]interface{}{
+									"app.kubernetes.io/managed-by": "Helm",
+									"helm.sh/chart":                "postgres-9.0.0",
+								},
+							},
+							"spec": map[string]interface{}{},
+						},
```
tremes, line 122: It would probably be good to add following test cases:

- object with empty labels is not processed/added
- object with different labels than `"helm.sh/chart"` is not processed added

rluders: I can create one single test to validate all the unprocessed data, it would be easier than repeating the test data...

ncaak (approved):
/lgtm

PR #888, opened 2023-12-18
NO-JIRA: (refactor) job completion uses event instead polling

rluders:
This PR replaces event polling from periodic job completion and replaces it with event listening.

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive

N/A

## Documentation

N/A

## Unit Tests

N/A

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-10669

### DIFF
--- a/pkg/controller/periodic/job.go
+++ b/pkg/controller/periodic/job.go
@@ -3,14 +3,12 @@ package periodic
 import (
 	"context"
 	"fmt"
-	"time"
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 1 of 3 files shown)


ncaak (approved):

pkg/controller/periodic/job.go
```diff
@@ -108,19 +106,33 @@ func (j *JobController) CreateGathererJob(ctx context.Context, dataGatherName, i
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case event, ok := <-watcher.ResultChan():
+			if !ok {
+				return fmt.Errorf("watcher channel was closed unexpectedly")
+			}
+			if event.Type == watch.Error {
+				return fmt.Errorf("watcher received error event: %v", event.Object)
+			}
+			job := event.Object.(*batchv1.Job)
```
tremes, line 129: What happens when this fails please?

tremes:
This is not working as is, because the `watch` permission (for `batch` group) is miising for the corresponding role.

pkg/controller/periodic/job.go
```diff
@@ -108,19 +106,33 @@ func (j *JobController) CreateGathererJob(ctx context.Context, dataGatherName, i
...
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case event, ok := <-watcher.ResultChan():
+			if !ok {
+				return fmt.Errorf("watcher channel was closed unexpectedly")
+			}
+			if event.Type == watch.Error {
+				return fmt.Errorf("watcher received error event: %v", event.Object)
```
tremes, line 127: I am not sure what the error event type is, but I think we don't need to log the complete object definition.

rluders: This one is a little bit trickier... I don't have much information about the `event.Object` or even the `event` that can be easy to log to identify it. That is why I'm recording the complete object definition. Let me try to find a better way to log it one more time...

tremes: Aren't we interested only the `modified` event? Perhaps we could ignore otherwise...

pkg/controller/periodic/job.go
tremes: Unfortunately this doesn't help. https://play.golang.com/p/l7mo2jQifAT

rluders: OK. I was reading the stuff, and it would not be a problem, once the watcher is looking just for Jobs:

```go
j.kubeClient.BatchV1().Jobs(insightsNamespace).Watch(ctx, metav1.ListOptions{})
```

The `event.Object` would be always a `runtime.Object` that can be converted to `batchv1.Job`. So, there is no reason for us to worry about it.

tremes: Yes I know it should not happen, but it's better to cover all possible scenarios than to be suddenly surprised in my opinion.

pkg/controller/periodic/periodic.go
```diff
@@ -346,15 +346,15 @@ func (c *Controller) runJobAndCheckResults(ctx context.Context, dataGather *insi
...
+	err = c.jobController.WaitForJobCompletion(ctx, gj.Name)
 	if err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
 			klog.Errorf("Failed to read job status: %v", err)
 			return
 		}
 		klog.Error(err)
 	}
-	klog.Infof("Job completed %s", gj.Name)
+	klog.Infof("Job completed %s", "")
```
tremes, line 357: I think the name still should be here.

pkg/controller/periodic/job.go
```diff
@@ -123,19 +123,20 @@ func (j *JobController) WaitForJobCompletion(ctx context.Context, job *batchv1.J
...
 			if job == nil {
-				return fmt.Errorf("cannot cast event objeto into job: %v", event.Object)
+				return fmt.Errorf("failed to cast job event: %v", event.Object)
 			}
 			if job.Status.Succeeded > 0 {
 				return nil
 			}
 			if job.Status.Failed > 0 {
-				return fmt.Errorf("job %s failed", job.Name)
+				return fmt.Errorf("job %s failed: %s", job.Name, job.Status.Conditions[0].Message)
```
tremes, line 139: Sorry, but this doesn't make much sense to me. Why the first condition message? What if there are no condtions?

pkg/controller/periodic/job.go
```diff
@@ -108,19 +106,38 @@ func (j *JobController) CreateGathererJob(ctx context.Context, dataGatherName, i
...
+			if !ok {
+				return fmt.Errorf("watcher channel was closed unexpectedly")
+			}
+
+			if event.Type != watch.Modified {
+				continue
+			}
+
+			job, fail := event.Object.(*batchv1.Job)
+			if job == nil || fail {
```
tremes, line 132: This always leads to the error below when the type conversion is successful (i.e it's better to call the second var as `ok`). You also don't need the `job == nil`

JoaoFula:
@rluders is there a prediction that this PR might affect the time it takes for the Insights ClusterOperator to degrade?
It seems to consistently fail in that step during one test (keeps timing out)

rluders:
@JoaoFula well... it is possible. It isn't pooling anymore to check the status, instead, it is watching the job status to be completed. But, I don't know how it would/could impact the test and cause some timeout. 🤔

PR #920, opened 2024-03-15
NO-JIRA: delete all active jobs during restart

tremes:
This is adding logic (in techpreview) that after start removes all the active Jobs with `periodic-gathering-` prefix (I think it's not very nice to kill job spawned by user). This is to prevent potential race condition for the future remote configuration (rapid recommendations) based on the OCP version. What can theoretically happen is the following:

- operator creates a job with OCP version 1.2.3 (so the remote configuration is in version 1.2.3 too)
- CVO updates the operator to OCP version 1.2.4 (which leads to re-redeployment), but the job still uses the older OCP version 

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Privacy

## Breaking Changes

No

## References

### DIFF
--- a/pkg/controller/periodic/job.go
+++ b/pkg/controller/periodic/job.go
@@ -131,6 +131,10 @@ func (j *JobController) WaitForJobCompletion(ctx context.Context, job *batchv1.J
 				return fmt.Errorf("watcher channel was closed unexpectedly")
 			}
 
+			if event.Type == apiWatch.Deleted {
+				return nil
+			}
+
 			if event.Type != apiWatch.Modified {
 				continue
 			}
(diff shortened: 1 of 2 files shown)


pkg/controller/periodic/job.go
```diff
@@ -124,6 +124,10 @@ func (j *JobController) WaitForJobCompletion(ctx context.Context, jobName string
 				return fmt.Errorf("watcher channel was closed unexpectedly")
 			}
 
+			if event.Type == watch.Deleted {
+				return nil
+			}
```
tremes, line 129: This is also pretty import, because one can delete the job and in that case it would end up in infinite loop (never ending watch)

ncaak (approved):

ncaak (approved):
/lgtm

PR #936, opened 2024-05-09
OCPBUGS-33082: Add new use cases that requires anonymization

ncaak:

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- ``

## Documentation

- `docs/gathered-data.md`

## Unit Tests

- `pkg/gatherers/clusterconfig/gather_cluster_infrastructure_test.go`

## Privacy

## Changelog

- Yes

## Breaking Changes

- No

## References

https://issues.redhat.com/browse/OCPBUGS-33082

### DIFF
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -124,14 +124,14 @@ func NewAnonymizerFromConfigClient(
 	if err != nil {
 		return nil, err
 	}
-	anonBuilder.AddSensitiveValue(baseDomain, ClusterBaseDomainPlaceholder)
+	anonBuilder.WithSensitiveValue(baseDomain, ClusterBaseDomainPlaceholder)
 
 	APIServerURLs, err := utils.GetClusterAPIServerInfo(ctx, configClient)
 	if err != nil {
 		return nil, err
 	}
 	for _, v := range APIServerURLs {
-		anonBuilder.AddSensitiveValue(v, ClusterAPIServerPlaceholder)
+		anonBuilder.WithSensitiveValue(v, ClusterAPIServerPlaceholder)
 	}
 
 	return anonBuilder.Build()
@@ -343,19 +343,17 @@ func (anonymizer *Anonymizer) AnonymizeMemoryRecord(memoryRecord *record.MemoryR
 		}
 	})
 
-	if len(anonymizer.sensitiveValues) != 0 {
-		for value, placeholder := range anonymizer.sensitiveValues {
-			memoryRecord.Data = bytes.ReplaceAll(
-				memoryRecord.Data,
-				[]byte(value),
-				[]byte(placeholder),
-			)
-			memoryRecord.Name = strings.ReplaceAll(
-				memoryRecord.Name,
-				value,
-				placeholder,
-			)
-		}
+	for value, placeholder := range anonymizer.sensitiveValues {
+		memoryRecord.Data = bytes.ReplaceAll(
+			memoryRecord.Data,
+			[]byte(value),
+			[]byte(placeholder),
+		)
+		memoryRecord.Name = strings.ReplaceAll(
+			memoryRecord.Name,
+			value,
+			placeholder,
+		)
 	}
 
 	memoryRecord.Data = anonymizer.ipNetworkRegex.ReplaceAllFunc(memoryRecord.Data, func(originalIPBytes []byte) []byte {
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -145,7 +145,13 @@ func NewAnonymizerFromConfigClient(
 		return nil, err
 	}
 	secretsClient := kubeClient.CoreV1().Secrets(secretNamespace)
-	a, err := NewAnonymizer(baseDomain, []string{}, secretsClient, configurator, dataPolicy)
+	anonBuilder := &AnonBuilder{}
+	anonBuilder.
+		WithClusterBaseDomain(baseDomain).
+		WithConfigurator(configurator).
+		WithDataPolicy(dataPolicy).
+		WithSecretsClient(secretsClient)
+	a, err := anonBuilder.Build()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


pkg/anonymization/anonymizer.go
```diff
@@ -362,17 +343,19 @@ func (anonymizer *Anonymizer) AnonymizeMemoryRecord(memoryRecord *record.MemoryR
...
-			memoryRecord.Data,
-			[]byte(anonymizer.clusterBaseDomain),
-			[]byte(ClusterBaseDomainPlaceholder),
-		)
-		memoryRecord.Name = strings.ReplaceAll(
-			memoryRecord.Name,
-			anonymizer.clusterBaseDomain,
-			ClusterBaseDomainPlaceholder,
-		)
+	if len(anonymizer.sensitiveValues) != 0 {
```
tremes, line 346: I think this condition is not needed.

pkg/anonymization/builder.go
```diff
@@ -0,0 +1,92 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	k8snet "k8s.io/utils/net"
+)
+
+type AnonBuilder struct {
+	anon     Anonymizer
+	networks []string
+}
+
+func (b *AnonBuilder) AddSensitiveValue(value, placeholder string) *AnonBuilder {
```
tremes, line 21: I think at least this function would deserve some docs/comment.

pkg/anonymization/builder.go
```diff
@@ -0,0 +1,92 @@
...
+	if b.anon.sensitiveValues == nil {
+		b.anon.sensitiveValues = make(map[string]string)
+	}
+	v := strings.TrimSpace(value)
+	if value != "" {
+		b.anon.sensitiveValues[v] = placeholder
+	}
```
tremes, line 28: Perhaps we could do at the beginning of the function:
```
if value == "" {
  return b
}
```

pkg/anonymization/builder_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// When
+			test, err := tc.builder.Build()
+
+			// Assert
+			assert.NoError(t, err)
+			assert.IsType(t, Anonymizer{}, *test)
+			if tc.sensitiveValues != nil {
```
tremes, line 100: Are these conditions required? I guess when you provide `nil` input, you can also expect `nil` output or?

pkg/utils/apiserver.go
```diff
@@ -0,0 +1,31 @@
...
+//
+// It uses this API https://docs.openshift.com/container-platform/4.7/rest_api/config_apis/dns-config-openshift-io-v1.html
+func GetClusterAPIServerInfo(ctx context.Context, configClient configv1client.ConfigV1Interface) ([]string, error) {
+	infra, err := configClient.Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return []string{}, err
+	}
+
+	urls := []string{}
+	for _, url := range []string{infra.Status.APIServerURL, infra.Status.APIServerInternalURL} {
```
tremes, line 24: This will cause panic if the state is not set (probably an edge case).

pkg/anonymization/builder_test.go
```diff
@@ -0,0 +1,133 @@
...
+			if tc.kubeClient != nil {
+				assert.Equal(t, tc.kubeClient, test.gatherKubeClient)
+			}
+			if tc.networkClient != nil {
+				assert.Equal(t, tc.networkClient, test.networkClient)
+			}
+			if len(tc.networks) > 0 {
+				assert.Equal(t, len(tc.networks), len(test.networks))
+			}
+			if tc.runningInCluster {
```
tremes, line 121: I think only this condition is really required :)

pkg/anonymization/builder.go
tremes: What about to call it `WithSensitiveValue` (to keep the "pattern" below) ?

pkg/anonymization/builder_test.go
ncaak: Solved in bc66284efa66957e06efbc3efd2c1f893793085e, thanks for the review

pkg/anonymization/builder_test.go
ncaak: Solved in bc66284efa66957e06efbc3efd2c1f893793085e, thanks for the review

pkg/anonymization/builder.go
ncaak: Solved in bc66284efa66957e06efbc3efd2c1f893793085e, thanks for the review

pkg/anonymization/builder.go
ncaak: Solved in bc66284efa66957e06efbc3efd2c1f893793085e, thanks for the review

pkg/anonymization/anonymizer.go
ncaak: Solved in bc66284efa66957e06efbc3efd2c1f893793085e, thanks for the review

pkg/utils/apiserver.go
ncaak: I'm going to add some unit tests there and refactor the logic. Thanks for the review.

ncaak: Solved in 1e7431666a292f50e34e76fa700279364f700078

pkg/utils/apiserver.go
```diff
@@ -0,0 +1,35 @@
...
+// For example, given the base domain `openshift.example.com`,
+// an API server could be in `api.hypershift.local`.
+//
+// It uses this API https://docs.openshift.com/container-platform/4.7/rest_api/config_apis/dns-config-openshift-io-v1.html
+func GetClusterAPIServerInfo(ctx context.Context, configClient configv1client.ConfigV1Interface) ([]string, error) {
+	infra, err := configClient.Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return []string{}, err
+	}
+	if infra == nil {
```
tremes, line 23: The idea below was to check `infra.Status == nil` to prevent possible panic.

ncaak: I know, but through the unit tests I couldn't set it as nil because of a golang failure. Do you think it's possible for the struct to be initialized as `nil`, even if golang causes an error, as to reach there that way?

tremes: Oh OK so the Status attribute is not a pointer so it can't be nil. Sorry for the noise.

ncaak: No problem. Actually, `infra` object could be nil, so I caught that possibility. Thanks for the revision.

PR #917, opened 2024-03-06
NO-JIRA: (feat) ingress certificates gather

rluders:
This PR implements a new data enhancement to...

## Categories

- [ ] Bugfix
- [X] Data Enhancement
- [ ] Feature
- [ ] Backporting
- [ ] Others (CI, Infrastructure, Documentation)

## Sample Archive

- `path/to/sample_data.json`

## Documentation

- `path/to/documentation.md`

## Unit Tests

- `path/to/file_test.go`

## Privacy

Yes. There are no sensitive data in the newly collected information.

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browse/???
https://access.redhat.com/solutions/???

PR #945, opened 2024-06-03
OCPBUGS-35727: Ingress controller related certificates' validate dates gathering

ncaak:
This PR continues the work of #917

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -65,6 +65,7 @@ linters:
   disable-all: true
   enable:
     - bodyclose
+    - deadcode
     - dogsled
     - dupl
     - errcheck
@@ -89,11 +90,13 @@ linters:
     - noctx
     - rowserrcheck
     - staticcheck
+    - structcheck
     - stylecheck
     - typecheck
     - unconvert
     - unparam
     - unused
+    - varcheck
     - whitespace
 
   # don't enable:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -65,7 +65,6 @@ linters:
   disable-all: true
   enable:
     - bodyclose
-    - deadcode
     - dogsled
     - dupl
     - errcheck
@@ -90,13 +89,11 @@ linters:
     - noctx
     - rowserrcheck
     - staticcheck
-    - structcheck
     - stylecheck
     - typecheck
     - unconvert
     - unparam
     - unused
-    - varcheck
     - whitespace
 
   # don't enable:
--- a/pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
+++ b/pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
@@ -1,0 +1,161 @@
+package clusterconfig
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"time"
+
+	"k8s.io/klog/v2"
+
+	v1 "k8s.io/api/core/v1"
+
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/openshift/insights-operator/pkg/record"
+)
+
+// This map defines the namespace and the certificates that we are looking for
+var ingressCertsMap = map[string][]string{
+	"openshift-ingress-operator": {"router-ca"},
+	"openshift-ingress":          {"router-certs-default"},
+}
+
+type IngressCertificateInfo struct {
+	Name      string    `json:"name"`
+	NotBefore time.Time `json:"not_before"`
+	NotAfter  time.Time `json:"not_after"`
+}
+
+type IngressControllerInfo struct {
+	Name                         string                   `json:"name"`
+	OperatorGeneratedCertificate []IngressCertificateInfo `json:"operator_generated_certificate"`
+	CustomCertificates           []IngressCertificateInfo `json:"custom_certificates"`
+}
+
+func (g *Gatherer) GatherClusterIngressCertificates(ctx context.Context) ([]record.Record, []error) {
+	gatherKubeClient, err := kubernetes.NewForConfig(g.gatherProtoKubeConfig)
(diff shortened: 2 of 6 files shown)


pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
```diff
@@ -0,0 +1,168 @@
...
+var ingressNamespaces = []string{
+	"openshift-ingress-operator",
+	"openshift-ingress",
+}
+
+type CertificateInfo struct {
+	Name        string           `json:"name"`
+	Namespace   string           `json:"namespace"`
+	NotBefore   metav1.Time      `json:"not_before"`
+	NotAfter    metav1.Time      `json:"not_after"`
```
tremes, line 32: I think there's no reason to define those as `metav1.Time`. We can simply have them as `time.Time`.

.golangci.yml
```diff
@@ -90,13 +89,11 @@ linters:
     - noctx
     - rowserrcheck
     - staticcheck
-    - structcheck
     - stylecheck
     - typecheck
     - unconvert
     - unparam
     - unused
-    - varcheck
```
tremes, line 99: do we want these golanci removals here?

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
```diff
@@ -0,0 +1,239 @@
...
+			continue
+		}
+
+		// Step 3: Filter Ingress Controllers with spec.defaultCertificate and get certificate info
+		for i := range controllers.Items {
+			controller := controllers.Items[i]
+
+			// Step 4: Check the certificate limits
+			if len(certificatesInfo) >= ingressCertificatesLimits {
+				klog.V(2).Infof("Reached the limit of ingress certificates (%d), skipping additional certificates", ingressCertificatesLimits)
```
tremes, line 135: this message should be returned as an error so that it's available (visible) in the archive please.

.golangci.yml
ncaak: There were warnings about those linters being deprecated. This change comes from Ricardo, I could rollback it. No strong feelings, more than maybe checking for alternatives.

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
ncaak: Solved in 0c0ab88f4fead85c740764bd8d36b31229b4ddb5

.golangci.yml
ncaak: Solved in fa509166f7920c636eaa5dff66d7494c40cb4155

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
```diff
@@ -0,0 +1,236 @@
...
+	const (
+		routerCertsSecret    string = "router-certs-default"
+		routerCertsNamespace string = "openshift-ingress"
+	)
+	certInfo, err := getCertificateInfoFromSecret(ctx, coreClient, routerCertsNamespace, routerCertsSecret)
+	if err != nil {
+		return nil, err
+	}
+
+	return certInfo, nil
```
tremes, line 231: nit: you can just return `getCertificateInfoFromSecret(ctx, coreClient, routerCertsNamespace, routerCertsSecret)` and similar for the function above

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func infereKey(info *CertificateInfo) CertificateInfoKey {
```
tremes, line 234: nit: what about to call it something like `newCertificateInfoKey`?

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	var ci []*CertificateInfo
+	if len(certificatesInfo) > 0 {
+		// Step 6: Generate the certificates record
+		ci = make([]*CertificateInfo, len(certificatesInfo))
+		i := 0
+		for _, v := range certificatesInfo {
+			ci[i] = v
+			i++
```
tremes, line 168: probably better to call just `ci = append(ci, v)`

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
ncaak: oh, true, I was refactoring this function and just missed that. Thanks.

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
ncaak: Solved in 82311f791c8e70eac909fe064f3a3eef081ea172

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
ncaak: Solved in 82311f791c8e70eac909fe064f3a3eef081ea172

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
```diff
@@ -0,0 +1,225 @@
...
+						ControllerInfo{Name: controller.Name, Namespace: controller.Namespace},
+					)
+					certificatesInfo[newCertificateInfoKey(certInfo)] = certInfo
+				}
+			}
+		}
+	}
+
+	var ci []*CertificateInfo
+	if len(certificatesInfo) > 0 {
```
tremes, line 164: This is also not necessary, but let's not run the CI again :)

tremes:
Looks mostly good and seems to work OK :) 
/lgtm

pkg/gatherers/clusterconfig/gather_cluster_ingress_certificates.go
ncaak: Aw yeah... After the refactor of the inner code the conditional is of no use :(

I'll leave it marked in my local repo for the next PR to include it as an improvement. Thanks.

PR #949, opened 2024-06-18
Gather Workload Runtime Info From Containers

jmesnil:
Implement RFE  https://github.com/openshift/enhancements/pull/1584

This PR implements a new data enhancement to the insights operator workloads.

It gathers runtime data from running containers such as Operating System and runtime stack (Golang, Node.js, Java, etc.)

The actual component extracting the data is at https://github.com/jmesnil/insights-operator-runtime.
It's bundled in a container image that is used by the insights operator to augment the workloads data.

## Categories

- [X] Data Enhancement

## Sample Archive

- TBD docs/insights-archive-sample/config/workload_info.json

## Documentation

- TBD

## Unit Tests

- TBD

## Privacy

The additional data are truncated and hashed similarly to the existing workload_info data

## Changelog

TBD

## Breaking Changes

No

## References

TBD


Dockerfile-insights-operator-runtime
jmesnil, line 1: Temporary file to deploy the insights operator with the integrated enhancement.
This file will be removed before final integration

deploy-insights-operator-with-runtime.sh
jmesnil, line 1: Temporary file to deploy the insights operator with the integrated enhancement.
This file will be removed before final integration

go.mod
jmesnil, line 1: Additional Go modules so that the insights-operator can invoke a `kubectl exec` operations on the insights-operator-runtime pods

manifests/010-clusterrole-insights-operator-runtime.yaml
jmesnil, line 1: The insights operator runtime needs its own roles and SCC to be deployed with the required privileges

manifests/010-insights-operator-runtime.yaml
jmesnil, line 1: This file would be used if the Insights Operator deploys the runtime container image as installation time.
In this draft, it is deployed whenever the gather operation is performed but @tremes is arguing against that (as it will bring complexity with on-demand gather jobs)

pkg/gatherers/workloads/gather_workloads_info.go
```diff
@@ -79,20 +87,35 @@ func (g *Gatherer) GatherWorkloadInfo(ctx context.Context) ([]record.Record, []e
 		return nil, []error{err}
 	}
 
-	return gatherWorkloadInfo(ctx, gatherKubeClient.CoreV1(), gatherOpenShiftClient)
+	return gatherWorkloadInfo(ctx, gatherKubeClient, gatherOpenShiftClient, imageConfig)
```
jmesnil, line 90: I shifted a bit the parameters to be able to use the client to connect to the `insights-operator-runtime` pod.

pkg/gatherers/workloads/gather_workloads_info.go
```diff
@@ -496,11 +525,26 @@ func calculateWorkloadContainerShapes(
 			firstArg = shortHash
 		}
 
+		var runtimeInfo workloadRuntimeInfoContainer
```
jmesnil, line 528: when workload data is gathered, we go through the `namespace / pod / containerID / workloadRuntimeInfos` map and populate the shapes with them

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
```diff
@@ -0,0 +1,328 @@
...
+	h hash.Hash,
+	coreClient corev1client.CoreV1Interface,
+	appClient appsv1client.AppsV1Interface,
+	restConfig *rest.Config,
+) (workloadRuntimes, error) {
+	start := time.Now()
+
+	workloadRuntimeInfos := make(workloadRuntimes)
+
+	klog.Infof("Deploying insights-operator-runtime...\n")
```
jmesnil, line 50: as mentioned in the comment above, this could be changed so that the `insights-operator-runtime` daemonset is deployed when the operator is installed instead of doing it dynamically

Dockerfile-insights-operator-runtime
```diff
@@ -0,0 +1,15 @@
...
+RUN microdnf update -y && microdnf -y install \
+    tar
+
+COPY ./bin/insights-operator /usr/bin/
+COPY config/pod.yaml /etc/insights-operator/server.yaml
+COPY manifests /manifests
+# Do not start this instance of the insights-operator 
+# as it would conflict with the one already installed by OLM.
+# We just sleep so that the gather operation can be executed on that instance
+ENTRYPOINT ["sleep", "360000"]
```
tremes, line 15: This file is just temporary right? In sense there is no plan to merge it in the repository. Do I understand correctly? Same for the script below.

jmesnil: yes, that's correct.

pkg/gatherers/workloads/gather_workloads_info.go
```diff
@@ -41,6 +41,13 @@ const (
 	podsLimit = 8000
 )
 
+// keys are namespace / pod / containerID and the final value is the workloadRuntimeInfoContainer for this container
+type workloadRuntimes map[string]map[string]map[string]workloadRuntimeInfoContainer
```
tremes, line 45: perhaps we could build a new type encompassing the strings for `namespace`, `pod` and `containerID` and then use it as the map key?

pkg/gatherers/workloads/gather_workloads_info.go
```diff
@@ -108,7 +127,11 @@ func gatherWorkloadInfo(
 	handleWorkloadImageInfo(ctx, &info, start, imagesDoneCh)
 
 	if limitReached {
-		return records, []error{fmt.Errorf("the %d limit for number of pods gathered was reached", podsLimit)}
+		errors = append(errors, fmt.Errorf("the %d limit for number of pods gathered was reached", podsLimit))
+	}
+
+	if len(errors) > 0 {
```
tremes, line 133: This checks looks unnecessary.

pkg/gatherers/workloads/gather_workloads_info.go
```diff
@@ -228,12 +251,12 @@ func podCanBeIgnored(pod *corev1.Pod) bool {
...
 	var podShape workloadPodShape
 	var ok bool
-	podShape.InitContainers, ok = calculateWorkloadContainerShapes(h, pod.Spec.InitContainers, pod.Status.InitContainerStatuses)
+	podShape.InitContainers, ok = calculateWorkloadContainerShapes(h, pod.ObjectMeta, pod.Spec.InitContainers, pod.Status.InitContainerStatuses)
 	if !ok {
 		return workloadPodShape{}, false
 	}
 
-	podShape.Containers, ok = calculateWorkloadContainerShapes(h, pod.Spec.Containers, pod.Status.ContainerStatuses)
+	podShape.Containers, ok = calculateWorkloadContainerShapes(h, pod.ObjectMeta, pod.Spec.Containers, pod.Status.ContainerStatuses)
```
tremes, line 259: nit: Perhaps we could pass the complete Pod reference then...

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return workloadRuntimeInfos, nil
+}
+
+// List the pods of the insights-operator-runtime component
+// and returns a map where the key is the name of the worker nodes
+// and the value the name of the  insights-operator-runtime's pod running on that worker node.
+func getInsightsOperatorRuntimePods(
+	coreClient corev1client.CoreV1Interface,
+	ctx context.Context,
```
tremes, line 75: nit: context is usually the first parameter.

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
```diff
@@ -0,0 +1,215 @@
...
+
+	err = exec.StreamWithContext(ctx, remotecommand.StreamOptions{
+		Stdout: &execOut,
+		Stderr: &execErr,
+		Tty:    false,
+	})
+	if err != nil {
+		fmt.Printf("got insights operator runime error: %s\n", err)
+		fmt.Printf("command error output: %s\n", execErr.String())
+		fmt.Printf("command output: %s\n", execOut.String())
```
tremes, line 204: I guess this all is TBD, but the errors should be logged properly or returned by the function (and handled later on)

jmesnil:
@tremes I cherry picked your commits and updated the integration to use a HTTP server to gather the runtime infos.
The code is not complete yet (error handling is missing) and it's missing the `kube-rbac-proxy` integration to restrict access to the `insights-operator` role.

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
```diff
@@ -0,0 +1,185 @@
...
+// Get all WorkloadRuntimeInfos for a single Node (using the insights-runtime-extractor pod running on this node)
+// FIXME return an (workloadRuntimes, error)
+func getNodeWorkloadRuntimeInfos(
+	ctx context.Context,
+	h hash.Hash,
+	runtimePodIP string,
+) workloadRuntimes {
+
+	extractorURL := fmt.Sprintf("http://%s:8000/gather_runtime_info", runtimePodIP)
+	request, err := http.NewRequestWithContext(ctx, "GET", extractorURL, nil)
```
tremes, line 161: I think it would be good to narrow down the context here used for the HTTP request. I mean something like:
```
httpCtx := context.WithTimeout(ctx, 2*time.Minute)
```
because the input `ctx` parameter has timeout in hours.

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
```diff
@@ -0,0 +1,185 @@
...
+		List(ctx, metav1.ListOptions{
+			LabelSelector: "app.kubernetes.io/name=insights-runtime-extractor",
+		})
+	if err != nil {
+		return nil, err
+	}
+
+	var runtimePods []podWithNodeName
+	for _, pod := range pods.Items {
+		runtimePods = append(runtimePods, podWithNodeName{
```
tremes, line 86: Would it make sense to check the status of the Pod here? For example, if it is running, because we may end up with HTTP requests to `pending` Pod for example, so it might be a good idea to filter them out here. WDYT?

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
jmesnil: fixed, thanks

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
jmesnil: fixed, thanks

pkg/gatherers/workloads/gather_workloads_info.go
```diff
@@ -225,15 +241,15 @@ func podCanBeIgnored(pod *corev1.Pod) bool {
...
 	var podShape workloadPodShape
 	var ok bool
-	podShape.InitContainers, ok = calculateWorkloadContainerShapes(h, pod.Spec.InitContainers, pod.Status.InitContainerStatuses)
+	podShape.InitContainers, ok = calculateWorkloadContainerShapes(h, pod.ObjectMeta, pod.Spec.InitContainers, pod.Status.InitContainerStatuses, workloadInfo)
 	if !ok {
 		return workloadPodShape{}, false
 	}
 
-	podShape.Containers, ok = calculateWorkloadContainerShapes(h, pod.Spec.Containers, pod.Status.ContainerStatuses)
+	podShape.Containers, ok = calculateWorkloadContainerShapes(h, pod.ObjectMeta, pod.Spec.Containers, pod.Status.ContainerStatuses, workloadInfo)
```
tremes, line 252: Maybe it's time to pass the complete `Pod` reference here? Just suggestion.....

jmesnil: that'll be cleaner. I'll do it

jmesnil: @tremes Do you mean passing the pod instead of the `pod.ObjectMeta` (the method is called twice for the Containers and the imiticontainers)

tremes: ohh right it needs to distinguish the containers and initContainers....humm so perhaps easier to keep it as it is :thinking:

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
```diff
@@ -0,0 +1,173 @@
...
+	nodeWorkloadCh := make(chan workloadRuntimesResult)
+	var receiveWg sync.WaitGroup
+	receiveWg.Add(1)
+
+	go func() {
+		defer receiveWg.Done()
+		for infosRes := range nodeWorkloadCh {
+			if infosRes.Error != nil {
+				errors = append(errors, infosRes.Error)
+				return
```
tremes, line 48: I guess `continue` instead of `return` ?

jmesnil: 🤕 thanks!

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+	resp, err := http.DefaultClient.Do(request)
+	if err != nil {
+		return workloadRuntimesResult{
+			Error: err,
+		}
+	}
+	if resp.StatusCode != 200 {
+		return workloadRuntimesResult{}
```
tremes, line 139: I guess we still want some info (error) about this situation so probably something like:
```
	return workloadRuntimesResult{
			Error: insightsclient.HttpError{
				StatusCode: resp.StatusCode,
				Err:        fmt.Errorf(resp.Status),
			},
		}
```

pkg/gatherers/workloads/gather_workloads_runtime_infos.go
```diff
@@ -0,0 +1,173 @@
...
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return workloadRuntimesResult{
+			Error: err,
+		}
+	}
+	var nodeOutput nodeRuntimeInfo
+	json.Unmarshal(body, &nodeOutput)
```
tremes, line 150: The returned error should not be ignored here. I am going to write some unit test for this function.

jmesnil:
@tremes I updated the PR to use the `kube-rbac-proxy` in commit https://github.com/openshift/insights-operator/pull/949/commits/e000acd9eca3f621097fb5978783eff8079fdb8a.
This ensures that only the `operator` service account can access the insights-runtime-extractor API from their pods.

manifests/010-insights-runtime-extractor.yaml
```diff
@@ -0,0 +1,89 @@
...
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: insights-runtime-extractor
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: insights-runtime-extractor
+    spec:
+      serviceAccountName: insights-runtime-extractor-sa
```
tremes, line 17: We probably need to add following `securityContext` definition here:
```
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop:
          - ALL
        readOnlyRootFilesystem: true
        runAsGroup: 65532
        runAsNonRoot: true
        runAsUser: 65532
        seccompProfile:
          type: RuntimeDefault
```

manifests/06a-deployment-with-runtime.yaml
```diff
@@ -0,0 +1,96 @@
...
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: RELEASE_VERSION
+          value: "0.0.1-snapshot"
```
tremes, line 96: The container is missing arguments definition in my opinion:
```
        args:
        - start
        - --config=/etc/insights-operator/server.yaml
```

jmesnil: this manifest will go away before the PR is merged. This "alternate" deployment is started without the argument (it just `sleeps`) so that I can then perform a `gather` operation on it with `kubectl exec`

tremes: ahh yeah I see. Let's close this thread then.

manifests/03-clusterrole.yaml
```diff
@@ -353,6 +353,12 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
```
jmesnil, line 356: This can be removed as the communication is done with HTTP and no longer with a `pod/exec`.

joepvd:
Confirming that the new entries in image-references are present in the imageStream for 4.19 and 4.18, and that this is good to go from that (limited) perspective.

manifests/02-namespace-privileged.yaml
```diff
@@ -0,0 +1,21 @@
...
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    workload.openshift.io/allowed: "management"
+    capability.openshift.io/name: Insights
+  labels:
+    openshift.io/cluster-monitoring: "true"
+    openshift.io/run-level: "" # specify no run-level turns it off on install and upgrades
+    name: openshift-insights
+    # Requires privileged for the insights-runtime-extractor container
+    release.openshift.io/feature-set: TechPreviewNoUpgrade
```
tremes, line 18: This must be annotation and not label

xueqzhan:
/payload-job periodic-ci-openshift-release-master-ci-4.18-e2e-aws-ovn-techpreview

PR #953, opened 2024-06-24
NO-JIRA: introduce JSON schema & validation for rapid container logs

tremes:

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive

- no new data

## Documentation

- 

## Unit Tests

- 

## Privacy

## Breaking Changes

Yes/No

## References

### DIFF
--- a/pkg/gatherers/conditional/container_log.schema.json
+++ b/pkg/gatherers/conditional/container_log.schema.json
@@ -1,0 +1,45 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "container_log.schema.json",
+    "title": "Container Log Request",
+    "descriptions": "Container Log Request used in Rapid Recommendations",
+    "examples": [
+        {
+            "namespace": "openshift-etcd",
+            "pod_name_regex": ".*",
+            "messages": [
+                "wal:\\ max\\ entry\\ size\\ limit\\ exceeded"
+            ]
+        }
+    ],
+    "type": "object",
+    "required": [
+        "namespace",
+        "pod_name_regex",
+        "messages"
+    ],
+    "properties": {
+        "namespace": {
+            "type": "string",
+            "description": "OpenShift namespace name",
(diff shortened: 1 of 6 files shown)


pkg/gatherers/conditional/container_log.schema.json
```diff
@@ -0,0 +1,45 @@
...
+            "pattern": "^openshift-[a-zA-Z0-9_.-]{1,128}$|^kube-[a-zA-Z0-9_.-]{1,128}$"
+        },
+        "pod_name_regex": {
+            "type": "string",
+            "description": "Regular expression used to find matching Pods"
+        },
+        "messages": {
+            "type": "array",
+            "description": "The list of regular expressions used to filter the container logs",
+            "uniqueItems": true,
```
tremes, line 34: @jholecek-rh can/should we do this? i.e require regular expressions to be unique in one particular "container log request".

jholecek-rh: What would we seek to achieve by requiring unique regexes?

1. Smaller remote config? I think there are much more significant factors that affect the size. 
2. Faster regex execution/ parsing? The Insights Operator can deduplicate the regexes before combining them into the single one used for parsing, but I think the performance gains would be negligible if any at all (I don't expect duplicates, deduplication would take a little time).

If I compare this with the potential complexity of making sure that there are no duplicates (along with timely feedback to data requestors), I wouldn't bother and allow duplicates.

tremes: Mainly the point number 1. The deduplication is in place. OK so let's remove this validation requirement. Thanks

ncaak (approved):
/lgtm

tremes:
This is not such complex change and techpreview job is failing probably because cloud-credential is marked as degraded for some reason. 
/override ci/prow/e2e-gcp-ovn-techpreview

PR #979, opened 2024-08-13
OCPBUGS-38727: Enable Insight Operator entitlements for multi arch clusters

joselsegura:
This PR implements a new mechanism to gather a set of architectures used by nodes in the current cluster.

This set of architectures is not used yet, but if Insights Operator is not able to retrieve it, it will make the entitlements retrieval process to fail

## Categories
- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive

The archive won't change with this feature

## Documentation
No documentation update

## Unit Tests
No tests were updated

## Privacy

## Changelog
No

## Breaking Changes
No

## References
https://issues.redhat.com/browse/OCPBUGS-38727

### DIFF
--- a/pkg/ocm/sca/architectures_gather.go
+++ b/pkg/ocm/sca/architectures_gather.go
@@ -1,0 +1,23 @@
+package sca
+
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// gatherArchitectures connects to K8S API to retrieve the list of
+// nodes and create a set of the present architectures
+func (c *Controller) gatherArchitectures(ctx context.Context) (map[string]struct{}, error) {
+	nodes, err := c.coreClient.Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	architectures := make(map[string]struct{})
+	for i := range nodes.Items {
+		nodeArch := nodes.Items[i].Status.NodeInfo.Architecture
+		architectures[nodeArch] = struct{}{}
+	}
+	return architectures, nil
+}
--- a/pkg/ocm/sca/sca.go
+++ b/pkg/ocm/sca/sca.go
@@ -98,7 +98,11 @@ func (c *Controller) requestDataAndCheckSecret(ctx context.Context, endpoint str
 	klog.Infof("Pulling SCA certificates from %s. Next check is in %s", c.configurator.Config().SCA.Endpoint,
 		c.configurator.Config().SCA.Interval)
 
-	data, err := c.requestSCAWithExpBackoff(ctx, endpoint)
+	architectures, err := c.gatherArchitectures(ctx)
+	if err != nil {
+		klog.Warningf("Gathering nodes architectures failed: %s", err.Error())
+	}
+	responses, err := c.requestSCAWithExpBackoff(ctx, endpoint, architectures)
 	if err != nil {
 		httpErr, ok := err.(insightsclient.HttpError)
 		errMsg := fmt.Sprintf("Failed to pull SCA certs from %s: %v", endpoint, err)
@@ -125,17 +129,19 @@ func (c *Controller) requestDataAndCheckSecret(ctx context.Context, endpoint str
 		return
 	}
 
-	var ocmRes Response
-	err = json.Unmarshal(data, &ocmRes)
-	if err != nil {
-		klog.Errorf("Unable to decode response: %v", err)
-		return
-	}
-	// check & update the secret here
-	err = c.checkSecret(ctx, &ocmRes)
-	if err != nil {
-		klog.Errorf("Error when checking the %s secret: %v", secretName, err)
-		return
+	for idx := range responses {
+		var ocmRes Response
+		err = json.Unmarshal(responses[idx], &ocmRes)
+		if err != nil {
+			klog.Errorf("Unable to decode response: %v", err)
+			return
+		}
+		// check & update the secret here
+		err = c.checkSecret(ctx, &ocmRes)
+		if err != nil {
+			klog.Errorf("Error when checking the %s secret: %v", secretName, err)
+			return
+		}
 	}
 
 	klog.Infof("%s secret successfully updated", secretName)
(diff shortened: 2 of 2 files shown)


pkg/ocm/sca/sca_test.go
```diff
@@ -19,7 +19,7 @@ var (
 func Test_SCAController_SecretIsCreated(t *testing.T) {
 	kube := kubefake.NewSimpleClientset()
 	coreClient := kube.CoreV1()
-	scaController := New(context.TODO(), coreClient, nil, nil)
+	scaController := New(context.TODO(), coreClient, nil, nil, nil)
```
tremes, line 22: nit (not related to this change): we could change `context.TODO` to `context.Background`

pkg/ocm/sca/sca.go
```diff
@@ -219,6 +222,11 @@ func (c *Controller) requestSCAWithExpBackoff(endpoint string) ([]byte, error) {
 	var data []byte
 	err := wait.ExponentialBackoff(bo, func() (bool, error) {
 		var err error
+		ctx := context.TODO()
+		_, err = gatherArchitectures(ctx, c.gatherProtoKubeConfig)
+		if err != nil {
+			return true, err
+		}
```
tremes, line 229: Note that the `ExponentialBackoff` function can be called multiple times and I think we don't need to gather the architectures repeatedly in this case. So I would suggest to move this logic out of `requestSCAWithExpBackoff` and it should also help with the `TODO` context usage.

pkg/controller/operator.go
```diff
@@ -268,7 +268,7 @@ func (s *Operator) Run(ctx context.Context, controller *controllercmd.Controller
 		return fmt.Errorf("unable to set initial cluster status: %v", err)
 	}
 
-	scaController := sca.New(ctx, kubeClient.CoreV1(), configAggregator, insightsClient)
+	scaController := sca.New(ctx, kubeClient.CoreV1(), configAggregator, insightsClient, gatherProtoKubeConfig)
```
tremes, line 271: Is this required? :thinking: We are already passing `kubeClient.CoreV1()` so I guess we could use it for listing Nodes.

pkg/ocm/sca/sca_test.go
joselsegura: Reverted, as the config is not needed anymore

pkg/ocm/sca/sca.go
joselsegura: You are right. Moved out of the body of the inner function

pkg/controller/operator.go
joselsegura: I didn't realise! Reverted this change and refactored the gathered function to use that client instead of creating a new one

tremes:
Thank you!
/approve
/lgtm

pkg/ocm/sca/sca.go
```diff
@@ -221,30 +223,34 @@ func (c *Controller) requestSCAWithExpBackoff(ctx context.Context, endpoint stri
...
-				return true, err
-			}
-			httpErr := err.(insightsclient.HttpError)
-			// try again only in case of 500 or higher
-			if httpErr.StatusCode >= http.StatusInternalServerError {
-				// check the number of steps to prevent "timeout waiting for condition" error - we want to propagate the HTTP error
-				if bo.Steps > 1 {
-					klog.Errorf("%v. Trying again in %s", httpErr, bo.Step())
-					return false, nil
+		for arch := range architectures {
```
tremes, line 229: I think we don't need to do this. I think we can append the architecture to one string - e.g "x86_64,aarch64"

.githooks/pre-commit
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
tremes, line 1: This change has been already merged or not? :thinking:

pkg/ocm/sca/sca.go
joselsegura: Then, we will need to split the string and iterate in the `insightsclient.Client.RecvSCACerts` method, as the used API only allow to perform 1 request per architecture

pkg/insights/insightsclient/insightsclient.go
```diff
@@ -33,7 +33,7 @@ import (
 const (
 	responseBodyLogLen = 1024
 	insightsReqId      = "x-rh-insights-request-id"
-	scaArchPayload     = `{"type": "sca","arch": "x86_64"}`
```
prb112, line 36: Hey @Prashanth684 @aleskandro since we have multi-arch compute.... I'm assuming this would not work? Thoughts? Thanks, Paul

prb112: just read the rest of the code... makes sense

pkg/ocm/sca/architectures_gather.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// gatherArchitectures connects to K8S API to retrieve the list of
+// nodes and create a set of the present architectures
+func (c *Controller) gatherArchitectures(ctx context.Context) (map[string]struct{}, error) {
```
prb112, line 11: it might be best to request all arches. MachineSets could be on an autoscaler when the gather action is executed.

joselsegura: I don't get your point. This method is requesting all archs from the nodes list returned by k8s API. Could you please elaborate more?

prb112: Right - I see that. Since you can scale machinesets from zero, you may be temporarily excluding some arches from your retrieval.

prb112: Example https://issues.redhat.com/browse/OCPBUGS-27509 and these would be the second set of MachineSets created.

pkg/ocm/sca/sca.go
joselsegura: I have checked the `certificates` API and it doesn't support this kind of request where the architecture is a string with multiple values separated by commas.

I tried with an array of strings too, and it doesn't work neither (different API errors), so I guess the only way to implement this is retrieving one certificate at a time

pkg/ocm/sca/architectures_gather.go
joselsegura: I get your point now. If at the moment of retrieving the nodes list some machineset is scaled to 0, how can I get the node architecture? I think we can only gather the arch for the current existing nodes, not from those that are unavailable because any reason, but maybe there is another way to retrieve the nodes list...

prb112:
Hey @joselsegura will this be merged at some point?

opokornyy:
> Hey @joselsegura will this be merged at some point?

Hey, just a quick heads-up: we are now waiting for the implementation of a new endpoint that will return a list of certificates for specified architectures, so we won't need to send multiple requests. The MR is currently under review, and once it is merged, we will resume work on this

prb112:
Thank you for the update @opokornyy

opokornyy:
> Thank you for the update @opokornyy

Hi, we will close this MR and the work will continue in this one - https://github.com/openshift/insights-operator/pull/1066

PR #1009, opened 2024-10-09
OCPBUGS-42961: update the filtering of the container logs

tremes:

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

 no new data

## Unit Tests

- `path/to/file_test.go`

## Privacy

## Breaking Changes

Yes/No

## References

https://issues.redhat.com/browse/OCPBUGS-42961

### DIFF
--- a/pkg/gatherers/conditional/filter_and_gather_container_logs.go
+++ b/pkg/gatherers/conditional/filter_and_gather_container_logs.go
@@ -97,8 +97,9 @@ func gatherContainerLogs(
 	return records, errs
 }
 
+// mapKeysToSlice converts the set of keys of provided map to a slice of strings
 func mapKeysToSlice(m map[PodNameRegexPrevious]sets.Set[string]) []string {
-	var keys []string
+	keys := make([]string, 0, len(m))
 	for k := range m {
 		keys = append(keys, k.PodNameRegex)
 	}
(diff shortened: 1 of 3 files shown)


pkg/gatherers/conditional/filter_and_gather_container_logs.go
```diff
@@ -84,75 +85,72 @@ func gatherContainerLogs(
...
+		go filterContainerLogs(shorterCtx, coreClient, logRequest, &sendWG, recCh)
 	}
 	sendWG.Wait()
 	close(recCh)
 	receiveWG.Wait()
 	return records, errs
 }
 
+func mapKeysToSlice(m map[PodNameRegexPrevious]sets.Set[string]) []string {
+	var keys []string
```
ncaak, line 101: knowing the map length wouldn't be better (efficient) to use `make([]string, len(m))` and then assign values instead of appending?

ncaak (approved):

pkg/gatherers/conditional/filter_and_gather_container_logs.go
tremes: :+1: yes. Thank you.

ncaak (approved):
/lgtm

tremes:
One minor test failure (it looks like the order of the error message is different). We need to check, but let's unblock it for now.
/override ci/prow/insights-operator-e2e-tests

PR #1018, opened 2024-10-23
NO-JIRA: Add gather server error known codes to the checks

ikerreyes:
Extend test cases for conditional gathering to include server known error codes.

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Privacy

### DIFF
--- a/pkg/gatherers/conditional/conditional_gatherer_test.go
+++ b/pkg/gatherers/conditional/conditional_gatherer_test.go
@@ -1,6 +1,7 @@
 package conditional
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


pkg/gatherers/conditional/conditional_gatherer_test.go
```diff
@@ -280,15 +281,15 @@ func TestGetGatheringFunctions(t *testing.T) {
 	tests := []struct {
 		name                 string
 		endpoint             string
-		remoteConfig         string
+		remoteConfig         *MockGatheringRulesServiceClient
```
tremes, line 284: nit: I would probably rename also the variable name to something like `mockClient` or `remoteConfigClient`

tremes:
Thanks for the PR @ikerreyes !
/approve

pkg/gatherers/conditional/conditional_gatherer_test.go
ikerreyes: OK, for me, having the `remote` there is helpful because then I easily relate it to the remote server, so I have used `remoteMockClient`. Let me know if you have any other preference.

tremes:
with your permission @ikerreyes :) 
/label qe-approved

PR #1070, opened 2025-01-29
CCXDEV-9905: Introduce a new config option for storing archives to persistent volume

opokornyy:
This PR adds a new config option for user to specify PVC name where archives produced by on-demand gathering jobs will be stored. It also introduces a cleanup job that removes old archives from the persistent volume.

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive
None

## Documentation

None

## Unit Tests

- `pkg/controller/periodic/job_test.go`

## Privacy

## Breaking Changes

No

## References

https://issues.redhat.com/browse/CCXDEV-9905


pkg/cmd/start/start.go
```diff
@@ -104,95 +104,10 @@ func NewGatherAndUpload() *cobra.Command {
 		Short: "Runs the data gathering as job, uploads the data, waits for Insights analysis report and ends",
 		Run:   runGatherAndUpload(operator, cfg),
 	}
+	cmd.PersistentFlags().String("storagePath", "", "Path to store the gathered data")
```
tremes, line 107: Why do we need this flag please? My thinking was that we can read this from API and if it's empty then use the default value or am I missing something?

pkg/controller/gather_commands.go
```diff
@@ -399,7 +420,7 @@ func wasDataProcessed(ctx context.Context,
 // If not, non-nill error is returned.
 func (g *GatherJob) storagePathExists() error {
 	if _, err := os.Stat(g.StoragePath); err != nil && os.IsNotExist(err) {
-		if err = os.MkdirAll(g.StoragePath, 0777); err != nil {
+		if err = os.MkdirAll(g.StoragePath, 0o777); err != nil {
```
tremes, line 423: why is this required please? :thinking:

docs/on-demand-gather.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: DataGather
+metadata:
+  name: on-demand-gather-job
+spec:
+  storage:
+    type: PersistentVolume
+    mountPath: /data
+    persistentVolume:
+      claim:
+        name: on-demand-gather-pvc
```
tremes, line 25: I think we can put this to the official OCP documentation but I would probably not keep it here. Just my 2 cents...

pkg/controller/gather_commands.go
opokornyy: It is the newer Go syntax to write the octal numbers that should be more readable and mine IDE updated this line. It is not required and we could stick with the old format if you prefer that.

pkg/cmd/start/start.go
opokornyy: I was using this flag to propagate the value from the API and IO to the container where we are executing the command. In the `runGatherAndUpload` function I am then just checking if the flag was provided, if so then I will use it, otherwise it uses default path. I think the main reason was that the command is used for both periodic and on-demand gathering jobs and if the path was changed in the `DataGather` CRD then I needed to use it in a job command. But there might be an easier way to achieve this that I haven't found...

opokornyy: I have reworked this to not use the flag. I did not notice that we are passing the `DataGather` name to the job as a environment variable and then getting the CRD in a job... Thanks for pointing this out!

docs/on-demand-gather.yaml
opokornyy: I think it's nice to have it in a repository so anyone can use it to create the on-demand gathering job without going to the OCP docs. Otherwise, feel free to drop, but let's see what @ncaak thinks about it being here

pkg/controller/periodic/job.go
```diff
@@ -160,3 +160,57 @@ func (j *JobController) WaitForJobCompletion(ctx context.Context, job *batchv1.J
...
+				},
+			}
+		}
+	}
+
+	klog.Info("Creating volume source with EmptyDir")
+	return corev1.VolumeSource{
+		EmptyDir: &corev1.EmptyDirVolumeSource{},
+	}
+}
```
tremes, line 192: I think this function would deserve some comment/documentation - especially the fact that in case of an error the `EmptyDirVolumeSource` is used. I was thinking if failing the job would be better (in case of an error with PVC) but I don't have strong opinion.

pkg/controller/periodic/job.go
```diff
@@ -29,7 +33,12 @@ func NewJobController(kubeClient kubernetes.Interface) *JobController {
 
 // CreateGathererJob creates a new Kubernetes Job with provided image, volume mount path used for storing data archives and name
 // derived from the provided data gather name
-func (j *JobController) CreateGathererJob(ctx context.Context, dataGatherName, image, archiveVolumeMountPath string) (*batchv1.Job, error) {
+func (j *JobController) CreateGathererJob(
+	ctx context.Context, dataGatherName, image string, dataReporting *config.DataReporting, storage *insightsv1alpha1.Storage,
```
tremes, line 37: nit: we could save one argument here by passing the pointer to the full `insightsv1alpha1.DataGather`

pkg/recorder/diskrecorder/diskrecorder.go
```diff
@@ -101,40 +101,93 @@ func (d *DiskRecorder) SaveAtPath(records record.MemoryRecords, path string) (re
 	return completed, nil
 }
 
+// Prune the archives when there are more than count archives
+func (d *DiskRecorder) PruneByCount(countTreshold int) error {
```
tremes, line 105: It would be nice to have a test for this one :)

opokornyy: Done

pkg/controller/periodic/job.go
opokornyy: Added a comment for the function, hopefully it is sufficient. 

I think logging the error message when failing to get the PVC and then using an ephemeral volume instead should be enough, but I am open to changing it if needed

tremes:
I didn't try this, but the changes look good. 
/lgtm

docs/on-demand-gather.yaml
ncaak: No strong opinions here. I think we could put this in the OCP documentation, for sure. But I don't see why it shouldn't be here as an example. Maybe with a more explanatory comment that gives some insight into what it is used for?
@tremes what's your concern to not have this example here?

tremes: Not really a concern, but my take is that the PVC definition can be easily found in Kubernetes docs and in general it would be nice to have some context for those yaml files. For example something like if you want to do A then you need to configure B,C etc. 
Perhaps `docs/arch.md` could be extended.

opokornyy: I will try to update `docs/arch.md`

opokornyy: `docs/arch.md` is updated now, so you can review it

tremes:
Only docs update
/lgtm

BaiyangZhou:
/override pull-ci-openshift-insights-operator-master-e2e-gcp-ovn-techpreview

PR #1075, opened 2025-02-26
feat: add gatherer name validation to gendocs

opokornyy:
This commit adds validation for gatherer names to validate that they follow the naming convention.

This PR adds validation for gatherer names to validate that they follow the naming convention.

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

None

## Documentation

None

## Unit Tests

None

## Privacy

## Breaking Changes

No

## References

This is the PR that introduced the naming convention to openshift API for our CRDs (DataGather, InsightsDataGather)
https://github.com/openshift/api/pull/2200

### DIFF
--- a/cmd/gendoc/main.go
+++ b/cmd/gendoc/main.go
@@ -12,20 +12,26 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 	"time"
 )
 
+const maxGathererNameLength = 256
+
 var (
-	inPath          string
-	outPath         string
-	mdf             *os.File
-	randSource      = rand.NewSource(time.Now().UnixNano())
-	reGather        = regexp.MustCompile(`^((Build)?Gather)(.*)`)
-	reExample       = regexp.MustCompile(`^(Example)(.*)`)
-	reSampleArchive = regexp.MustCompile(`docs/(insights-archive-sample/.*)`)
-	cleanRoot       = "./"
+	inPath                   string
+	outPath                  string
+	mdf                      *os.File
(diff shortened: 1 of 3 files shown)


cmd/gendoc/main.go
```diff
@@ -12,20 +12,26 @@ import (
...
+	inPath                   string
+	outPath                  string
+	mdf                      *os.File
+	randSource               = rand.NewSource(time.Now().UnixNano())
+	reGather                 = regexp.MustCompile(`^((Build)?Gather)(.*)`)
+	reExample                = regexp.MustCompile(`^(Example)(.*)`)
+	reSampleArchive          = regexp.MustCompile(`docs/(insights-archive-sample/.*)`)
+	reGathererName           = regexp.MustCompile("(?m)^###\\s*Config ID\\s*\\n`([^`]*)`")
+	reGathererNameValidation = regexp.MustCompile("^[a-z]+[_a-z]*[a-z]([/a-z][_a-z]*)?[a-z]$")
+	gathererNameExceptions   = []string{"ContainersLogs"}
```
tremes, line 33: Is this exception here because of https://github.com/openshift/insights-operator/blob/master/pkg/gatherers/conditional/filter_and_gather_container_logs.go#L51 ?

opokornyy: Yes, it has different structure and the regex does not match the gatherer name, because there is none.

opokornyy: If it possible to add the name then let's do that

ncaak (approved):
/lgtm

PR #1081, opened 2025-04-03
CCXDEV-15083: Move the Insights runtime extractor feature to GA

jmesnil:
## Categories

- [X] Feature

## Documentation

TBD

## Breaking Changes

No

## References

https://github.com/openshift/enhancements/blob/master/enhancements/insights/insights-runtime-extractor.md#tech-preview---ga


jmesnil:
@tremes As we discussed, the [feature gate  for the insights runtime extractor](https://github.com/openshift/api/blob/6fb534fddf16f9265a5eb02c572599ec373ca2d4/features/features.go#L119) will be udpated to become GA.

This PR ensures that the resources will be applied by default (so the DaemonSet will always be deployed) but this leaves the opportunity for the users to disable the gathering of the runtime data by disabling the feature gate.

Is that ok with you?

jmesnil:
@tremes I've removed the check to the `FeatureGateInsightsRuntimeExtractor`.

I'm not sure what to do with the openshift/api regarding the definition of that feature though? Should it be removed, moved to `Default`?

pkg/gatherers/workloads/gather_workloads_info.go
opokornyy, line 1: Should we also remove the `Tech Preview` here?

```// - Image repository is now collected if it comes from outside the Red Hat domain
// - [Tech Preview] runtime info for workloads are collected (since 4.18.0)
func (g *Gatherer) GatherWorkloadInfo(ctx context.Context) ([]record.Record, []error) {

opokornyy:
I checked the changes and it looks good to me, thanks!

tremes:
@jmesnil I am not 100% sure, but I would try to remove it. This PR would also deserve some JIRA link in my opinion (I don't know if we have something). 

I provisioned a cluster based on this PR and basic functionality seemed OK.
/approve
/lgtm

jmesnil:
> @jmesnil I am not 100% sure, but I would try to remove it. This PR would also deserve some JIRA link in my opinion (I don't know if we have something).

I've created https://issues.redhat.com/browse/CCXDEV-15083 to track it. thanks!

BaiyangZhou:
/override pull-ci-openshift-insights-operator-master-e2e-gcp-ovn-techpreview

PR #1086, opened 2025-05-19
OCPBUGS-55804: Report an error when the insights-runtime-extractor is …

jmesnil:
…not running

If there are no running pods for the insights-runtime-extractor, report it as an error. There should be 1 running pods on every cluster's worker node. The number can differ based on the cluster size but the absence of any running pods is an error to report to the user.

This fixes https://issues.redhat.com/browse/OCPBUGS-55804

## Categories

- [X] Bugfix

## Breaking Changes

No

## References

https://issues.redhat.com/browse/OCPBUGS-55804

### DIFF
--- a/pkg/gatherers/workloads/gather_workloads_runtime_infos.go
+++ b/pkg/gatherers/workloads/gather_workloads_runtime_infos.go
@@ -102,7 +102,7 @@ func getInsightsOperatorRuntimePodIPs(
 		return nil, err
 	}
 
-	var runtimePods []podWithNodeName
+	runtimePods := []podWithNodeName{}
 	for i := range pods.Items {
 		pod := &pods.Items[i]
 		running := pod.Status.Phase == corev1.PodRunning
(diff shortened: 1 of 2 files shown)


jmesnil:
I'm not sure that the failures are related to the bug fix.

opokornyy:
> I'm not sure that the failures are related to the bug fix.

No they are not, can you please rebase the PR? It should then pass the tests

jmesnil:
@opokornyy I will set [OCPBUGS-55804](https://issues.redhat.com//browse/OCPBUGS-55804)'s target version to 4.20.
I think it is worth considering fixing it for OCP 4.19.
I can clone the issue to target 4.19 and open a PR against `release-4.19` to make sure this issue is fixed in the upcoming 4.19. wdyt?

opokornyy:
> @opokornyy I will set [OCPBUGS-55804](https://issues.redhat.com//browse/OCPBUGS-55804)'s target version to 4.20. I think it is worth considering fixing it for OCP 4.19. I can clone the issue to target 4.19 and open a PR against `release-4.19` to make sure this issue is fixed in the upcoming 4.19. wdyt?

I think it is a good idea to also fix it in the 4.19, so lets do that

PR #1186, opened 2025-11-27
CCXDEV-15591: on demand gathering job limit

opokornyy:
This PR adds an informer that watches updates to the DataGather CR and introduces functionality to limit the number of running on-demand gathering jobs.

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [X] Feature
- [ ] Backporting

## Sample Archive

- `None`

## Documentation

- `None`

## Unit Tests

- `pkg/controller/periodic/periodic_test.go`
- `pkg/controller/periodic/datagather_informer_test.go`

## Privacy

## Changelog

- `None`

## Breaking Changes

No

## References

[CCXDEV-15591](https://issues.redhat.com/browse/CCXDEV-15591)

### DIFF
--- a/pkg/controller/periodic/periodic.go
+++ b/pkg/controller/periodic/periodic.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"golang.org/x/exp/slices"
+	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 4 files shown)


pkg/controller/periodic/periodic.go
```diff
@@ -311,6 +312,8 @@ func (c *Controller) periodicTrigger(stopCh <-chan struct{}) {
 	}
 }
 
+const gatheringLimit = 3
```
opokornyy: This value was used for testing, and we should still discuss whether it should be configurable or not. If not, we need to decide what value should be used here.

PR #1254, opened 2026-03-19
OCPBUGS-78587: add WorkloadAnonymizer with obfuscation config

opokornyy:
This PR is adding WorkloadAnonymizer that is used to load correct configuration for obfuscation from ConfigMap, DataGather and InsightsDataGather. The anonymization logic is still implemented directly in the gatherer and should be implemented in the follow up task https://redhat.atlassian.net/browse/CCXDEV-15394.

## Categories

- [X] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

- `None`

## Documentation

- `None`

## Unit Tests

- `TODO`

## Privacy

## Breaking Changes

No

## References

https://redhat.atlassian.net/browse/OCPBUGS-78587

### DIFF
--- a/pkg/anonymization/anonymizer.go
+++ b/pkg/anonymization/anonymizer.go
@@ -30,7 +30,8 @@ import (
 type AnonymizerType string
 
 const (
-	NetworkAnonymizerType AnonymizerType = "networking"
+	NetworkAnonymizerType       AnonymizerType = "networking"
+	WorkloadNamesAnonymizerType AnonymizerType = "WorkloadNames"
 )
 
 type DataAnonymizer interface {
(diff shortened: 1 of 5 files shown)


jmesnil:
I'm not sure if it is relevant but the workload runtime info collected by the insights-runtime-extractor are hashed  by the extractor with the same code than in the IO:

* https://github.com/openshift/insights-operator/blob/c3583e22871aec1175dcc4ed2ae6edf516fb4208/pkg/gatherers/workloads/gather_workloads_info.go#L442
* https://github.com/openshift/insights-runtime-extractor/blob/ce30b4f9bc3ec867b976886a5207d36c50a396d9/exporter/pkg/utils/hash.go#L16

If one day, you decide to change that behavior, the extractor HTTP API has query parameter `hash` (which is `true` by default) that can be set to `false` if you want to do all the hashing from the IO side instead

PR #1271, opened 2026-04-15
CCXDEV-15602: Retry refactor

katushiik11:
This PR implements a refactor to centralize retry logic.

## Categories

- [ ] Bugfix
- [ ] Data Enhancement
- [ ] Feature
- [ ] Backporting

## Sample Archive

None

## Documentation

None

## Unit Tests

- `pkg/retry/retry_test.go`

## Privacy

## Breaking Changes

No

## References

https://redhat.atlassian.net/browse/CCXDEV-15602

## Summary by CodeRabbit

* **New Features**
  * Added a shared exponential-backoff retry utility for more resilient HTTP operations.

* **Refactor**
  * Improved reliability of remote configuration fetching, report uploads, cluster transfers, and certificate requests by standardizing retry behavior and error handling.

* **Tests**
  * Added unit tests covering retry decision logic and backoff scenarios to ensure consistent recovery and failure reporting.

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -1,0 +1,94 @@
+// Package retry provides shared retry logic with exponential backoff for HTTP operations.
+//
+// Usage example (once implemented):
+//
+//	response, err := retry.RetryWithExpBackOff(
+//		wait.Backoff{
+//			Duration: interval/32,
+//			Factor: 2,
+//			Steps: ocm.FailureCountThreshold,
+//			Cap: interval,
+//		},
+//		retry.RetryOn500HTTP,
+//		func() (*Response, error) {
+//			return client.RecvSCACerts(ctx, endpoint, nodeArchs)
+//		},
+//	)
+package retry
+
+import (
+	"net/http"
+
+	"github.com/openshift/insights-operator/pkg/insights/insightsclient"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+)
+
+type RetryStrategy int64
+
+const (
+	// RetryOn500HTTP retries only on HTTP 500+ errors, skips retry for non-HTTP errors (disconnected env)
+	// Used by: sca.go, clustertransfer.go
+	RetryOn500HTTP RetryStrategy = iota
+
+	// RetryOnNon200HTTP retries on any non-200 HTTP status code
+	// Used by: conditional_gatherer.go
+	RetryOnNon200HTTP
+
+	// RetryOnAll retries on all errors
+	// Used by: insightsuploader.go
+	RetryOnAll
+)
+
+// shouldRetry determines if an error should be retried based on the strategy.
+// Returns true if retry should be attempted (when steps remain).
+func shouldRetry(err error, strategy RetryStrategy) bool {
+	switch strategy {
+	case RetryOn500HTTP:
+		// Only retry HTTP 500+ errors, skip non-HTTP errors (disconnected env)
+		if !insightsclient.IsHttpError(err) {
+			return false
+		}
+		httpErr := err.(insightsclient.HttpError)
+		return httpErr.StatusCode >= http.StatusInternalServerError
+
+	case RetryOnNon200HTTP:
+		// Retry on any non-200 HTTP status, or non-HTTP errors
+		if !insightsclient.IsHttpError(err) {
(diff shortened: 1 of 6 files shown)


pkg/retry/retry.go
```diff
@@ -0,0 +1,94 @@
...
+		httpErr := err.(insightsclient.HttpError)
+		return httpErr.StatusCode != http.StatusOK
+
+	case RetryOnAll:
+		// Retry on all errors
+		return true
+
+	default:
+		// Unknown strategy, don't retry
+		klog.Infof("Unknown strategy %d", strategy)
```
ncaak, line 69: I don't expect anyone to ever see this log message because there are no use cases for it. However, if someone does see it, it might be a good idea to explain what happened. "Unknown strategy" seems common and could be confusing if it's seen in a log file amongst 1,000 other lines.

pkg/retry/retry.go
```diff
@@ -0,0 +1,94 @@
...
+		return false
+	}
+}
+
+func RetryWithExpBackOff[T any](bo wait.Backoff, strategy RetryStrategy, operation func() (T, error)) (T, error) {
```
ncaak, line 74: I would prefer a more deterministic approach than accepting `T any`. Overall, it works, and this is more about future maintainability than an immediate problem.

There's no need to change it now, but it'd be better if we agreed on an "interface" for the embedded functions instead of accepting different parameter types, which would make maintenance more complex.

We can make this change in the future if we reuse this function again. For now, it's OK.

pkg/retry/retry_test.go
ncaak, line 1: The tests looks OK. I would have liked to add them in "table format" as they are very similar.

pkg/retry/retry.go
```diff
@@ -0,0 +1,94 @@
...
+		// Retry on any non-200 HTTP status, or non-HTTP errors
+		if !insightsclient.IsHttpError(err) {
+			return true // retry non-HTTP errors
+		}
+		httpErr := err.(insightsclient.HttpError)
+		return httpErr.StatusCode != http.StatusOK
+
+	case RetryOnAll:
+		// Retry on all errors
+		return true
```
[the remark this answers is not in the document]
ncaak: This issue seems fair, and I think there is room for improvement. We could address it as a follow-up after this is merged. It looks like a good idea to put them to use by having the context in the parent functions and an existing function to accept those contexts in the `wait` package.

@katushiik11 please, note it for the near future.

pkg/retry/retry.go
```diff
@@ -0,0 +1,94 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+)
+
+type RetryStrategy int64
+
+const (
+	// RetryOn500HTTP retries only on HTTP 500+ errors, skips retry for non-HTTP errors (disconnected env)
+	// Used by: sca.go, clustertransfer.go
+	RetryOn500HTTP RetryStrategy = iota
```
ncaak, line 32: As a very picky nit (sorry), I would rename this field to something like `RetryOnServerError` or `RetryOn50xHTTP` as it includes >500 values in the logic.

pkg/retry/retry.go
opokornyy: Agree, maybe a result struct could work and instead of generics we could wrap the data into concrete struct and return it, for example something like this.

```
type Result struct {
	Data       []byte // Response data
	StatusCode int    // HTTP status code
	RequestID  string // Request ID
}
```

Then it could be called like this

```
result, err := retry.RetryWithExpBackOff(c.backoff, retry.RetryOnAll, func() (retry.Result, error) {
		requestID, statusCode, err := c.client.SendAndGetID(ctx, c.configurator.Config().DataReporting.UploadEndpoint, *s)
		return retry.Result{RequestID: requestID, StatusCode: statusCode}, err
	})
```

pkg/retry/retry.go
```diff
@@ -0,0 +1,111 @@
...
+	// Used by: insightsuploader.go
+	RetryOnAll
+)
+
+// Result holds the response data from retry operations
+type Result struct {
+	Data       []byte // Response data
+	StatusCode int    // HTTP status code
+	RequestID  string // Request ID
+}
```
opokornyy, line 52: Nit: I would drop these comments as they don't add much value, IMO.

PR #1299, opened 2026-06-03
OCPBUGS-78774: kube-rbac-proxy must honor Central TLS configuration

jmesnil:
The runtime extractor's kube-rbac-proxy sidecar was using default TLS settings instead of the cluster's centralized TLS security policy from the APIServer CR. This change reads the TLS profile from apiservers.config.openshift.io/cluster and injects --tls-cipher-suites and --tls-min-version arguments into the kube-rbac-proxy container.

The controller now watches for APIServer TLS profile changes and re-applies the DaemonSet when the profile is updated. Cipher names are converted from OpenSSL to IANA format using library-go, with graceful fallback to Intermediate profile when ciphers are unsupported.

- [X] Feature

## Unit Tests

- `pkg/controller/runtimeextractor/resources/daemonset_test.go`

## Breaking Changes

No

## References

https://redhat.atlassian.net/browse/OCPBUGS-78774

## Summary by CodeRabbit

* **New Features**
  * Runtime-extractor now retrieves the cluster’s APIServer TLS security profile at startup and applies it to runtime-extractor resources during deployments and drift reconciliation.
  * kube-rbac-proxy TLS settings (minimum TLS version and cipher suites) are now generated from the cluster TLS profile, with fallback behavior when needed.
* **Bug Fixes**
  * Reconciliation now skips runtime-extractor updates when the required resources are not present.
* **Tests**
  * Added unit tests for kube-rbac-proxy TLS argument generation and ensuring the arguments are applied correctly.

### DIFF
--- a/pkg/controller/operator.go
+++ b/pkg/controller/operator.go
@@ -450,7 +450,7 @@ func createTechPreviewInformers(
 
 // fetchTLSProfile reads the TLS security profile from the cluster's APIServer configuration.
 // Returns nil if the profile cannot be read, which will default to Intermediate.
-func fetchTLSProfile(ctx context.Context, configClient *configv1client.Clientset) *configv1.TLSSecurityProfile {
+func fetchTLSProfile(ctx context.Context, configClient configv1client.Interface) *configv1.TLSSecurityProfile {
 	apiServer, err := configClient.ConfigV1().APIServers().Get(ctx, "cluster", metav1.GetOptions{})
 	if err != nil {
 		klog.Warningf("Failed to get APIServer config, defaulting to Intermediate TLS profile: %v", err)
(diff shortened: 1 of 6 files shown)


jmesnil:
@opokornyy I've followed your advice to simplify the code and removed the watcher. I can safely rely on the operator's own controller for that.

opokornyy:
@coderabbitai full review

opokornyy:
/retitle OCPBUGS-78774: kube-rbac-proxy must honor Central TLS configuration

opokornyy:
/override ci/prow/insights-operator-e2e-tests

Known bug in the kubernetes python library

opokornyy:
/override ci/prow/insights-operator-e2e-tests

Known bug in the kubernetes python library

PR #1302, opened 2026-06-07
OCPBUGS-87889: fall back to kube-system/global-pull-secret for Insights token

judexzhu:
## Summary

On ARO HCP clusters, `openshift-config/pull-secret` only contains the ACR registry credential — no `cloud.openshift.com` token. Customers add their Red Hat pull secret (including `cloud.openshift.com`) day-2 via `kube-system/additional-pull-secret`, which HCCO merges into `kube-system/global-pull-secret`. The Insights Operator currently only checks `openshift-config/pull-secret`, so it reports `NoToken` / `GatheringDisabled` even though the token exists on the cluster.

This change makes `updateToken()` check `kube-system/global-pull-secret` as a read-only fallback when `openshift-config/pull-secret` has no `cloud.openshift.com` token.

**Changes:**
- Generalize `fetchSecret()` to accept a namespace parameter
- Add fallback lookup to `kube-system/global-pull-secret` in `updateToken()`
- Add read-only RBAC (`get` only, no update/patch) for `global-pull-secret` in `kube-system`
- Include namespace in `fetchSecret` log/error messages for debuggability
- Add tests for fallback and primary-wins-over-fallback precedence

**Behavior:**
- On standard OpenShift clusters: no change — `openshift-config/pull-secret` has the token, fallback is never reached
- On ARO HCP clusters: finds the token in `kube-system/global-pull-secret` when primary lacks `cloud.openshift.com`
- `global-pull-secret` is read-only — the operator does not manage, write, or claim ownership of this secret
- If `global-pull-secret` doesn't exist (NotFound) or is inaccessible (Forbidden), the fallback is silently skipped

## Test plan

- [x] `make test` — all unit tests pass
- [x] `make lint` — 0 issues
- [x] New test: fallback finds token from `kube-system/global-pull-secret` when primary lacks `cloud.openshift.com`
- [x] New test: primary `openshift-config/pull-secret` takes precedence when both have `cloud.openshift.com`
- [ ] Verify on ARO HCP test cluster that operator picks up token from `global-pull-secret`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Configuration**
  * Added scoped permissions and a retrieval fallback to a global pull secret so reporting can use an alternate pull secret when the primary is unavailable.
* **Tests**
  * Added test cases to validate token selection and priority between the primary pull secret and the global fallback.

### DIFF
--- a/pkg/config/configobserver/secretconfigobserver.go
+++ b/pkg/config/configobserver/secretconfigobserver.go
@@ -19,6 +19,14 @@ import (
 	"github.com/openshift/insights-operator/pkg/config"
 )
 
+const (
+	openshiftConfigNamespace = "openshift-config"
+	pullSecretName           = "pull-secret" //nolint: gosec
+	supportSecretName        = "support"     //nolint: gosec
+	kubeSystemNamespace      = "kube-system"
+	globalPullSecretName     = "global-pull-secret" //nolint: gosec
+)
+
 type ConfigReporter interface {
 	SetConfig(*config.Controller)
 }
@@ -134,7 +142,7 @@ func (c *Controller) fetchSecret(ctx context.Context, namespace, name string) (*
 // Updates the stored tokens from the secrets in the cluster. (if present)
 func (c *Controller) updateToken(ctx context.Context) error {
 	klog.V(2).Infof("Refreshing configuration from cluster pull secret")
-	secret, err := c.fetchSecret(ctx, "openshift-config", "pull-secret")
+	secret, err := c.fetchSecret(ctx, openshiftConfigNamespace, pullSecretName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


judexzhu:
Seems the failed job not relate to my PR change?

`identical failure: Got an unexpected keyword argument 'watch' to method read_namespaced_pod_log` and `TypeError: 'NoneType' object is not iterable`

opokornyy:
> Seems the failed job not relate to my PR change?
> 
> `identical failure: Got an unexpected keyword argument 'watch' to method read_namespaced_pod_log` and `TypeError: 'NoneType' object is not iterable`

yeah, there is an issue in the python kubernetes client that our tests are using, I will override the test failure https://github.com/kubernetes-client/python/issues/2610

pkg/config/configobserver/secretconfigobserver.go
```diff
@@ -134,7 +134,7 @@ func (c *Controller) fetchSecret(ctx context.Context, name string) (*v1.Secret,
 // Updates the stored tokens from the secrets in the cluster. (if present)
 func (c *Controller) updateToken(ctx context.Context) error {
 	klog.V(2).Infof("Refreshing configuration from cluster pull secret")
-	secret, err := c.fetchSecret(ctx, "pull-secret")
+	secret, err := c.fetchSecret(ctx, "openshift-config", "pull-secret")
```
opokornyy: Could you make the namespace names and secret names constants?

opokornyy:
/retitle OCPBUGS-87889: fall back to kube-system/global-pull-secret for Insights token

pkg/config/configobserver/secretconfigobserver.go
judexzhu: Please take a look again. thank you

PR #1325, opened 2026-07-24
OCPBUGS-74211: fix: tolerate all node taints for insights-runtime-extractor DaemonSet

akshitkumawat:
## Summary
- The insights-runtime-extractor DaemonSet had no explicit tolerations, so it only
  ran on the fixed set of taints the DaemonSet controller tolerates by default
  (not-ready, unreachable, disk-pressure, memory-pressure, pid-pressure, unschedulable).
- Any custom taint added by a cluster admin (e.g. node-role.kubernetes.io/gpu) prevented
  the DaemonSet's pods from scheduling on that node, leaving incomplete cluster coverage.
- Added a wildcard toleration (`operator: Exists`) to the DaemonSet pod spec so it
  schedules on every node regardless of taints, per Kubernetes best practice for
  cluster-wide DaemonSets.
## Changes
- `pkg/controller/runtimeextractor/resources/manifests/runtime-extractor-daemonset.yaml`:
  add wildcard toleration.
- `pkg/controller/runtimeextractor/resources/daemonset_test.go`: add
  `Test_loadRuntimeExtractorDaemonSet_TolerateAllTaints` verifying the toleration.
## References
https://issues.redhat.com/browse/OCPBUGS-74211
## Test plan
- `go test ./pkg/controller/runtimeextractor/...` passes, including new and existing tests.
- `go build ./pkg/controller/runtimeextractor/...` succeeds.

## Summary by CodeRabbit

* **New Features**
  * Updated the runtime extractor DaemonSet to run across all node taints (including custom taints) using a catch-all toleration.
  * Refined scheduling to exclude control-plane-only nodes while still targeting worker-labeled nodes for runtime data collection.
* **Tests**
  * Added unit tests to verify catch-all taints toleration behavior.
  * Added unit tests to confirm node affinity excludes master/control-plane roles while allowing workers.

### DIFF
--- a/pkg/controller/runtimeextractor/resources/daemonset_test.go
+++ b/pkg/controller/runtimeextractor/resources/daemonset_test.go
@@ -30,6 +30,21 @@ func Test_loadRuntimeExtractorDaemonSet(t *testing.T) {
 	assert.Len(t, ds.Spec.Template.Spec.Containers, 3)
 }
 
+// Test_loadRuntimeExtractorDaemonSet_TolerateAllTaints ensures the DaemonSet tolerates
+// all node taints (including custom ones added by cluster admins), rather than relying
+// on the fixed set of well-known taints the DaemonSet controller tolerates by default.
+func Test_loadRuntimeExtractorDaemonSet_TolerateAllTaints(t *testing.T) {
+	ds, err := loadRuntimeExtractorDaemonSet()
+	assert.NoError(t, err)
+	assert.NotNil(t, ds)
+
+	tolerations := ds.Spec.Template.Spec.Tolerations
+	assert.Len(t, tolerations, 1, "expected a single wildcard toleration")
+	assert.Equal(t, corev1.TolerationOpExists, tolerations[0].Operator)
+	assert.Empty(t, tolerations[0].Key, "key must be empty to match all taints")
+	assert.Empty(t, tolerations[0].Effect, "effect must be empty to match all taint effects")
+}
+
 func Test_applyDaemonSet(t *testing.T) {
 	tests := []struct {
 		name      string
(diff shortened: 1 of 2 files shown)


pkg/controller/runtimeextractor/resources/manifests/runtime-extractor-daemonset.yaml
```diff
@@ -35,6 +35,11 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
       priorityClassName: openshift-user-critical
+      # Tolerate all taints so this DaemonSet schedules on every node, including
+      # nodes with custom taints (e.g. node-role.kubernetes.io/gpu), consistent
+      # with its purpose of running on every node in the cluster.
```
opokornyy: I would remove the comment since the field is fairly self explanatory

pkg/controller/runtimeextractor/resources/daemonset_test.go
```diff
@@ -30,6 +30,21 @@ func Test_loadRuntimeExtractorDaemonSet(t *testing.T) {
...
+	ds, err := loadRuntimeExtractorDaemonSet()
+	assert.NoError(t, err)
+	assert.NotNil(t, ds)
+
+	tolerations := ds.Spec.Template.Spec.Tolerations
+	assert.Len(t, tolerations, 1, "expected a single wildcard toleration")
+	assert.Equal(t, corev1.TolerationOpExists, tolerations[0].Operator)
+	assert.Empty(t, tolerations[0].Key, "key must be empty to match all taints")
+	assert.Empty(t, tolerations[0].Effect, "effect must be empty to match all taint effects")
+}
```
opokornyy, line 107: I would also drop the test since it isn't testing any logic and therefore doesn't add much value.

opokornyy:
@akshitkumawat  Failures in the [ci/prow/insights-runtime-extractor-tests](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_insights-operator/1325/pull-ci-openshift-insights-operator-master-insights-runtime-extractor-tests/2081715425829195776) seem to be related to this change

pkg/controller/runtimeextractor/resources/daemonset_test.go
akshitkumawat: @opokornyy thanks for your inputs. I have pushed the new changes to the branch.

